PR #16, opened 2016-07-19
Removes invalid json-p(dot)org link

sixlettervariables:
Switched to Wikipedia as the clearinghouse for additional information on JSONP as the previously referenced site is now just a spam landing page.


RobDolinMS:
Thanks very much @sixlettervariables 
LGTM 👍

cleemullins:
Thanks. Nice catch. 

Now, the world just needs a way to keep hyperlinks working for a longer period of time... (this has been a problem since the 90's, and I would send you links, but they're all broken now...)

Issue #18, opened 2016-07-19
Section 11.3.2: the date used in the example does not match its description


francescorw:
The date written in the example is _2015-02-**15**T13:15Z_ but two lines before this date is introduced as "February **13**, 2015, at 1:15 p.m. UTC"

RobDolinMS:
Nice catch @francescorw.  Would you want to submit a pull request to fix this?

PR #19, opened 2016-07-19
Fix microsoft/api-guidelines#18

francescorw:
Corrected the date according to the example.


RobDolinMS:
Thanks for the fix @francescorw  LGTM 👍

PR #20, opened 2016-07-19
Correcting the Idempotency of OPTIONS method

manoj-agarwal:
According to rfc 2616 OPTIONS is an idempotent method: https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html


cleemullins:
I've reached out to the relevant folks (including one of the RFC 2616 author) to make sure there isn't any pushback on this internally, and will update this issue once I hear back. I suspect it's just an oversight, as I can't envision any scenario where having OPTIONS be a mutating, non-idempotent, call makes any sense.

johngossman:
Looks like a copy/paste error from the original table.

cleemullins:
Looks good to me. I've confirmed internally, so we're good.

PR #29, opened 2016-07-20
Replace REST and RESTful appropriately

AlexZeitler:
As Roy T. Fielding suggested on [Twitter](https://twitter.com/fielding/status/755551887059857408), I replaced the terms "REST" and "RESTful" with "HTTP" / "HTTP API".
I think the guide is a good idea in general but it doesn't consider all aspects of REST (esp. Hypermedia / HATEOAS) so it should not use that terminology.
If plain HTTP APIs are the design goal (which is ok) they should just be called HTTP APIs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,5 @@
-# Contributing to the Microsoft REST API Guidelines
-The Microsoft REST API Guidelines is a Microsoft-wide initiative to develop consistent design guidelines for REST APIs. The initiative requires input and feedback from a variety of individuals both inside and outside of Microsoft.
+# Contributing to the Microsoft HTTP API Guidelines
+The Microsoft HTTP API Guidelines is a Microsoft-wide initiative to develop consistent design guidelines for HTTP APIs. The initiative requires input and feedback from a variety of individuals both inside and outside of Microsoft.
 
 To provide feedback, please follow the guidance in this document. Please note that these are just guidelines, not rules. Use your best judgment and feel free to propose changes to anything in this repository, including the contribution guidance.
 
@@ -12,12 +12,12 @@ Please note that this project is released with a [Contributor Code of Conduct][c
 ## Creating issues
 - You can [create an issue][new-issue], but before doing that please read the bullets below and include as many details as possible.
 - Perform a [cursory search][issue-search] to see if a similar issue has already been submitted.
-- Reference the version of the Microsoft REST API Guidelines you are using.
+- Reference the version of the Microsoft HTTP API Guidelines you are using.
 - Include the guidance you expected and other places you've seen that guidance, e.g. [White House Web API Standards][white-house-api-guidelines].
 - Include sample requests and responses whenever possible.
 
 ### Related repositories
-This is the repository for Microsoft REST API Guidelines documentation only. Please ensure that you are opening issues in the right repository.
+This is the repository for Microsoft HTTP API Guidelines documentation only. Please ensure that you are opening issues in the right repository.
 
 ## Recommended setup for contributing
 - Install [Atom][atom]
--- a/Guidelines.md
+++ b/Guidelines.md
@@ -1,5 +1,5 @@
-# Microsoft REST API Guidelines
-## Microsoft REST API Guidelines Working Group
+# Microsoft HTTP API Guidelines
+## Microsoft HTTP API Guidelines Working Group
 
  | | |
 ---------------------------- | -------------------------------------- | ----------------------------------------
@@ -13,18 +13,18 @@ Document editors: John Gossman (C+E), Chris Mullins (ASG), Gareth Jones (ASG), R
 
 
 
-# MICROSOFT REST API GUIDELINES
+# MICROSOFT HTTP API GUIDELINES
 ## 1 ABSTRACT
-The Microsoft REST API Guidelines, as a design principle, encourages application developers to have resources accessible to them via a RESTful HTTP interface. To provide the smoothest possible experience for developers on platforms following the Microsoft REST API Guidelines, REST APIs SHOULD follow consistent design guidelines to make using them easy and intuitive.  
+The Microsoft HTTP API Guidelines, as a design principle, encourages application developers to have resources accessible to them via a HTTP interface. To provide the smoothest possible experience for developers on platforms following the Microsoft HTTP API Guidelines, HTTP APIs SHOULD follow consistent design guidelines to make using them easy and intuitive.  
 
-This document establishes the guidelines Microsoft REST APIs SHOULD follow so RESTful interfaces are developed consistently.
+This document establishes the guidelines Microsoft HTTP APIs SHOULD follow so HTTP interfaces are developed consistently.
 
 ## 2 TABLE OF CONTENTS
 <!-- TOC depthFrom:1 depthTo:3 withLinks:1 updateOnSave:1 orderedList:0 -->
 
-- [Microsoft REST API Guidelines 2.3](#microsoft-rest-api-guidelines-23)
-	- [Microsoft REST API Guidelines Working Group](#microsoft-rest-api-guidelines-working-group)
-- [MICROSOFT REST API GUIDELINES](#microsoft-rest-api-guidelines)
+- [Microsoft HTTP API Guidelines 2.3](#microsoft-http-api-guidelines-23)
+	- [Microsoft HTTP API Guidelines Working Group](#microsoft-http-api-guidelines-working-group)
+- [MICROSOFT HTTP API GUIDELINES](#microsoft-http-api-guidelines)
 	- [1 ABSTRACT](#1-abstract)
 	- [2 TABLE OF CONTENTS](#2-table-of-contents)
 	- [3 INTRODUCTION](#3-introduction)
@@ -44,7 +44,7 @@ This document establishes the guidelines Microsoft REST APIs SHOULD follow so RE
 		- [6.1    IGNORE RULE](#61-ignore-rule)
 		- [6.2    VARIABLE ORDER RULE](#62-variable-order-rule)
 		- [6.3    SILENT FAIL RULE](#63-silent-fail-rule)
-	- [7    REST CONSISTENCY FUNDAMENTALS](#7-rest-consistency-fundamentals)
+	- [7    HTTP API CONSISTENCY FUNDAMENTALS](#7-rest-consistency-fundamentals)
 		- [7.1    URL STRUCTURE](#71-url-structure)
 		- [7.2    URL LENGTH](#72-url-length)
 		- [7.3    CANONICAL IDENTIFIER](#73-canonical-identifier)
@@ -78,7 +78,7 @@ This document establishes the guidelines Microsoft REST APIs SHOULD follow so RE
 		- [10.5    USING A DELTA LINK](#105-using-a-delta-link)
 	- [11    JSON STANDARDIZATIONS](#11-json-standardizations)
 		- [11.1    JSON FORMATTING STANDARDIZATION FOR PRIMITIVE TYPES](#111-json-formatting-standardization-for-primitive-types)
-		- [11.2    REST GUIDELINES FOR DATES AND TIMES](#112-rest-guidelines-for-dates-and-times)
+		- [11.2    HTTP API GUIDELINES FOR DATES AND TIMES](#112-rest-guidelines-for-dates-and-times)
 		- [11.3    JSON SERIALIZATION OF DATES AND TIMES](#113-json-serialization-of-dates-and-times)
 		- [11.4    DURATIONS](#114-durations)
 		- [11.5    INTERVALS](#115-intervals)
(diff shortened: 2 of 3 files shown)


hhariri:
The document makes no mention of any of the constraints a "RESTful" system would need to abide to. It makes no mention of hypermedia as @AlexZeitler mentions and completely muddies an already misused term to refer to something that has nothing to do with REST. The document itself can be very useful, but please call it what it is, which is your HTTP API guidelines. JSON and pretty URLs aren't what make REST.

mcintyre321:
Yes I also agree, this document seems to be about HTTP APIs not REST and should be updated to remove reference to REST.

AydinMirMohammadi:
Yes - I also agree. Please replace REST with HTTP.

forki:
omg

mcintyre321:
@AlexZeitler You should update the PR to modify the readme.md and contributing.md also.

ianbattersby:
+1

glennblock:
Yes. REST is a very clear set of constraints: client server, uniform interface, layered system, stateless, catchable, code on demand and hypermedia.

* It is not about URIs
* It is not about dates and timestamps
* It is not about JSON
* It is not about objects

This document in its current form is misleading and harmful. It further confuses / muddies the term REST. Beyond that it adds ZERO value.

AlexZeitler:
@mcintyre321 done.

Oceanswave:
👍

hpoom:
👍

sharwell:
> REST is a very clear set of constraints: client server, uniform interface, layered system, stateless, cacheable, code on demand and hypermedia.

I find that in practice, REST is frequently very clear but rarely consistent. I support the use of "HTTP" over "REST" even in cases where the service is "RESTful" simply because it eliminates some of the retraining burden for all the different but strongly-held views on what REST is/isn't.

cleemullins:
I'm leaning against this, as I believe most developers mean "CRUD API using easy-to-understand URLs, URI segments, and standard HTTP commands" when they say REST. 

My preference is to meet the expectations of the majority of API consumers, and I believe "HTTP API" wouldn't be what they expect. Personally, if I read "HTTP API" I'm expecting either SOAP or very RPC style APIs. 

I do agree, that the lack of (for example) Hypermedia makes this less than ideal from a "Strict Definition of REST" perspective. That's why we have linked off to more authoritative definitions everywhere we can. 

I'm very open to updating the intro to be more in line with what's being said in this thread.

forki:
@cleemullins why do we give things names anyway!?

hhariri:
@cleemullins Everyone has a REST API because an HTTP API doesn't sell from a marketing perspective.

jeppec:
@cleemullins I'm think you're right that most developers mean "CRUD API using easy-to-understand URLs, URI segments, and standard HTTP commands" when they say REST.
In the document you mention that the guideline should "Adhere as closely as possible to accepted REST/HTTP best practices in the industry at-large". CRUD API using easy-to-understand URLs, URI segments, and standard HTTP commands, in my definition, isn't best practice - it's just common practice.

For a guidelines like this I'm not sure applying the common practice of most developers will yield a result that brings us much forward. 
I support renaming from REST to HTTP to make it clear that the approach isn't Restful/HATEOAS or Level 3 according to the Richardson maturity model - http://martinfowler.com/articles/richardsonMaturityModel.html

johngossman:
My question for the thread is whether folks believe anything in the body of the guidelines is "wrong" or contradicts building a RESTful service? We point people to REST references and then the document is about the "protocol", hence things like JSON formatting, data formats, etc.

By analogy, back in the day when we had a C++ style guide (which we probably still do), it said "Learn the C++ language, do OO programming, here's those references) then the bulk of the document talked about how to use spaces and tabs and where to put brackets...none of which taught the developers about C++ programming or OO.

If we change the wording to HTTP, we're still going to advise people to write RESTful services (as a SHOULD not a MUST).

Ironically, we wrote the document to try to avoid arguing over the definition of REST.

sharwell:
> Ironically, we wrote the document to try to avoid arguing over the definition of REST.

I had a feeling this was the case.

> If we change the wording to HTTP, we're still going to advise people to write RESTful services (as a SHOULD not a MUST).

I recommend a MAY instead of SHOULD, on the premise that service developers should be able to adhere to all of the SHOULD portions in a design guidelines document without substantially impairing the efficiency of the service placed in widespread use. In this case, the use of link relations instead of direct URI construction for API calls introduces a substantial overhead in response sizes, increasing bandwidth requirements for servers and slowing down client operations. Since the benefits these relations provide in terms of discoverability are generally meaningless to the overwhelming population who are just users of applications that communicate over HTTP, services should instead prefer documented and stable URI construction practices for efficiency. The MAY regarding a RESTful service only comes into play for a "developer mode", though clear documentation of the API remains a more powerful tool of discoverability than this developer mode ever could.

asbjornu:
@johngossman, If the guidelines have no mention of hypermedia or hypertext as the engine of application state (HATEOAS), it by definition does not describe anything reminiscent of REST. REST stands for Representational State Transfer, which explicitly requires the use of hypermedia. If there's no hypermedia, it's not RESTful.

So while the guidelines don't work contrary to building a RESTful API, they don't really help either.

mcintyre321:
Here's a quote from the guy who invented the ten REST:


'''

What needs to be done to make the REST architectural style clear on the notion that hypertext is a constraint? In other words, if the engine of application state (and hence the API) is not being driven by hypertext, then it cannot be RESTful and cannot be a REST API. Period.


'''

garethj-msft:
My view, frankly, is that the title of the document is an attractor more than it is necessarily accurate in the scholarly sense.  I'm supportive of dialing down the use of RESTful and REST within the document where they are technically inaccurate against the original definition of the term per @fielding, perhaps in favor of resource-based api or something (still a bit of a mouthful).  However, I'm a huge believer that language is a tool for achieving ends, and at the title/readme level, the best communication service I believe we can bring is to get folks who consider that they are building REST APIs, whether they are or not per definition, to take a look at this work and see if they find value in it.  I believe we'll reduce the set of folks who do so if we change the title and readme to remove REST and I feel its worth sacrificing some accuracy against the original definition to avoid that.

glennblock:
The biggest issue I have is this document is that it leaves the reader who does not understand what REST is that the contents therein are what define a good RESTful service. And that is not true actually it is orthogonal. The links to Wikipedia and the REST IN Practice book make this even worse.

The majority of the guidelines though here have nothing to do with REST. I am not saying they are good or bad, they are just something that the REST constraints do not address. So the word ends up really having no value other than maybe saying it is not SOAP? 

For example,  ""11.2 REST guidelines for dates and times". If I drop the "REST" from that phrase does it have any different meaning? No. Does REST address dates and times no? In most of the cases where REST is used, barring from the section where the guidelines encourage folks to learn about REST, the term is misused.

+1 on toning down completely the usage of the term REST within the document. Encouraging people to learn about REST, Hypermedia, is great.

hhariri:
@johngossman In answer to your question, I believe there is something wrong, because the document is misleading. It is leading people to believe that was is described here is how you create a RESTful service. 

Given the opportunity you have here, and as you point out, your own intentions to stop having people debate on the term, why not make the suggested corrections and then add a section specifically about RESTful constraints and the consequent benefits of imposing these constraints. 
This way those that want to adhere to REST know exactly what they need to do and the benefits it provides. Those that don't, can simply use this document as a standard for HTTP APIs up to the point where you talk about REST. 

You have a great opportunity to make things rights and contribute to clearing up the term that is so very often misunderstood. 

Please reconsider.

darrelmiller:
Personally, I think the ship has already sailed on the misuse of the term REST, and there are many players in the industry to be blamed for that, however I fear that if MSFT don't use the term REST here, then many customers who might benefit from some the contained guidance will not use it because "they want to do REST"!.  Jon Moore told us in the keynote at Restfest 2011 to give up on policing the REST term and just use Hypermedia API if you want to talk about hypermedia APIs.  I think that wisdom is appropriate here.

@johngossman The formal definition of REST versus the popular definition of REST applies very specific rules on how clients and servers can be coupled.  There are some things in the guidelines that go against the REST constraints. e.g.
 
- Specifically defining the format and shape of an error payload.  Using the content-type header to identifier a media type such as the one described in [RFC7807](https://tools.ietf.org/html/rfc7807) would comply with the REST constraints.
- Requiring that all services MUST support `application/json` may not directly contradict a REST constraint, but it does seem to violate the intent of building evolvable systems.
- Making CORS a MUST is similarly constraining.  Many APIs are not designed to be consumed by web browsers.
- Attempting to standardize URL structures such as `https://{serviceRoot}/{collection}/{id}` encourages additional coupling between clients and servers leading to difficulty in deploying services and evolving services without breaking clients.  This is discussed further in [RFC 7320](https://tools.ietf.org/html/rfc7320)
- The requirements that APIs MUST include a version identifier in a URL is another example of a guideline that is contrary to the formal REST constraints.

These are just some of the things that I have seen so far that are not consistent with goal of building highly evolvable systems.  That doesn't mean they are wrong, or bad guidelines, they just achieve a different set of system effects than those that the REST constraints are aiming for.

 It might be worth adding a paragraph, that explains the use of the term, for those who are offended by its liberal application.

glennblock:
I am not against the use of REST as its intended use is clarified. When we say REST, we mean HTTP services, etc.

cleemullins:
@darrelmiller 

Thanks for the detailed feedback. I think some of this is explained in our goals section - which perhaps should be made a bit stronger. For example:

> Requiring that all services MUST support application/json may not directly contradict a REST constraint, but it does seem to violate the intent of building evolvable systems.

That's not a REST thing, but rather a practicality of life at Microsoft issue. We have "REST" API's that return JSON, some that return XML, some that return, well, something else. In order to make life practical for our consumers, we as a company decided to just say "JSON!". By putting this into the guidelines, and giving the API review board(s) the power enforce, at least we're not building new API's that are XML first and "maybe have JSON some day". I don' think this speaks to general REST at all, but rather building API's here at Microsoft. 

A similar point with our error payloads. We have hundreds of teams building APIs, and each of them did/does return errors in a different way. We understand this makes our APIs difficult to consume so we tried to standardize our error format. Again, this isn't a commentary on REST, rather it's "To build an API at Microsoft, you should return errors this way so that our customers don't have to learn yet another error format".

hhariri:
@cleemullins So in essence you yourself are pointing out that this is about building API's at Microsoft and following certain conventions. So why not call it "Microsoft HTTP API's"? As opposed to Microsoft REST API's?

AlexZeitler:
@cleemullins btw there's a RFC for how errors can be returned in REST/HTTP APIs: https://tools.ietf.org/html/rfc7807

cleemullins:
@AlexZeitler Thanks for that RFC Pointer. I hadn't seen it before, and just spent an hour reading it. I've opened up an internal discussion to gauge the mood for migrating off our existing "Standard" to the one specified by the RFC. 

Are you aware of any other large API providers snapping to this spec? A success story (or 2 or 3) around the format, and the wins from standardizing on it, would be very helpful.

cleemullins:
@hhariri Thanks Hadi for your questions. This doc is **absolutely** about building API's at Microsoft following our internal conventions. We think the conventions we're using internally are useful enough that it was worth publishing so others can hopefully benefit from some of the work. If that message didn't come across, then it's my fault and I'll try to correct it. 

This doc is saying, "To better serve the many developers who consume our public facing APIs, we should try to make them more consistent.".  This doc is an acknowledgement that if you have 100 teams building APIs, then you get 100 very different APIs. We are trying really hard to improve that, and this doc is one of those efforts. 

I'm avoiding the HTTP v REST discussion, as I don't really have anything to add there.

AlexZeitler:
@cleemullins We're currently using it in a SaaS-Project on Azure but it's not officially available at this time.
Maybe @darrelmiller has further info as he has written the ASP.NET Web API implementation for Problem+JSON: https://github.com/tavis-software/Tavis.Problem

Also maybe @mnot has more infos to share as he as contributed to the RFC.

bwatts:
How about "Web APIs"? It aligns with well-known product nomenclature and avoids the co-opting and dilution of the term REST.

hhariri:
@bwatts It also might confuse people with a now obsolete stack. Is there anything wrong with actually using HTTP given that it seems the entire guidelines rest on this protocol?

bwatts:
@hhariri not at all, but it doesn't really seem to enthuse anyone either. I'm just looking for something that will have an impact on the authors of this document. The tone of the conversation right now is "yeah, we know it's not right, but meh" and it's unclear what anyone could say to move the needle.

hhariri:
@bwatts I'd actually love to hear from the original authors their thoughts, because if they don't care, then I'm pretty much done here.

mcintyre321:
If HTTP API is too broad (and unmarketable) a term, perhaps we can coin UNREST, as in UNREST's not REST (a la GNU). 

It would be great to have Microsoft back such a move as all the grassroots efforts to save the term REST have seemingly failed.

distobj:
Or, you know, just add a firm recommendation that hypermedia be used; minimize use of ids that aren't URLs, etc..

glennblock:
Or just API. Look around, if you go to Yelp's page it is the Yelp API, GitHub API, Bing API.

People get it....MIcrosoft API.

darrelmiller:
@distobj That's just crazy talk ;-)  It would be awesome if we could get there one day, but there are just not enough successful examples to point to (excluding of course the elephant in the room) to convince people that it will work.

johngossman:
I think @glennblock has the proper answer in #48

tgschulte:
Maybe just call it "RESTful API" and not "REST API" that seems how most employ an adjective to note that on the spectrum of dogmatic to pragmatic, they are some distance from pure dogmatic implementation.

darrelmiller:
@tgschulte Personally I don't think that making a distinction between REST and RESTFul helps the situation.  And just an FYI, I personally find it offensive when applying all the REST constraints rigorously is described as "dogmatic".  I apply the constraints because I know the costs and the benefits of applying them, not because I read somewhere that I must apply them to get official seal of REST.  
All application of the constraints described by REST should be done pragmatically, whether you apply only one, or apply all of them.

tgschulte:
@darrelmiller Just to be clear, I didn't mean anything offensive by "dogmatic" and I apologize if I did offend. 

We here are wrestling with the way to describe succinctly our API as having REST characteristics while acknowledging some specified REST characteristic is not implemented or implemented differently for reasons of practicality. ATM, "RESTful" is feeling like an economic shorthand for that reality especially in emails/posts/etc. where more elaborate disclaimers seem out of place. I have been watching this thread to see what direction it would go for what we could apply.

darrelmiller:
@tgschulte No worries :-)

I agree that it would be ideal if there were a term for the "architectural style" for the REST constraints, minus hypermedia and self-descriptive messages.  However, I think "RESTFul" and "REST" are used interchangably in too many places to consider trying to make them distinct, despite the fact a few people have been trying to do that for a number of years.  The challenge is that "REST" has the marketing buzz, so any new name that we come up with will be rejected.

distobj:
In my many years of implementing Web based systems (including some time consulting at Microsoft) I can count on one hand the number of times it was advantageous to disregard the hypermedia constraint. I seriously doubt that there are practical reasons why this was not included in the guidelines. They are *guidelines* after all, not canon; put in the recommendation to use it, and leave it up to the individual groups to decide the cost for themselves. If these guidelines continue to RFC 2119 language (ugh), make it a SHOULD if you must (pun intended).

darrelmiller:
@distobj The practicalities are that there are not enough developers who understand how to take advantage of hypermedia. There are not enough people training other developers on how to do it, and there it a significant amount of unlearning that needs to be done before practical benefits can be seen.  Oh, and tooling vendors haven't figured out how to automate the production of hypermedia APIs.  Also, hypermedia is a long term play with some higher upfront costs.  Sadly many people don't build software with those goal.

mcintyre321:
Having a paragraph explaining that this document is for a "[Richardson Maturity Model Level 2](http://martinfowler.com/articles/richardsonMaturityModel.html)" API might go some way to striking a balance between correctness and incorrectness and providing a pointer to those interested to learn more.

AlexZeitler:
> I agree that it would be ideal if there were a term for the "architectural style" for the REST constraints, minus hypermedia and self-descriptive messages.

Borrowing from Scrum: RESTBut :smile:

darrelmiller:
@mcintyre321 I'd be more enthused to support that if Leonard hadn't decided his MM was a bad idea. ;-)

glennblock:
@darrelmiller @mcintyre321 yes exactly. Richardson is distancing himself from RMM, because he never meant in to be a yard stick for RESTfullness.

distobj:
Talk about setting the bar low, folks. Hypermedia is hardly rocket science, it's just a different way of using already familiar technologies. How about something as simple as "Identifiers SHOULD be URLs"?

glennblock:
@darrelmiller a lot of Microsoft's APIs (at least Azure) are based on OData and do incorporate hypermedia.

akuckartz:
:+1: no REST without HATEOAS

garethj-msft:
I think we've made the modification to tone down that we intend to make, and that we've made it clear that we don't intend to change the overall title, which while factually debatable brings, as @darrelmiller says the largest audience who might benefit from the guidance.  if there are any more targeted changes at this point that would help clarify further I'm happy to entertain them.

PR #32, opened 2016-07-20
Use a check mark instead of true/false for readability

sharwell:
I find that the interleaving of the work "False" with "True" makes it more difficult to read the **Is Idempotent** column than is necessary. In this review, I updated the table to use a check mark instead of "True", and simply omit the word "False".

### DIFF
--- a/Guidelines.md
+++ b/Guidelines.md
@@ -235,13 +235,13 @@ Below is a list of verbs that Microsoft REST services SHOULD support. Not all re
 
 Verb    | Description                                                                                                                | Is Idempotent
 ------- | -------------------------------------------------------------------------------------------------------------------------- | -------------
-GET     | Return the current value of an object                                                                                      | True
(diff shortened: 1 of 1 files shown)


dodekeract:
Why not use [ ] and [x]?

garethj-msft:
I actually find the opposite due to the empty lines.   Could we get some more comment please on how people find it?

dodekeract:
@garethj-msft What about inserting ✘ or a similar character into the "False" lines?

garethj-msft:
There doesn't seem a lot of energy for this, so I'm leaning towards rejecting it.

RobDolinMS:
I agree with @sharwell that we could have a stronger contrast.  I'm not an expert in Markdown, but it would be nice to possibly use Check+Green and X+Red (where the selected shades of green and red are strongly contrasting)

garethj-msft:
I'm going to close this rather than leaving it languishing.

PR #34, opened 2016-07-21
Change URL length to be more general

jamesjnadeau:
there are a lot of other browsers than IE that impose a URL length limit, trying to make this a little less biased....

### DIFF
--- a/Guidelines.md
+++ b/Guidelines.md
@@ -219,7 +219,6 @@ Services that can generate URLs longer than 2,083 characters MUST make accommoda
 
  * [http://stackoverflow.com/a/417184](http://stackoverflow.com/a/417184)
  * [https://blogs.msdn.microsoft.com/ieinternals/2014/08/13/url-length-limits/](https://blogs.msdn.microsoft.com/ieinternals/2014/08/13/url-length-limits/)
- * [https://boutell.com/newfaq/misc/urllength.html](https://boutell.com/newfaq/misc/urllength.html)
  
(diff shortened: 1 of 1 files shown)


sluu99:
@jamesjnadeau if we are going to make this general, maybe we should add more links to the line below as well? Right now this link only describes the limitation of IE: https://blogs.msdn.microsoft.com/ieinternals/2014/08/13/url-length-limits/

jamesjnadeau:
@sluu99:
Agreed, I just added some more, would love any other links anyone would recommend. 

I also added a note about checking your technology stacks limitations.

cleemullins:
@jamesjnadeau This is a nice change. The only issue I have is that I'm not comfortable with the link to the [boutell.com]( https://boutell.com/newfaq/misc/urllength.html) site. That seems seems non-authoritative, has many broken links, almost always gives me a "Connection Reset" error when viewing in Chrome. 

If there is a better source of this info, I'm all for it. 

When I originally wrote the section, I looked through all of the RFCs and noticed that all of the URL's (except IE) were so big it wasn't an issue for APIs as we're describing them.

jamesjnadeau:
I'm fine with removing the boutell.com site. Most other sources I looked at referenced it, so that's why it made the cut.

jamesjnadeau:
PS never had any connection issues with viewing it on chrome locally

looking at it again, it's rather dated

> 2006-10-13

cleemullins:
This looks good to me. 

@garethj-msft or @RobDolinMS ?

rgregg:
Looks good to me.

PR #36, opened 2016-07-21
fix typos

aqnouch:


cleemullins:
This looks fine. Good eyes. My eyes just skip over typos like that, regardless of how often I see them. 

Do to the simplicity of this change, I'm going to break the rules and approve it with just myself as the approver. :)

aqnouch:
:eyes:  thank you

PR #37, opened 2016-07-22
Fix typo in Guidelines.md

christianberg:


cleemullins:
@christianberg Thanks for catching that. Makes me say, "Duh!" that I got it wrong in the first place, especially as the correct spelling is right there in the Wikipedia link! 

LGTM. Can I get a 2nd?

rgregg:
:+1:

Issue #7, opened 2016-07-19
Section 7.4 Supported Verbs should really be Supported Methods


darrelmiller:
Although it is common practice to use the phrase HTTP verb, it is inconsistent with the [HTTP specification](https://tools.ietf.org/html/rfc7231#section-4.3), the [.net framework](https://msdn.microsoft.com/en-us/library/system.net.http.httpmethod%28v=vs.118%29.aspx?f=255&MSPPError=-2147217396) and the [IANA Method Registry](https://www.iana.org/assignments/http-methods/http-methods.xml). 

Although most experienced developers will know that verb and method is synonymous in the case of HTTP, this introduces a potential for confusion for new developers.

cleemullins:
Up for writing the PR?

darrelmiller:
@cleemullins Sure.  I'm going to finish the 2616->723x PR first.  That's taking a bit longer than I thought because there were a few things I need to research around the reference to header values.

cleemullins:
The PR for this has been merged.

PR #42, opened 2016-07-22
Replace the term verb with method #7

darrelmiller:


cleemullins:
I was all ready to push back on this, with "I've read these RFC's hundreds of times, and it's VERB not Method". Then, just now, I went and re-read the relevant RFC's yet one more time. 

Damn.

All these years I've been reading the word "Method" and thinking it said "Verb". 

Thanks for this. Good to get it right, especially when "Right" is so clearly "method" and NOT "verb". 

LGTM. Can I get a second?

darrelmiller:
:-) I've been there.

sharwell:
:+1:

glennblock:
@cleemullins it is super common that people use VERB though it is never mentioned in the RFC.

RobDolinMS:
Thanks for this @darrelmiller.  The PR looks like a solid find/replace of verb -> method.  
Per my own sometimes referencing of HTTP methods as HTTP verbs, and @cleemullins's and @glennblock's comments above; I hesitate on a LGTM.  
Would it make sense to also include a sentence in the opening paragraph of section 7.4 mentioning that these are sometimes known as "verbs"?

johngossman:
I agree with @RobDolinMS on a quick sentence about methods often being called verbs.

glennblock:
That seems reasonable to me.

cleemullins:
Ok. @darrelmiller If you could add in the quick sentence that Rob & John are referring to, I can merge once that's done.

cleemullins:
@darrelmiller I'm good to merge this. @garethj-msft Your turn!

garethj-msft:
I would like to get the sentence per @glennblock @RobDolinMS & @johngossman if you don't mind, then I'll go ahead and merge Darrel.

darrelmiller:
@garethj-msft I have merged master and added the line about the use of HTTP verb.  So, if that sentence works for you, this should be good to go.

garethj-msft:
LGTM

PR #45, opened 2016-07-22
Use backticks instead of quotes for code literals

sharwell:
:memo: This request only affects §7.10.2. If you prefer this formatting to the current, the rest of the document will need to be updated for consistency. I didn't want to do all the work without getting some feedback first. :smile: 

* Use `key` instead of "key" as the formatting for naming literal keys in JSON data
* Use `"Value"` instead of "Value"` as the formatting for example JSON values

### DIFF
--- a/Guidelines.md
+++ b/Guidelines.md
@@ -374,25 +374,25 @@ Accept: application/json
 #### 7.10.2 ERROR CONDITION RESPONSES
 For nonsuccess conditions, developers SHOULD be able to write one piece of code that handles errors consistently across different Microsoft REST API Guidelines services. This allows building of simple and reliable infrastructure to handle exceptions as a separate flow from successful responses. The following is based on the OData v4 JSON spec. However, it is very generic and does not require specific OData constructs. APIs SHOULD use this format even if they are not using other OData constructs.
 
-The error response MUST be a single JSON object. This object MUST have a name/value pair named "error." The value MUST be a JSON object.
+The error response MUST be a single JSON object. This object MUST have a name/value pair named `error`. The value MUST be a JSON object.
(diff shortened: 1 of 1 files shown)


cleemullins:
@sharwell , Thanks Sam. What's the motivation driving a preference of 'tick' over "quote"? I'm pretty neutral, and will snap to whatever is "Standard" if there is such a thing. I do think the conversion would be a fair bit of work...

sharwell:
@cleemullins It's a personal preference, but I would say it was influenced by the following things:

1. I find it slightly easier to read (which is weird, I expect non-programmers would say the opposite)
2. Quotes are used for a few purposes in the document, so this helps readers see cases where exact text is sent to/from the API
3. I get frustrated by the English language's placement of punctuation inside of the closing quote, especially when it means the reader has to actually decipher whether or not the punctuation should appear in the code

dodekeract:
@cleemullins I also believe that this greatly improves readability. I've also seen similar usage in other projects and usually do this in my projects too.
It's just so much easier to see the difference between language and code. In addition to that, code actually uses a monospaced font, which is what you'll see in your editor/IDE too. :)

cleemullins:
I don't have any strong preference here other than the doc, as written, uses quotes and it's quite a bit of work to change. 

I would like to leave things as they are.

sharwell:
@cleemullins Can you clarify which of the following is your view?

1. It's quite a bit of work to change, and the current form is preferred so we'll go with the current form.
1. It's quite a bit of work to change, and you'd prefer to not take on that work right now. However, if someone (say me) was to apply the change uniformly then it might be merged then.

PR #46, opened 2016-07-22
Add a data model for error responses

sharwell:
The data model for error responses is currently described in text along with a couple of concrete examples. However, there is no explicit data model defined in terms of objects and their allowed properties. I have implemented one possible approach for defining data models and applied it to §7.10.2.

> :bulb: This data model is intended to supplement the existing descriptive language.

This form can be modified as necessary to meet the needs of this document. For example, in a similar document at work we have a **Default** column instead of **Required**, and indicate required items by the lack of a default value. I avoided that approach here because I wasn't sure what the default values would be, or if that approach would even apply to all examples in this documentation.

### DIFF
--- a/Guidelines.md
+++ b/Guidelines.md
@@ -392,6 +392,31 @@ Error responses MAY contain [annotations][odata-json-annotations] in any of thei
 
 We recommend that for any transient errors that may be retried, services SHOULD include a Retry-After HTTP header indicating the minimum number of seconds that clients SHOULD wait before attempting the operation again.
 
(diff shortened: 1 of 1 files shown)


sharwell:
:thought_balloon: It may be appropriate to rename the **Type** column to **Value**.

cleemullins:
@daspek Hey Daron, could you take a look at this and provide feedback?

garethj-msft:
Thanks sharwell - this looks like a good tabular representation of the description above.  Merging.

PR #49, opened 2016-07-24
Fix typo in SLRO section

StephenCleary:
Compare with the text in the RELO section 13.1:

> In this scenario the databases segment is processing the PUT operation.

It seems clear that the SLRO section should say `operations` here:

> In this scenario the _databases_ segment is processing the PUT operation.

### DIFF
--- a/Guidelines.md
+++ b/Guidelines.md
@@ -1134,7 +1134,7 @@ Services MAY enable PUT requests for entity creation.
 PUT https://api.contoso.com/v1.0/databases/db1
 ```
 
-In this scenario the _databases_ segment is processing the PUT operation.
+In this scenario the _operations_ segment is processing the PUT operation.
 
 ```http
 HTTP/1.1 202 Accepted


johngossman:
I'm honestly not sure what the second sentence was intended to say. It may have been copy/pasted from the RELO section. I think it is misleading to say the operations segment is processing the PUT operation...the service behind the database segment is the one processing the operation either way, just reporting progress through the operations url.

StephenCleary:
The second paragraph has the service italicized; this is commonly done to draw attention to the different term. It just looks to me that the term itself mistakenly wasn't changed.

cleemullins:
Unless I'm mistaken, I believe the original is correct. The segment is _databases,_ which is where the operation is being handled.

garethj-msft:
I agree that it is actually correct - however, on reflection, I do think that the repeated statement is adding very little value in either instance.

PR #53, opened 2016-07-25
Fixes for Guidelines.md

zaccharles:
I just finished reading through the whole guidelines. There are some really good recommendations in there. I noticed some issues along the way and fixed them up. Most are very minor.

### DIFF
--- a/Guidelines.md
+++ b/Guidelines.md
@@ -1709,6 +1709,8 @@ All service URLs must be HTTPS (that is, all inbound calls MUST be HTTPS). Servi
 
 We recommend that services that allow client defined Web Hook Callback URLs SHOULD NOT transmit data over HTTP. This is because information can be inadvertently exposed via client, network, server logs and other mechanisms.
 
+However, there are scenarios where the above recommendations cannot be followed due to client endpoint or software limitations.  Consequently, services MAY allow web hook URLs that are HTTP.  
+
 Furthermore, services that allow client defined HTTP web hooks callback URLs SHOULD be compliant with privacy policy specified by engineering leadership. This will typically include recommending that clients prefer SSL connections and adhere to special precautions to ensure that logs and other service data collection are properly handled.
 
 For example, services may not want to require developers to generate certificates to onboard. Services might only enable this on test accounts.  
--- a/Guidelines.md
+++ b/Guidelines.md
@@ -283,7 +283,7 @@ Under UPSERT semantics, a PATCH call to a nonexistent resource is handled by the
 If a service does not support UPSERT, then a PATCH call against a resource that does not exist MUST result in an HTTP "409 Conflict" error.
 
 #### 7.4.4 Options and link headers
-OPTIONS allows a client to retrieve information about an resource, at a minimum by returning the Allow header denoting the valid method verbs for this resource.  
+OPTIONS allows a client to retrieve information about a resource, at a minimum by returning the Allow header denoting the valid method verbs for this resource.  
 
 In addition, services SHOULD include a Link header (see [RFC 5988][rfc-5988]) to point to documentation for the resource in question:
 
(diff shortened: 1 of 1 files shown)


Guidelines.md
```diff
@@ -1577,7 +1577,7 @@ resource        | Yes      | Resource path to watch.
 notificationUrl | Yes      | The target web hook URL.
 clientState     | No       | Opaque string passed back to the client on all notifications. Callers may choose to use this to provide tagging mechanisms.
 
-If the subscription was successfully created, the service MUST respond with the status code 201 CREATED and a body containing at least the following properties, in the same order as the request array, of the subscription:
+If the subscription was successfully created, the service MUST respond with the status code 201 CREATED and a body containing at least the following properties:
```
cleemullins: Why the removal of  the order/request array text?

Guidelines.md
```diff
@@ -1662,10 +1662,10 @@ DELETE https://api.contoso.com/files/v1.0/$subscriptions/{id} HTTP 1.1
 Authorization: Bearer {UserPrincipalBearerToken}
 ```
 
-As with update, the service MUST return 204 NO RESPONSE for a successful delete, and an error status code and body object to indicate failure.
+As with update, the service MUST return `204 No Content` for a successful delete, or an error body and status code to indicate failure.
```
cleemullins: Nice catch on this one!

Guidelines.md
```diff
@@ -1662,10 +1662,10 @@ DELETE https://api.contoso.com/files/v1.0/$subscriptions/{id} HTTP 1.1
...
 
 #### 14.7.4 Enumerating subscriptions
-To get a list of active subscriptions, clients issue a GET request against the subscriptions entity set using a User + Application or Application-Only bearer token:
+To get a list of active subscriptions, clients issue a GET request against the subscriptions resource using a User + Application or Application-Only bearer token:
```
cleemullins: Good change.

Guidelines.md
```diff
@@ -1709,11 +1709,9 @@ All service URLs must be HTTPS (that is, all inbound calls MUST be HTTPS). Servi
 
 We recommend that services that allow client defined Web Hook Callback URLs SHOULD NOT transmit data over HTTP. This is because information can be inadvertently exposed via client, network, server logs and other mechanisms.
 
-However, there are scenarios where the above recommendations cannot be followed due to client endpoint or software limitations.  Consequently, services MAY allow web hook URLs that are HTTP.  
+However, there are scenarios where the above recommendations cannot be followed due to client endpoint or software limitations.  Consequently, services MAY allow web hook URLs that are HTTP. Another example is that services may not want to require developers to generate certificates during onboarding. In this case, services might allow HTTP web hook URLs to be used on test accounts.
```
cleemullins: I don't think this change is needed. Just leaving it as "MAY" is good for me.

cleemullins: In this case, I prefer the original text. :)

Guidelines.md
garethj-msft: I agree with this one - we say earlier in the doc that order should be ignored, so it seems odd to mandate it here.

garethj-msft:
Agreed, great set of minor changes.  Thanks very much for the contribution.  If you could make the small changes we've suggested we can get this merged.

Guidelines.md
zaccharles: The way this sentence is written is a little confusing. At first, it sounds like it's talking about the order of the properties on the response object. As @garethj-msft pointed out, the document previously states that clients should not depend on the order of properties. That would be enough reason to remove it. However, when reading the sentence again I also notice that it says:

> the same order as the /request array/

This is weird because the request is not an array, and nothing in the request is an array. The subscription section reads as though it previously said you could create multiple subscriptions in a single request by sending in an array of objects. That tells me this sentence could also be a strangely worded artefact because now we have this:

>  The POST request contains a single subscription object to be created.

Guidelines.md
zaccharles: I'll change it back because I'm not that bothered by it; the information is still readable and correct. I just didn't the usage of `For example`. That would usually be used to give an example after explaining something, but, in this case, it just seems to random.

zaccharles:
I've made the requested changes. Please let me know if you'd like me to squash the commits, otherwise, I believe this is ready to merge.

dodekeract:
@zaccharles Since a recent change in GitHub, they should be able to "Merge and Squash" a pull-request, so I don't think you should have to do that youself.

zaccharles:
Oh, very cool @dodekeract - I didn't see that yet. Thanks for letting me know :+1:

zaccharles:
Does this need anything else to be merged @garethj-msft ?

cleemullins:
Looks great. Merging now.

PR #59, opened 2016-08-11
Fix #21

### LINKED ISSUE #21: Bug: Example GET request/response in 9.3.1 seems wrong
GET https://api.contoso.com/v1.0/people/123/addresses is shown as an example and returns a people entity back.  Either the request should not have "addresses" or the response should show the addresses collection.

weltan:
Not sure what original author's intent was, but I changed to a more likely JSON response.

### DIFF
--- a/Guidelines.md
+++ b/Guidelines.md
@@ -577,14 +577,8 @@ GET https://api.contoso.com/v1.0/people/123/addresses
 ```json
 {
   "value": [
-    {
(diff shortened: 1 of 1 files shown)


garethj-msft:
Thank you very much weltan, looks good to me.  Appears the original author was thinking of the expand pattern.

cleemullins:
This does look good.

Issue #60, opened 2016-08-17
docs should be one-sentence-per-line


daspek:
The docs should be reformatted to be one sentence per line.
This makes PRs easier to review.
It also makes edits easier by removing the burden of reflowing existing text.

RobDolinMS:
Good suggestion @daspek.  Would you mind drafting a PR?

RobDolinMS:
Closing this issue as @daspek has submitted PR #61

PR #61, opened 2016-08-18
reformatting to one-sentence-per-line

daspek:
see https://github.com/Microsoft/api-guidelines/issues/60


garethj-msft:
Looks good to me.

Issue #62, opened 2016-08-24
Naming guidelines are needed


garethj-msft:
The document is missing detailed guidelines on how to name resources, types in metadata documents, headers etc.

dodekeract:
Are you referring to specific names like "always call X Y" or rules of thumb, like "as short, but precise as possible"?

garethj-msft:
PR #63 is what we believe is needed for our teams.  That has both your categories in a fairly prescriptive stance.

garethj-msft:
Added to vNext branch

PR #63, opened 2016-08-24
Naming guidelines

gareth-a-jones:
Add a new section on naming to address issue #62.
Including practices in wide use across services, especially the Microsoft Graph.

### DIFF
--- a/Guidelines.md
+++ b/Guidelines.md
@@ -103,9 +103,18 @@ This document establishes the guidelines Microsoft REST APIs SHOULD follow so RE
 	- [15    Unsupported requests](#15-unsupported-requests)
 		- [15.1    Essential guidance](#151-essential-guidance)
 		- [15.2    Feature allow list](#152-feature-allow-list)
-	- [16     Appendix](#16-appendix)
-		- [16.1    Sequence diagram notes](#161-sequence-diagram-notes)
-	
+	- [16    Naming guidelines](#16-naming-guidelines)
+		- [16.1    Approach](#161-approach)
+		- [16.2    Casing](#162-casing)
+		- [16.3    Names to avoid](#163-names-to-avoid)
+		- [16.4    Forming compound names](#164-forming-compound-names)
+		- [16.5    Identity properties](#165-identity-properties)
+		- [16.6    Date and time properties](#166-date-and-time-properties)
+		- [16.7    Name properties](#167-name-properties)
+		- [16.8    Collections and counts](#168-collections-and-counts)
+		- [16.9    Common property names](#169-common-property-names)
+	- [17     Appendix](#17-appendix)
+		- [17.1    Sequence diagram notes](#171-sequence-diagram-notes)
 
 <!-- /TOC -->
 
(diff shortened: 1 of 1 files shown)


Guidelines.md
```diff
@@ -1998,12 +2007,100 @@ Content-Type: application/json
...
+- For the overall name of a resource typically shown to users, services MUST use the property name 'displayName'.
+- Services MAY use other common naming properties, e.g. givenName, surname, signInName.
+
+### 16.8 Collections and counts
+- Services MUST name collections as plural nouns or plural noun phrases using correct English.
+- Services MAY use simplified English for nouns that have plurals not in common verbal usage.
+	- e.g. schemas MAY be used instead of schemata.
+- Services MUST name counts of resources with a noun or noun phrase suffixed with 'Count'.
+
+### 16.9 Common property names
```
esatterwhite: I think this should be omitted. It makes too many assumptions about what the underling service does and the contract it exposes.  Some of the names are trying to indicate a data type where others are not which adds a bit of confusion about the reliability of the contract. More importantly, it indicates to consumers that it is OK to make some of those assumptions.

Also, both the `OData` and `OpenAPI` specifications include descriptive schema endpoints which should be used by clients to discern the type and meaning of the data associated with each field. I think it is in the best interest of both the API server and the consumer to separate data format from meaning.

There is also a layer of inflexibility imposed with some of the naming conventions of the fields here. `createdDateTime`, for example would be better simply as `created` and the service schema / metadata would indicate it as a date time field. Additionally, changes to the type do not mean that we need to change the name of the field and/or version the API ( but that is a different discussion ).

garethj-msft: Thanks for the feedback.  What we're trying to drive by this is a set of services that use the same name for similar thing s for Microsoft's services, particularly those exposed on our Graph and ARM endpoints. 

We do not believe that most clients look at the API description contracts before consuming, so naming and docs are key to first level understanding.

As regards the inflexibility, we definitely believe that changing the type of `created` WOULD constitute a breaking change and thus require a major API version revision.  Clients will undoubtedly write code to parse the specific type returned, and changing that will break them.  Making clients dynamically work out what type to parse for a field (based on metadata or some other mechanism) would be too high a bar for a general-purpose API and push into the territory of pretty much requiring client libraries.

esatterwhite: Sure, I see where you are coming from. However, I think clients not inspecting the contract is what make So many systems brittle and tightly coupled. The idea of *Out of Band Communication* about the api creates individual knowledge silos about how particular parts of the system work / have to work. The entire ecosystem around Swagger/OpenAPI and OData is about consuming & trusting the schema contract. I agree with you, many clients do not, but many clients **should**.

I think what really prompted me to insert myself into the conversation more than anything else was the type inference on a field name, but only on the `date` & `datetime` fields. It is very inconsistent within the other naming conventions - nothing else has a type. At the very least, maintain the convention within it self. All fields should include the type, or they should not include a type. I'm in favor of the former.

darrelmiller:
@esatterwhite I prefer to think of the `Date` suffix as a semantic qualifier rather than a contract datatype.  Just like `ActiveFlag`, `ItemCount`, `OrderDescription`.  The actual datatype might be `string` as I might want to return the date in a localized, human friendly form: "CreateDate" : "Lundi, 24 Janvier 2016"

I'm not quite sure I understand the point you are trying to make about OpenAPI.  There is no requirement to include schema in  OpenAPI, it can simply be used to identify resources in an API and it is the very definition of "out-of-band" information.

garethj-msft:
Replaced with an equivalent PR into vNext branch.

PR #64, opened 2016-08-24
Update info about Date headers

### LINKED ISSUE #57: Date Headers
Two issues:

1. `Date` is listed as both a standard request and response header (7.5/7.6), but this is typically found only as a response header.  The HTTP spec ([RFC 7231 § 7.1.1.2](https://tools.ietf.org/html/rfc7231#section-7.1.1.2)) states:

  > A user agent MAY send a Date header field in a request, though generally will not do so unless it is believed to convey useful information to the server.  For example, custom applications of HTTP might convey a Date if the server is expected to adjust its interpretation of the user's request based on differences between the user agent and server clocks.

2. In both places, the description says that the `Date` header is in RFC 3339 format, which is incorrect.  Unfortunately, the HTTP spec doesn't use RFC 3339.  It uses the older format found in [RFC 5322 § 3.3](https://tools.ietf.org/html/rfc5322#section-3.3)  (previously listed under RFC 2822 and RFC 822)

  Example:

      Date:Wed, 03 Aug 2016 20:54:19 GMT

Let me know if you'd like a PR for this.  Thanks.

mj1856:
Fixes #57

### DIFF
--- a/Guidelines.md
+++ b/Guidelines.md
@@ -351,7 +351,7 @@ Many HTTP headers are defined in [RFC7231][rfc-7231], however a complete list of
 Header                            | Type                                  | Description
 --------------------------------- | ------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Authorization                     | String                                           | Authorization header for the request
-Date                              | Date                                             | Timestamp of the request in [RFC 3339][rfc-3339] format
+Date                              | Date                                             | Timestamp of the request, based on the client's clock, in [RFC 5322][rfc-5322-3-3] date and time format.  The server SHOULD NOT make any assumptions about the accuracy of the client's clock.  This header MAY be included in the request, but MUST be in this format when supplied.  Greenwich Mean Time (GMT) MUST be used as the time zone reference for this header when it is provided.  For example: `Wed, 24 Aug 2016 18:41:30 GMT`.  Note that GMT is exactly equal to UTC (Coordinated Universal Time) for this purpose.
 Accept                            | Content type                                     | The requested content type for the response such as: <ul><li>application/xml</li><li>text/xml</li><li>application/json</li><li>text/javascript (for JSONP)</li></ul>Per the HTTP guidelines, this is just a hint and responses MAY have a different content type, such as a blob fetch where a successful response will just be the blob stream as the payload. For services following OData, the preference order specified in OData SHOULD be followed.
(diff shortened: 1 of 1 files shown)


Guidelines.md
```diff
@@ -365,7 +365,7 @@ Services SHOULD return the following response headers, except where noted in the
 
 Response Header    | Required                                      | Description
 ------------------ | --------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
-Date               | All responses                                 | The date the request was processed, in [RFC 3339][rfc-3339] format
+Date               | All responses                                 | Timestamp the response was processed, based on the server's clock, in [RFC 5322][rfc-5322-3-3] date and time format.  This is mandatory for response headers.  Coordinated Universal Time is the preferred time zone reference for this header,  designated as either `GMT` or `+0000`.  For example: `Wed, 24 Aug 2016 18:41:30 GMT`
 Content-Type       | All responses                                 | The content type
```
garethj-msft: Can we switch to normative language pls.  'Coordinated Universal Time SHOULD be used as the preferred time zone reference for this header".

Actually for the response, I think we should consider upping this to a MUST.  I can't see a reason for a service to return this in anything other than UTC.

Guidelines.md
```diff
@@ -351,7 +351,7 @@ Many HTTP headers are defined in [RFC7231][rfc-7231], however a complete list of
...
 Authorization                     | String                                           | Authorization header for the request
-Date                              | Date                                             | Timestamp of the request in [RFC 3339][rfc-3339] format
+Date                              | Date                                             | Timestamp of the request, based on the client's clock, in [RFC 5322][rfc-5322-3-3] date and time format.  This is optional for request headers, but must be in this format if supplied.  Coordinated Universal Time is the preferred time zone reference for this header,  designated as either `GMT` or `+0000`.  For example: `Wed, 24 Aug 2016 18:41:30 GMT`
```
garethj-msft: Can we switch the 'preferred' to normative language please.  See comment below.

garethj-msft:
Thanks very much for your contribution Matt.  I put a couple of small nits in and a suggestion to up a should to a must.  After those, this LGTM.

Guidelines.md
mj1856: Actually, just found that [RFC2616§3.3.1](https://tools.ietf.org/html/rfc2616) says:

> All HTTP date/time stamps MUST be represented in Greenwich Mean Time
   (GMT), without exception. For the purposes of HTTP, GMT is exactly
   equal to UTC (Coordinated Universal Time). This is indicated in the
   first two formats by the inclusion of "GMT" as the three-letter
   abbreviation for time zone, and MUST be assumed when reading the
   asctime format.

So, I will update accordingly.

mj1856:
How's it look now?

garethj-msft:
LGTM

cleemullins:
Looks good to me as well. Merging.

PR #73, opened 2016-11-14
Reword to be better reflect RFC 2119.

royshea:
Section 7.4.3 used the capitalized verb "MAY" without the semantics imposed by
RFC 2119.  The offending phrase is simply describing a situation that can
arise in the real world, rather than a requirement of the specification.  The
requirement put forth in this specification is defined in the next sentence
using the “MUST” verb.


garethj-msft:
Good point.  Merging.

PR #75, opened 2016-11-22
Added the PDF version.

mahdi:
I added the PDF version of markdown which can be easily shared as a single file readable on all devices.


garethj-msft:
Hi Mahdi.  As an evolving document, we can't really take a point in time PDF snapshot.  What toolset did you use to create this - maybe we can spin it up as a webhook on the repo so it stays up to date?

mahdi:
Hi @garethj-msft, I used an online convertor but it looks like this tool can do what you mentioned: [https://cloudconvert.com/api/convert/md-to-pdf](url)

RobDolinMS:
Towards the goal [Madhi stated](https://github.com/Microsoft/api-guidelines/pull/75#issue-190974094) of having something "easily shared as a single file readable on all devices" 

Since the GitHub repo is public and the guidelines.md file is rendered by GitHub as HTML in this view: https://github.com/Microsoft/api-guidelines/blob/master/Guidelines.md ... I believe we have a single file readable on all devices.  

Given the above and [Gareth's concern](https://github.com/Microsoft/api-guidelines/pull/75#issuecomment-262438214) about not wanting a snapshot that could get out-of-date, I'm closing this PR to add the PDF.

dodekeract:
Weird. I can't read a PDF in my terminal. I thought it's more compatible than markdown!

PR #81, opened 2017-03-28
Fix formatting of contibutor table

jawn:
Probably caused by different markdown implementation in github.


garethj-msft:
Yes, looks like something changed - thanks for caching this jawn.

Issue #79, opened 2017-01-27
Question regarding $top


trshiv:
I have a question on the $top URL parameter, which first appears in Section 9.8.2 - Client-driven paging. How is this parameter defined and what is its relation with the $maxpagesize parameter (appearing in Section 9.8.3 - Additional Considerations).

RobDolinMS:
Tagging @markdstafford who might want to comment on this.

mikepizzo:
This is actually a really great question.

$top is a client directive that says that the client is only interested in the first n records. The service returns up to that number of records, and then the response is complete -- there is no nextLink, etc.

Separately, a service can decide the maximum number of records to return in a single response.  So if the request represents 1000 records the service might decide to return that 100 records at a time; in this case, the client's request is not complete until all 1000 records have been returned, so each page contains a nextLink that the client navigates until all 1000 records have been returned.

The $maxpagesize preference is a _hint_ to the service to specify the maximum number of records per page the client would like to receive.  The service can decide to follow or ignore this hint (i.e., if the client asks for a huge $maxpagesize the service is still allowed to return smaller pages).

These two can work in concert.  Say that the client is interested in the first 10,000 records.  The client can specify $top=10000 and the service could decide to return those records 100 at a time, each up to the last page containing a nextLink. Or, if the client specifies a Prefer header with $maxpagesize=1000, the service could decide to honor the client's maximum page size request and return 1000 records at a time, each up to the last page containing a nextLink.

trshiv:
Mike,

Thanks for your response. It would be great to see this explanation included in the specification, especially the part about indicating the $maxpagesize via the Prefer header.

I have a follow-up question. How does $skip work with $top and $maxpagesize? My understanding of $skip is that the server skips the specified number of pages and then serves the top N entries. However, if the server is free to disregard the $maxpagesize specified by the client or choose another value for it, how should the client and server sync on what pagesize is actually determined by the server?

RobDolinMS:
@trshiv I believe $skip is to skip the top N entries (not the top N pages.)

Expanding on @mikepizzo's example, if the client had the first 10,000 entries and then decided they wanted another 10,000, the client might ask for $top=10000 and $skip=100000.  

Again, $maxpagesize would be optional for the client to express their preferred maximum number of entries per response.  

/cc @markdstafford and @mikepizzo if they want to add to or correct my response.

trshiv:
[...]

PR #92, opened 2017-08-10
Clarified processing of $top and $skip by the server when a client supplies both

trshiv:
This PR relates to issue #79.


gareth-a-jones:
LGTM

PR #94, opened 2017-08-25
Fixed typo in Guidelines.md

alex-grigoras:
Fixed typo: "Being text" => "Begin text" .  It had a matching "End text", so it was definitely meant to say Begin rather than Being

### DIFF
--- a/Guidelines.md
+++ b/Guidelines.md
@@ -2103,7 +2103,7 @@ All sequence diagrams in this document are generated using the WebSequenceDiagra
 #### 17.1.1 Push notifications, per user flow
 
 ```
-=== Being Text ===
+=== Begin Text ===
 note over Developer, Automation, App Server:
      An App Developer like MovieMaker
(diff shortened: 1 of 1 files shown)


gareth-a-jones:
There's a possibility this text actually came from the diagramming tool, rather than being hand-crafted.  Clearly it is a typo, but if it's a typo in the tool output I don't want to break the round tripping.  Needs investigation.

aaronpalmer:
Note, the === Being Text === and === End Text === are invalid syntax for the WebSequenceDiagrams.com tool.  It appears that the intended usage is to copy the text between these tokens and paste that into the tool to generate the sequence diagram.  A simple visit to the WebSequenceDiagrams site would validate this behavior.  IMO this PR is benign and should be accepted.

gareth-a-jones:
Thanks for the input Aaron - merging based on this. And thanks again Alex.

PR #98, opened 2017-10-06
Master

peseantben:


gareth-a-jones:
Hi Ben - can you clarify what this is about?  In general we only accept PRs against vNext.

gareth-a-jones:
No response, so tidying.

Issue #10, opened 2016-07-19
Potential alternative to the new "Operation-Location" header


darrelmiller:
ref: https://github.com/Microsoft/api-guidelines/blob/master/Guidelines.md#1321-put

For long running operations that return 202, it is suggested to use `Operation-Location` to refer to a resource that can be monitored to determine when the requested operation is complete.  Although, not explicitly specified by the HTTP spec, it is common practice to use the `Location` header to point to this status monitoring resource. See [RFC7240](https://tools.ietf.org/html/rfc7240#section-4.1) and  [RESTful Web Services Cookbook](]http://shop.oreilly.com/product/9780596801694.do)

It would seem prudent to follow common practice and avoid creating a new non-standard header for something that is commonly done.

For the hybrid scenario where the `Location` header is used to refer to the partially created resource and the `Operation-Location` header is used to point to the status resource, an alternative would be the following response:

```
HTTP/1.1 202 Accepted
Location: https://api.contoso.com/v1.0/operations/123
Content-Location: https://api.contoso.com/v1.0/databases/db1
{
  "databaseName": "db1",
  "color": "red",
  "Status": "Provisioning",
  [ … other fields for "database" …]
}

```
This approach is consistent with it's use of 202 to indicate that the creation process has not completed.  It continues to use the `Location` header to indicate the location of the status monitoring resource.  However, it also uses the Content-Location header to indicate that there is a response body that represents the current state of the resource being created and provides a URL to the final destination of the resource.

This approach provides equivalent functionality but doesn't require inventing a new header, it follows current common practices and it avoids having to impose the rules that the hybrid 201 must a have a 
body, and it must indicate that it is incomplete, and that 202 responses shouldn't return a body.

garethj-msft:
I really like this proposal.  I'm circulating it internally to see what reaction we have to changing to this as our preferred guidance.

jmealo:
Any updates on this? (For what it's worth, I agree with the proposal.)

darrelmiller:
I guess I should write a PR.  /cc Thoughts? @garethj-msft @RobDolinMS @markdstafford

freeranger:
@garethj-msft What was the internal reaction to this proposal?

letmaik:
@darrelmiller @garethj-msft Could you please add a final statement and close issues and PRs that have been abandoned? If this issue is indeed still actively under discussion then a status update would be good.

PR #101, opened 2018-02-06
Alternate long running operation strategy

darrelmiller:
Related to #10 

This change removes the need for the custom header `Operation-Location` and uses the standard `Content-Location` header to enable the hybrid flow.  This approach ensures that all long running operations use the standard 202 status code and Location header pattern.  It also removes the need to prevent a body from being returned with a 202 response.

### DIFF
--- a/Guidelines.md
+++ b/Guidelines.md
@@ -1338,13 +1338,10 @@ In this scenario the _databases_ segment is processing the PUT operation.
 
 ```http
 HTTP/1.1 202 Accepted
-Operation-Location: https://api.contoso.com/v1.0/operations/123
+Location: https://api.contoso.com/v1.0/operations/123
 ```
 
-For services that need to return a 201 Created here, use the hybrid flow described below.
-
-The 202 Accepted should return no body.
-The 201 Created case should return the body of the target resource.
+For services that need to return a partially created response here, use the hybrid flow described below.
 
 #### 13.2.2 POST
 Services MAY enable POST requests for entity creation.
@@ -1360,7 +1357,7 @@ POST https://api.contoso.com/v1.0/databases/
 
 ```http
 HTTP/1.1 202 Accepted
-Operation-Location: https://api.contoso.com/v1.0/operations/123
+Location: https://api.contoso.com/v1.0/operations/123
 ```
 
 #### 13.2.3 POST, hybrid model
@@ -1381,13 +1378,13 @@ Accept: application/json
 }
 ```
 
-Service response says the database has been created, but indicates the request is not completed by including the Operation-Location header.
-In this case the status property in the response payload also indicates the operation has not fully completed.
+Service response says the database is being created at the URL provided in the Content-Location header, but indicates the request is not completed by including a 202 status code and Location header pointing to an operation resource.
+The response body includes the incomplete representation of the resource that will eventually exist at the URL in the Content-Location header.
 
 ```http
-HTTP/1.1 201 Created
-Location: https://api.contoso.com/v1.0/databases/db1
-Operation-Location: https://api.contoso.com/v1.0/operations/123
+HTTP/1.1 202 Accepted
+Content-Location: https://api.contoso.com/v1.0/databases/db1
+Location: https://api.contoso.com/v1.0/operations/123
 
 {
   "databaseName": "db1",
@@ -1491,7 +1488,7 @@ Services MAY choose to delete tombstones after a service defined period of time.
 #### 13.2.7 The typical flow, polling
 - Client invokes a stepwise operation by invoking an action using POST
 - The server MUST indicate the request has been started by responding with a 202 Accepted status code. The response SHOULD include the location header containing a URL that the client should poll for the results after waiting the number of seconds specified in the Retry-After header.
-- Client polls the location until receiving a 200 OK response from the server.
+- Client polls the location until receiving a response that indicates the stepwise long running operation is complete.
 
 ##### 13.2.7.1 Example of the typical flow, polling
 Client invokes the restart action:
@@ -1510,7 +1507,7 @@ The server response indicates the request has been created.
 
 ```http
 HTTP/1.1 202 Accepted
-Operation-Location: https://api.contoso.com/v1.0/operations/123
+Location: https://api.contoso.com/v1.0/operations/123
 ```
 
 Client waits for a period of time then invokes another request to try to get the operation status.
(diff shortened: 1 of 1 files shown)


markdstafford (commented):
This looks good to me overall. Should we elaborate more about the responsibilities of the `Location` and `Content-Location` headers?

markdstafford (approved):

Guidelines.md
```diff
@@ -1360,7 +1357,7 @@ POST https://api.contoso.com/v1.0/databases/
 
 ```http
 HTTP/1.1 202 Accepted
-Operation-Location: https://api.contoso.com/v1.0/operations/123
+Location: https://api.contoso.com/v1.0/operations/123
```
cleemullins, line 1360: This is a breaking change. While I like the use of a standard header here, I don't think it meets the bar for a breaking change. Stability is a feature, in and of itself. 

I know we considered the location header at the time we wrote these, but I don't recall why it was rejected.

Guidelines.md
```diff
@@ -1338,13 +1338,10 @@ In this scenario the _databases_ segment is processing the PUT operation.
...
 ```
 
-For services that need to return a 201 Created here, use the hybrid flow described below.
-
-The 202 Accepted should return no body.
-The 201 Created case should return the body of the target resource.
+For services that need to return a partially created response here, use the hybrid flow described below.
```
cleemullins, line 1344: I like the words here.

Guidelines.md
```diff
@@ -1338,13 +1338,10 @@ In this scenario the _databases_ segment is processing the PUT operation.
 
 ```http
 HTTP/1.1 202 Accepted
-Operation-Location: https://api.contoso.com/v1.0/operations/123
+Location: https://api.contoso.com/v1.0/operations/123
 ```
 
-For services that need to return a 201 Created here, use the hybrid flow described below.
-
-The 202 Accepted should return no body.
```
cleemullins, line 1346: These requirements seem important. Not clear why they are removed.

Guidelines.md
```diff
@@ -1381,13 +1378,13 @@ Accept: application/json
...
-Service response says the database has been created, but indicates the request is not completed by including the Operation-Location header.
-In this case the status property in the response payload also indicates the operation has not fully completed.
+Service response says the database is being created at the URL provided in the Content-Location header, but indicates the request is not completed by including a 202 status code and Location header pointing to an operation resource.
+The response body includes the incomplete representation of the resource that will eventually exist at the URL in the Content-Location header.
 
 ```http
-HTTP/1.1 201 Created
-Location: https://api.contoso.com/v1.0/databases/db1
-Operation-Location: https://api.contoso.com/v1.0/operations/123
+HTTP/1.1 202 Accepted
```
cleemullins, line 1385: This is a breaking change.

Guidelines.md
```diff
@@ -1381,13 +1378,13 @@ Accept: application/json
...
+Content-Location: https://api.contoso.com/v1.0/databases/db1
```
cleemullins, line 1386: This is actually **scarier** than a breaking change. You're changing the meaning of the location header, which means developers that aren't paying attention will be actively doing the wrong thing. The interaction of old+new code will be very unpredictable here.

Guidelines.md
```diff
@@ -1491,7 +1488,7 @@ Services MAY choose to delete tombstones after a service defined period of time.
 #### 13.2.7 The typical flow, polling
 - Client invokes a stepwise operation by invoking an action using POST
 - The server MUST indicate the request has been started by responding with a 202 Accepted status code. The response SHOULD include the location header containing a URL that the client should poll for the results after waiting the number of seconds specified in the Retry-After header.
-- Client polls the location until receiving a 200 OK response from the server.
+- Client polls the location until receiving a response that indicates the stepwise long running operation is complete.
```
cleemullins, line 1491: I like the words here.

cleemullins (changes requested):
I agree the changes here are probably slightly cleaner. However, in the case I would opt for stabilty-as-a-feature and avoid the churn this breaking change would cause.

Guidelines.md
markdstafford: Isn't this part of why we adopted semantic versioning for the guidelines? Putting stability as the guiding priority means that we will grind forward progress to a halt. This is a positive change in the guidelines. It warrants a bump to version 3. As before, APIs are not required to make breaking changes just so they can adopt the new guidelines - but for new APIs, wouldn't we want the right baseline in place?

In my opinion, we should be putting additive changes into 2.x of the API guidelines, but not unnecessarily blocking a 3.x of the guidelines from being created. We *could* say that we haven't gathered sufficient breaking changes to warrant a 3.x at this point, but then I would ask where we plan to track those changes, and whether we can start on 3.x-preview1 even if we choose not to release 3.x until we do get to critical mass.

Guidelines.md
markdstafford: If I understand correctly, @darrelmiller is not changing the meaning of the `Location` header. He's pointing out that `201 Created` is a misnomer when we are talking about asynchronous resource creation. A strict interpretation of HTTP would lead one to believe that the resource is done being created. In reality, the create request has been `202 Accepted`, in which case the usage of `Content-Location` is additive to the examples we had before.

Guidelines.md
darrelmiller: This PR proposes that a 202 also be used for the hybrid flow and during that scenario a partially created response is returned.  Therefore saying that a body should not be returned with a 202 would be contradictory.

Guidelines.md
darrelmiller: The meaning of Location header is dependent on the response code.  A 202 response code means the Location header points at the status monitor resource (what we call operation). A 201 status code means the Location points to the created resource.  For 3xx status codes Location means something else.

Therefore because I'm also changing the status code that is returned for a hybrid flow, the meaning of Location header would be consistently the same as what Operation-Location currently means.  Old client code talking to a new API would be missing the Operation-Location header in a 202 response and new client code talking to an old service would see a 201 created and not be aware that it is a hybrid flow LRO.

AkJo:
Do we have any consensus on this? We are building a new API and were wondering if we should follow this new approach.

glennblock:
If there's concern on a breaking change, why not introduce a header or some way for clients to opt-in to the alternative behavior? The approach suggested seems like it is in-line with how the HTTP spec defines using the 202 header. On a different note, did anyone consider submitting the new header to the IETF, in order to get some discussion around its introduction?

PR #120, opened 2018-10-02
Change MAY to might

akatakritos:
In context, the word `may` here is not describing an optional activity for the implementer (the normal use of the capitalized word per [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) ). Rather it is describing a possible consequence as justification for the guideline.

It's saying "because doing so [might, could] make the value unreadable".


gareth-a-jones:
Thanks!

PR #125, opened 2018-10-05
Remove implication that code is an enumerated string in Response Type…

### LINKED ISSUE #100: camelCase vs PascalCase for string enumerations
In documentation at https://github.com/Microsoft/api-guidelines/blob/master/Guidelines.md#710-response-formats we can see
...
Property | Type | Required | Description
-- | -- | -- | --
code | String (enumerated) | ✔ | One of a server-defined set of error codes.
...
```csharp
{
  "error": {
    "code": "BadArgument",
    "message": "Previous passwords may not be reused",
    "target": "password",
    "innererror": {
      "code": "PasswordError",
      "innererror": {
        "code": "PasswordDoesNotMeetPolicy",
        "minLength": "6",
        "maxLength": "64",
        "characterTypes": ["lowerCase","upperCase","number","symbol"],
        "minDistinctCharacterTypes": "2",
        "innererror": {
          "code": "PasswordReuseNotAllowed"
        }
      }
    }
  }
}
```
Here property `code` is PascalCase'd as "BadArgument". and property `characterTypes` as camelCase'd.

This seems a bit confusing, since `characterTypes` seems to be enumeration of some sort, so why do they use camelCase but `code` use PascalCase?

Is there a guideline from Microsoft what casing enumerated string values should have? Is it PascalCase that is recommended?

Thanks.

juliegoat:
…s section

Resolves #100 

I know it's a small thing, but I'm trying to get my feet wet :)

### DIFF
--- a/Guidelines.md
+++ b/Guidelines.md
@@ -550,7 +550,7 @@ Property | Type | Required | Description
 
 Property | Type | Required | Description
 -------- | ---- | -------- | -----------
-`code` | String (enumerated) | ✔ | One of a server-defined set of error codes.
+`code` | String | ✔ | One of a server-defined set of error codes.
 `message` | String | ✔ | A human-readable representation of the error.
 `target` | String |  | The target of the error.
 `details` | Error[] |  | An array of details about specific errors that led to this reported error.


gareth-a-jones:
Thanks Julie - I'm taking this as it is an improvement to remove confusion.

PR #127, opened 2018-10-08
Update README.md

alljp:
Improve readability


alljp:
I think it is better to show enhancement, by putting in bold letters, instead of caps. 
As, in this case, `...REST APIs SHOULD follow consistent...`, 'should' in caps right next to 'REST APIs' look odd. If accepted, I can make the necessary changes in the other files too.

gareth-a-jones:
Hi alljp - thanks for your issue and PR - this use of inline CAPS is pretty standard in docs that conform to RFC2119.  I think we'll leave this as it is.

alljp:
Cool. Thanks for the reply.

PR #139, opened 2018-10-16
Correct the spelling of predate

jacobherrington:


gareth-a-jones:
Thanks for the catch.

PR #140, opened 2018-10-17
Reworded compliance text to fix grammar

daredevilgeek:
Reworded from "for compliance sake" to "for the sake of compliance"


gareth-a-jones:
Thanks!

PR #141, opened 2018-10-17
Clarify sentence

MacroChip:


gareth-a-jones:
Thanks, that is better. Less 'to's.

PR #142, opened 2018-10-17
Minor Typographical Fix

aadibajpai:
As **guidelines** and not a singular **guideline**, it should be plural to encompass all of them and not be limited to a single one. I'm sorry but it struck my eye while going through them so I decided to send a fix.


ctrlaltvikas (approved):

gareth-a-jones:
Thanks for the fix, but this file got rewritten, so it is no longer applicable.

PR #146, opened 2018-10-21
Added Guidelines.md inside README.md

Devil1729:
appended all the contents of Guidelines.md into README.md


gareth-a-jones:
We're not going to accept this as we feel that the two files have different purposes and allow us to add further files over time if we so wish.

PR #40, opened 2016-07-22
Replace references to RFC2616

darrelmiller:
Updates as per #6

PR #149, opened 2018-10-25
Add HTTP Status Code Definitions resource link

jonathanfrancisco:


MattDavisRM:
Seems like this should be a reference to RFC 7231 as it has replaced RFC 2616.  Appears there has already been prior work to remove the 2616 references.

#6 
#40

MattDavisRM (changes requested):
Seems like this should be a reference to RFC 7231 as it has replaced RFC 2616. Appears there has already been prior work to remove the 2616 references.
#6 
#40

PR #150, opened 2018-10-25
Update README.md

mriduladvani:


gareth-a-jones:
Thanks for the suggestion, but it's grammatically incorrect.

Issue #105, opened 2018-02-22
Put the Guidelines.md into the Readme.md


ckuijjer:
It's a minor issue, but when I search the web for _microsoft rest api guidelines_ I end up in the github repository and I'll have to manually click on the `Guidelines.md` to start reading. I think it's a nice idea to simply put the content of the guidelines into the `README.md`.

mgefimov:
+1

AndrewCraswell:
+1

TomasHubelbauer:
This would be a really nice improvements and one of the PRs is even approved, although that is from fours months ago and since then there have presumably been more changes. Hopefully it gets updated and merged at some point.

TomasHubelbauer:
A good suggestion by @tomkerkhove in #178 relating to preserving existing links:

> Concerning breaking linkes - That can easily be fixed by keeping the file with a pointer to new location.

Just keep `GUIDELINES.md`, delete content, keep only headers and underneath each header provide a link to the README or to the actual GitHub Pages site as per #178.

gareth-a-jones:
Thanks for the feedback and the votes but we've decided not to accept this suggestion.  We feel like the two files serve different purposes and allow us to add separate extra guidelines in future is we so choose.

PR #152, opened 2018-10-26
Included Guidelines into README

terumiisobe:
Solving the issue #105.


ctrlaltvikas (approved):

gareth-a-jones:
We are not going to accept this, as we feel the two files have separate purposes, and over time allow us to add extra files if we so wish.

PR #153, opened 2018-10-27
Fidel Committed here

fidellim:


gareth-a-jones:
Thanks for the submission Fidel, but I think the example is OK as it is.

PR #156, opened 2018-10-29
removed a red patch at the bottom

PradeepTheDev:
removed a red patch at the bottom  probably a whitespace


gareth-a-jones:
Pradeep, thanks for the submission, but I can't tell what value this PR is trying to add?

PR #160, opened 2018-10-30
Improved guidelines introduction

leoriviera:


gareth-a-jones:
Thanks for the input Leo, but I think we're happy with the intent of the introduction.  This proposal changes the sense of what we were trying to say.

PR #161, opened 2018-10-31
Added a book on fundamentals for RESTful API.

runandrerun:
Book added.


gareth-a-jones:
Thanks for the book recommendation.  

Could you make a couple of changes to the PR please?
1) Drop the spacing fixes that are causing merge errors.
2) Provide a cleaner URL without all the tracking stuff.

thanks!

PR #172, opened 2019-02-02
Polling on operation status to terminal state instead of 200 response

johanste:
@darrelmiller pointed out that Section 13.2.7. "The typical flow, polling" is unclear. The description says to poll on operations resource until a 200 response is received (which I believe is incorrect)

The example continues polling until a 200 response with a terminal status is received (which I believe is correct).


darrelmiller (approved):

gareth-a-jones:
Thanks for the PR - good catch.

PR #191, opened 2020-03-31
Added v3.0 of the Azure REST API Guidelines

adrianhall:
Per email discussion, this moves the current version of the Azure REST API Guidelines (converted from Word to Markdown, with minor spelling mistake corrections and formatting adjustments) into the api-guidelines repository.


johngossman:
We need to figure out what to do about the link to the Azure Stack versioning guidelines. I'd delete that section for now.

johngossman:
There is a link to an internal sharepoint site that is used for communicating deprecation. I'd suggest we just tell people to contact the API review board, which we do elsewhere in the deprecation guidance. That whole section needs to be read very closely that we aren't leaking internal addresses or processes that externals can't use, but I think it is nice to show customers what we do, and also for partners who are operating services, it is relevant...assuming they have a way to reach in to our comms and groups.

azure/Guidelines.md
```diff
@@ -0,0 +1,310 @@
...
+* **Authentication**.  Guidance around the authentication and its use.
+* **Encoding**.  Guidance around the use of JSON encoding and standards.
+* **Error responses**.  A standard format for error responses is closed.
+* **Dates and times**.  Detailed guidance around date and time formats and encoding is provided.
+* **Relationship to OData**.  Guidance around the relationship between OData and REST APIs.
+* **JSON representation**.  JSON is preferred over XML and other formats.
+* **Long running operations**.  The OneAPI guidelines are updated and simplified over the previous Azure guidelines.
+* **Push notifications via Webhooks**.
+
+### Updating from previous versions of the Azure API guidelines
```
johngossman: We should just delete this as its refering to an old version from 4 years ago

azure/Guidelines.md
```diff
@@ -0,0 +1,310 @@
...
+
+#### Asynchronous operations
+
+The OneAPI guidelines for Long Running Operations guidelines are an updated, clarified and simplified version of the Asynchronous Operations guidelines from the 2.1 version of the Azure API guidelines. Unfortunately, to generalize to the whole of Microsoft and not just Azure, the board decided to rename the HEADER used in the operation from `Azure-AsyncOperation` to `Operation-Location`. Services can and **SHOULD** support both `Azure-AsyncOperation` and `Operation-Location` HEADERS, even though they are redundant so that existing SDKs and clients will continue to operate. Clients that call these services **SHOULD** look for both HEADERS and prefer the `Operation-Location` version. Both HEADERS **MUST** return the same value.
+
+### Additional guidance for Azure Resource Manager resource providers
+
+Teams building ARM RPs MUST follow the additional guidance in the ARM RPC and related documents. These documents can be found here.
+
+* [Azure Resource Manager Documents][2] (Internal only)
```
johngossman: I think these are public on Github now

azure/Guidelines.md
```diff
@@ -0,0 +1,310 @@
...
+   Content-Location: http://00000000-0000-0000-C000-000000000046-account1.blobstore.azure.net/container1/blob2
+   ```
+
+## Versioning
+
+All Azure APIs **MUST** support explicit versioning.  It's critical that clients can count on services to be stable over time, and it's critical that Azure services can add features and make changes.
+
+The OneAPI guidelines offer a couple of different options on how to specify an API version and guidance on what constitutes a breaking change.  This section of the Azure API guidelines describes which options are required of Azure services as well as some guidance about deprecation policy.  There is also a section about additional versioning practices necessary to support Azure Stack and Azure compatibility.
+
+Further discussion on versioning principles and criteria can be found in the [Azure API versioning guidelines][4] (Internal only).
```
johngossman: Let's delete this for now and then fold in the versioning guidance

azure/Guidelines.md
```diff
@@ -0,0 +1,310 @@
...
+
+Even though we recommend clients ignore new fields, there are many libraries and clients that are strict. Therefore, Azure services **MUST** update the version number of their API even when adding optional fields. In fact, servers should be as strict as possible. Ignoring a field can result in the API accepting content that containered a typo or an element at the wrong level of nesting. If this missing field changes the semantics (for example, we have seen cases where security settings were misplaced and ignored, leaving the resources more exposed than intended) this can be a huge and hard to discover error.
+
+### Group versioning in Azure and Azure Stack
+
+Azure Stack allows customers and hosters to deploy their own small versions of Azure and upgrade it at a different pace than Azure. In order to make it possible to write an application or SDK that targets both Azure and Azure Stack, additional versioning policy is necessary. This guidance is still being worked on, but the latest can be found here: [Azure and Azure Stack Versioning][5].
+
```
johngossman, line 104: Let's just tell people to contact the Azure Stack team for more info

azure/Guidelines.md
```diff
@@ -0,0 +1,310 @@
...
+### Step 5: Communicate widely internally
+
+The decision should be communicated widely within Azure.  Mail describing the decision should be sent to the Azure API review board, WW – Communities and other appropriate aliases. There should be another short (2-5 working days) feedback period before moving on to the next step.
+
+### Step 6: Communicate externally, starting the 12 month clock
+
+The decision next **MUST** be communicated to customers and partners. The Service Notices site gives instructions for communicating this and other service changes. The Playbook contains a link to a submission form to request e-mail be sent.
+
+https://microsoft.sharepoint.com/teams/azurecomms/SitePages/Email-Request.aspx
```
johngossman: Let's change this to "contact the Azure API review board" on how to start communicate externally for now.

azure/Guidelines.md
adrianhall: Ack = removing the whole section.

azure/Guidelines.md
adrianhall: I'll dig - if anyone happens to know where, let me know.  I'd rather include public docs.

azure/Guidelines.md
adrianhall: Is that azureapirbcore or some other review board?

johngossman: Lets just say Azure API review board and make it harder for spammers to find the alias...

azure/Guidelines.md
adrianhall: The Azure API versioning guidelines is on an "expired" site - so I don't have access to it.  I'll remove the paragraph, and we can re-visit versioning as one of the topics.

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
+# Microsoft Azure REST Design Guidelines
+
+## History
+
+<dl>
+<dd>2020-Mar-31 v3.0</dd>
+<dt>Version 3.0 of the Azure REST API Guidelines was copied from Sharepoint
```
gareth-a-jones: This process story seems pretty irrelevant to a public doc.

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+into Markdown and uploaded to GitHub as the basis for future improvements.</dt>
+</dl>
+
+## Introduction
+
+The Azure REST API guidelines are an extension of the Microsoft-wide [OneAPI guidelines][1] (which historically drew heavily from an earlier version of the Azure REST API guidelines). Readers of this document are assumed to be also reading the [OneAPI guidelines][1] and be familiar with them.  Azure guidance is a superset of OneAPI guidelines and services should follow them *except* where this document outlines specific differences or exceptions to those guidelines. This document does contain additional Azure-specific guidance and additional details.
```
gareth-a-jones: We don't use the term OneAPI in public docs.  Need to decide what short name we're going to use for the core doc.

gareth-a-jones: Not really relevant.

gareth-a-jones: Not OneAPI

gareth-a-jones: Not OneAPI

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+
+The OneAPI guidelines are available on Github at [https://github.com/microsoft/api-guidelines][1].
```
gareth-a-jones: This seems a bit odd  now they're in the same repo

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+
+Highlights of the OneAPI guidelines include:
```
gareth-a-jones: This whole section is redundant now they share a repo - suggest completely remove it.

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+* **Dates and times**.  Detailed guidance around date and time formats and encoding is provided.
+* **Relationship to OData**.  Guidance around the relationship between OData and REST APIs.
+* **JSON representation**.  JSON is preferred over XML and other formats.
+* **Long running operations**.  The OneAPI guidelines are updated and simplified over the previous Azure guidelines.
+* **Push notifications via Webhooks**.
+
+#### Asynchronous operations
+
+The OneAPI guidelines for Long Running Operations guidelines are an updated, clarified and simplified version of the Asynchronous Operations guidelines from the 2.1 version of the Azure API guidelines. Unfortunately, to generalize to the whole of Microsoft and not just Azure, the board decided to rename the HEADER used in the operation from `Azure-AsyncOperation` to `Operation-Location`. Services can and **SHOULD** support both `Azure-AsyncOperation` and `Operation-Location` HEADERS, even though they are redundant so that existing SDKs and clients will continue to operate. Clients that call these services **SHOULD** look for both HEADERS and prefer the `Operation-Location` version. Both HEADERS **MUST** return the same value.
```
gareth-a-jones: not OneAPI

gareth-a-jones: normative language - don't need "can and'

gareth-a-jones: Reference to mysterious 'the board' seems weird in the context of an external doc.   Suggest just 'The header was renamed'

gareth-a-jones: My experience with Azure LROs is that there are MANY other differences, not just the header name.   Typically Azure LROs interpret the HTTP return codes completely differently to the core guidelines intent and for example Office.  e.g. Azure workload return 202 accepted repeatedly, whereas Office and core guidelines suggest just one 202 and then operations endpoint always returns 200.   In short, I think this section needs a lot more fleshing out and the Azure intent clarifying.

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+
+### Additional guidance for Azure Resource Manager resource providers
+
+Teams building ARM RPs MUST follow the additional guidance in the ARM RPC and related documents. These documents can be found here.
```
gareth-a-jones: Define RP in text before defining acronym.

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+
+* [Azure Resource Manager Wiki][2] (Internal only)
+* [Azure Resource Provider Contract][3]
+
+ARM RPs are a CEC requirement for Azure Services and ARM RP review is another mandatory review. Some of the guidance overlaps with general API review, but passing one review will generally make the other one go very quickly.
```
gareth-a-jones: Define CEC

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+
+## Swagger to describe API
+
+All Services **MUST** provide Swagger that describes their service. Swagger is a key element of the Azure SDK plan and essential to improving the usability and discoverability of services. Swagger is a CEC requirement for Azure Services.
```
gareth-a-jones: Swagger is a trademarked term.  We should be using OAS terminology.  As we publish the doc we need to snap to correct terms.

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+
+## URL structure
+
+In addition to the URL structure guidance in the OneAPI guidelines, Azure has specific guidance about service exposure for multi-tenant services
```
gareth-a-jones: Not OneAPI

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+
+### URL structure
+
+All services **MUST** expose their service to developers via the following URL pattern:
+
+```
+http(s)://<service>.azure.net/<unit-of-multi-tenancy>/<service-defined-root>
```
gareth-a-jones: DNS name seems out of date.  e.g. management.azure.com

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+   Or, with the GUID format:
+
+   ```
+   200 OK
+   Content-Location: http://00000000-0000-0000-C000-000000000046-account1.blobstore.azure.net/container1/blob2
+   ```
+
+## Versioning
+
+All Azure APIs **MUST** support explicit versioning.  It's critical that clients can count on services to be stable over time, and it's critical that Azure services can add features and make changes.
```
gareth-a-jones: what does 'support' mean in this context.  'use'

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+
+The OneAPI guidelines offer a couple of different options on how to specify an API version and guidance on what constitutes a breaking change.  This section of the Azure API guidelines describes which options are required of Azure services as well as some guidance about deprecation policy.  There is also a section about additional versioning practices necessary to support Azure Stack and Azure compatibility.
```
gareth-a-jones: OneAPI

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+
+### Specifying the version in Azure
+
+The OneAPI guidelines give two options for how services and clients communicate the version: a url segment and a query parameter. Azure services **MUST** use the api-version query parameter. For example:
```
gareth-a-jones: OneAPI

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+
+```
+GET http://blobstore.azure.com/foo.com/acct1/c1/blob2?api-version=1.0
```
gareth-a-jones: Is a 1.0 version number actually acceptable?

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+PUT http://blobstore.azure.com/foo.com/acct1/c1/b2?api-version=2014-12-07
+POST http://blobstore.azure.com/foo.com/acct1/c1/b2?api-version=2015-12-07
+```
+
+### Breaking changes in Azure
+
+A breaking change is any change in the API that may cause client or service code making the API call to fail. Obvious examples of such a change are the removal of an endpoint, adding or removing a required field or changing the format of the body (from XML to JSON for example).
+
+Even though we recommend clients ignore new fields, there are many libraries and clients that are strict. Therefore, Azure services **MUST** update the version number of their API even when adding optional fields. In fact, servers should be as strict as possible. Ignoring a field can result in the API accepting content that containered a typo or an element at the wrong level of nesting. If this missing field changes the semantics (for example, we have seen cases where security settings were misplaced and ignored, leaving the resources more exposed than intended) this can be a huge and hard to discover error.
```
gareth-a-jones: strict isn't the correct word here.   There are many clients that do not abide by this recommendation might be a better phrase.  'strict' im,plies they are doing something more correct which is hardly the case,

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+
+At a high level, changes to the contract of an API constitute a breaking change. Changes that impact backwards compatibility of an API is also considered a breaking change. Teams MAY define backwards compatibility as their business needs require. For example, Azure defines the addition of a new JSON field in a response to be not backwards compatible. Anything that would violate the Principle of Least Astonishment is considered a breaking change in Azure. Below are some concrete examples of what constitutes a breaking change. In the below breaking change scenarios, the API version must be changed.
```
gareth-a-jones: This sentence seems to contradict the one immediately following it.  Can Teams in Azure define backwards compatibility to be different to Azure's definition?

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+
+#### Existing property is removed
+
+If a property called `foo` that was present in v1 of the API needs to be removed, it must be done in a newer API version.
+
+#### New property added to response
+
+If a new property/field is added to the response an API, the GET-PUT pipeline will be broken. Consider the case where from portal a customer updates the value of a new property "A". Another customer does a GET of this resource using the SDK. The SDK will ignore the property since it does not understand it. From the SDK, the customer does a PUT using the model that was returned from the GET. This will overwrite the change made by the first customer from the portal.
```
gareth-a-jones: response from/response of

gareth-a-jones: a portal/the portal if so, which.

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+
+### Version discovery
+
+Simpler clients may be hardcoded to a single version of a service. Since Azure services offer each version for a well-known period of time, a client that’s regularly maintained can be always operational without further complexity as long as during regular maintenance the client is moved forward to new versions in advance of older ones being retired.
+
+API version discovery is needed when either a given hosted service may expose a different API version to different clients (e.g. latest API version only available in certain regions or to certain tenants) or the service itself may exist in different instances (e.g. a service that may be run on Azure or hosted on-premises). In both of those cases clients may get ahead of services in the API version they use. In might also be possible for a client version to ship ahead of its corresponding service update, leading to the same situation. Lastly, version discovery is useful for clients that want to warn operators that an API they depend on may expire soon.
```
gareth-a-jones, line 174: would you really want teams to do this to their customers?

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+2.	Services **MUST** include the `api-supported-versions` header, containing a comma-separated list of versions conforming to the Azure versioning scheme. This list must include all group versions as well as all major-minor versions supported by the target resource. For cases where no specific version applies (e.g. sometimes the root resource), the list still must contain the group versions supported by the service.
+3.	If a given service supports versions of the API that are known to be planned for deprecation in a year or less, it must include those versions (group and major.minor) in the `api-deprecated-versions` header.
+4.	In addition to the functionality described here, services may support HTTP `OPTIONS` requests for other purposes such as further discovery, CORS, etc.
+5.	Services may allow unauthenticated HTTP `OPTIONS` requests. When doing so authors need to consider whether HTTP `OPTIONS` requests against non-existing resources result in 404s and whether that is leaking sensitive information. Certain scenarios, such as support for CORS pre-flight requests, require allowing unauthenticated HTTP `OPTIONS` requests.
```
gareth-a-jones: **MAY**

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+6.	For services that do rolling updates where there is a point in time where some front-ends are ahead of others version-wise, all front-ends should report the previous version as the latest version until the rolling update covers all instances and only then switch over to reporting the new latest version. This ensures that clients will not detect a version and then get load-balanced into a front-end that does not support it yet.
+7.	If using OData and addressing an expanded resource, the HTTP `OPTIONS` request should return the group versions that are supported across the expanded set.
```
gareth-a-jones: **SHOUL** but really, only a shoudl here, I woudl have expected a MUST.

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+
+Azure does not have a single SLA for how long we will support all services. However, we have published expectations such as [the Azure Cloud Lifecycle FAQ][6].  The most relevant section of the document:
+
+> Azure Cloud Services will support no fewer than the latest two SDK versions for deploying new Cloud Services. Microsoft will provide notification 12 months before retiring a SDK in order to smooth the transition to a supported version.
+
+In practice, we have found this is a bare minimum of how long service endpoints must be supported. Services with any significant usage **SHOULD** expect to run until customers are no longer using them, which can be 10 years or more.
+
+Service teams **MUST** contact the Azure API review board before communicating the deprecation externally to customers and partners (which starts the 12 month clock).
+
+Refer to the Azure deprecation policy for more details.
```
gareth-a-jones, line 222: link

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+
+#### Special case for pre-release and beta APIs
+
+Pre-release and beta APIs are not covered by the normal API deprecation policy. Each team providing a preview API **SHOULD** communicate to customers what the policy is going to be for support and deprecation, even if that policy is “we may remove this at any time”. The Azure REST API Guidelines cover pre-release API versions. To summarize that section, they should be marked with a version tag like `2013-03-21-Preview`.
```
gareth-a-jones, line 226: I don't understand this sentence.  D'you mean to say 'This document is also applicable to pre-release APIs"?

gareth-a-jones, line 226: I'm not sure what you're summarizing here.

azure/Guidelines.md
```diff
@@ -0,0 +1,259 @@
...
+
+Though services may set their own deprecation policy for pre-release APIs, they should monitor these endpoints closely and consider following the normal deprecation policy and process. ** *Customers have suffered downtime because of deprecation of preview APIs* **. In some cases the code was written by consultants or employees without any awareness on behalf of the customer.
```
gareth-a-jones: Not sure what this sentence is trying to say.  IT seems to define customer employees as not the customer which I don't' think is helpful.

gareth-a-jones (changes requested):

azure/Guidelines.md
adrianhall: Agreed - intent was not to do functional changes in the guidelines, however, without approval from the board.  I'd like to revisit this as one of the topics during the discussions in a couple of weeks.

azure/Guidelines.md
adrianhall: Yes, it is - cognitive services has standardized on this.

azure/Guidelines.md
adrianhall: I've replaced it with a table and notes.  I've made this release notes a little "obscure"

azure/Guidelines.md
adrianhall: Agreed - and linked elsewhere.  I'll remove it.

azure/Guidelines.md
adrianhall: Agreed,

azure/Guidelines.md
adrianhall: Ack.

adrianhall: Ack.

azure/Guidelines.md
adrianhall: Ref: @johngossman  - what is CEC in this context?

azure/Guidelines.md
adrianhall: What is the correct terminology here?

azure/Guidelines.md
adrianhall: I;ve reworded, but I don't like the new wording - this is caused by the mash-up of the updated breaking change guidance into the new doc.  We should revisit this during the reconciliation and update meetings.

azure/Guidelines.md
adrianhall: I agree, and I don't think that is what the drafter meant.  @johngossman - would you be ok removing the "Teams MAY define backwards compatibility as their business needs require"?

azure/Guidelines.md
adrianhall: Clarified.

azure/Guidelines.md
adrianhall: I wouldn't :-)  I'll add this to the list of things to discuss with the board.

azure/Guidelines.md
adrianhall: Another topic to bring up :-)

azure/Guidelines.md
adrianhall: Sentence does not add substance to the statement, so I have removed it.

azure/Guidelines.md
adrianhall: I see it is "OpenAPI" - do we have a location where we define the swagger + microsoft definitions (that would be an appropriate link here).  Should we link to the open API hub?

azure/Guidelines.md
adrianhall: There are bunches of services that had their DNS grandfathered in.  All new services are based on azure.net unless there is a significant reason why not.

azure/Guidelines.md
adrianhall: Ref: @johngossman Is there a link I can link to for internal teams here?

azure/Guidelines.md
adrianhall: Versioning is one of those areas that need significant work.  I'd leave the update (since I know it is going to be updated) until the board discussion.

azure/Guidelines.md
gareth-a-jones: Open API Specification (OAS) is the artefact that replaces a swagger doc.  OpenAPI is the org.   @darrelmiller  is best placed to provide the right text here as he's on the board.

adrianhall: Changed it to "OpenAPI Specification" per the openapi website.

azure/Guidelines.md
adrianhall: Fixed.

azure/Guidelines.md
adrianhall: Fixed.

azure/Guidelines.md
adrianhall: Fixed.

azure/Guidelines.md
```diff
@@ -24,11 +24,11 @@ ARM RPs are a CEC requirement for Azure Services and ARM RP review is another ma
 
 ## Swagger to describe API
 
-All Services **MUST** provide Swagger that describes their service. Swagger is a key element of the Azure SDK plan and essential to improving the usability and discoverability of services. Swagger is a CEC requirement for Azure Services.
+All Services **MUST** provide as [OpenAPI Specification] that describes their service. The OpenAPI Specification is a key element of the Azure SDK plan and essential to improving the usability and discoverability of services.
```
gareth-a-jones: an?

azure/Guidelines.md
```diff
@@ -24,11 +24,11 @@ ARM RPs are a CEC requirement for Azure Services and ARM RP review is another ma
 
 ## Swagger to describe API
```
gareth-a-jones: Header too

azure/Guidelines.md
```diff
@@ -87,11 +87,11 @@ http(s)://<tenant-id>-<service-defined-root>.<service>.azure.net
 
 All Azure APIs **MUST** use explicit versioning.  It's critical that clients can count on services to be stable over time, and it's critical that Azure services can add features and make changes.
 
-The OneAPI guidelines offer a couple of different options on how to specify an API version and guidance on what constitutes a breaking change.  This section of the Azure API guidelines describes which options are required of Azure services as well as some guidance about deprecation policy.  There is also a section about additional versioning practices necessary to support Azure Stack and Azure compatibility.
+The Microsoft API guidelines offer a couple of different options on how to specify an API version and guidance on what constitutes a breaking change.  This section of the Azure API guidelines describes which options are required of Azure services as well as some guidance about deprecation policy.  There is also a section about additional versioning practices necessary to support Azure Stack and Azure compatibility.
```
gareth-a-jones: The docs is called the Microsoft REST API guidelines - we have lots of other API guidelines.

azure/Guidelines.md
```diff
@@ -0,0 +1,243 @@
+# Microsoft Azure REST Design Guidelines
+
+## History
+
+| Date | Version | Notes |
+| 2020-Mar-31 | v3.1 | Merge of Azure guidelines with breaking changes update |
```
cleemullins: Consider adding something to the effect of "1st public release of guidelines" .

azure/Guidelines.md
cleemullins: Why is the "Http(s)" there? Wouldn't this *always* be https?

azure/Guidelines.md
```diff
@@ -0,0 +1,243 @@
...
+When services produce URLs in response headers or bodies, they **MUST** use a consistent form – either always a GUID for tenant identifier or always a single verified domain - regardless of the URL used to reach the resource.
+
+### Direct endpoint URLs
+
+In addition to the required format above, services **MAY** also choose to expose direct endpoint for performance or routing reasons.  The direct endpoint should be discoverable by clients, to ensure that developers are presented with a consistent pattern for accessing Azure services.
+
+The format of the root of the direct endpoint **MUST** be as follows:
+
+```
+http(s)://<tenant-id>-<service-defined-root>.<service>.azure.net
```
cleemullins: Same question throughout: https is required, and this nomenclature seems to make it optional.

azure/Guidelines.md
```diff
@@ -0,0 +1,243 @@
...
+   Or, with the GUID format:
+
+   ```
+   200 OK
+   Content-Location: http://00000000-0000-0000-C000-000000000046-account1.blobstore.azure.net/container1/blob2
+   ```
+
+## Versioning
+
+All Azure APIs **MUST** use explicit versioning.  It's critical that clients can count on services to be stable over time, and it's critical that Azure services can add features and make changes.
```
cleemullins: Consider removing the editorial. It's already in the parent doc.

azure/Guidelines.md
cleemullins: Grammar nits: Consider removing "a couple of" and "as some". They're just extra words that don't add anything.

azure/Guidelines.md
```diff
@@ -0,0 +1,243 @@
...
+
+The Microsoft API guidelines offer a couple of different options on how to specify an API version and guidance on what constitutes a breaking change.  This section of the Azure API guidelines describes which options are required of Azure services as well as some guidance about deprecation policy.  There is also a section about additional versioning practices necessary to support Azure Stack and Azure compatibility.
+
+### Specifying the version in Azure
+
+The Microsoft API guidelines give two options for how services and clients communicate the version: a url segment and a query parameter. Azure services **MUST** use the api-version query parameter. For example:
```
cleemullins: As I think Gareth has already commented, "Microsoft REST API Guidelines". As a company we seem to have an infinite amount of (sometimes conflicting) API guidelines... :)

azure/Guidelines.md
```diff
@@ -0,0 +1,243 @@
...
+
+```
+GET http://blobstore.azure.com/foo.com/acct1/c1/blob2?api-version=1.0
+PUT http://blobstore.azure.com/foo.com/acct1/c1/b2?api-version=2014-12-07
+POST http://blobstore.azure.com/foo.com/acct1/c1/b2?api-version=2015-12-07
```
cleemullins: HTTPS for all of these. I don't think we should ever offer up "http" examples.

azure/Guidelines.md
```diff
@@ -0,0 +1,243 @@
...
+
+#### Existing property is removed
+
+If a property called `foo` that was present in v1 of the API needs to be removed, it must be done in a newer API version.
+
+#### New property added to response
+
+If a new property/field is added to the response of an API, the GET-PUT pipeline will be broken. Consider the case where a customer updates the value of a new property "A" from the Azure portal. Another customer does a GET of this resource using the SDK. The SDK will ignore the property since it does not understand it. From the SDK, the customer does a PUT using the model that was returned from the GET. This will overwrite the change made by the first customer from the portal.
```
cleemullins, line 111: This seems to be increasingly at-odds with standard web development practices. I'm not sure how long Azure will be able to hold the line on this.

azure/Guidelines.md
```diff
@@ -0,0 +1,243 @@
...
+
+### Version discovery
+
+Simpler clients may be hardcoded to a single version of a service. Since Azure services offer each version for a well-known period of time, a client that’s regularly maintained can be always operational without further complexity as long as during regular maintenance the client is moved forward to new versions in advance of older ones being retired.
+
+API version discovery is needed when either a given hosted service may expose a different API version to different clients (e.g. latest API version only available in certain regions or to certain tenants) or the service itself may exist in different instances (e.g. a service that may be run on Azure or hosted on-premises). In both of those cases clients may get ahead of services in the API version they use. In might also be possible for a client version to ship ahead of its corresponding service update, leading to the same situation. Lastly, version discovery is useful for clients that want to warn operators that an API they depend on may expire soon.
+
+Azure services **SHOULD** support API version discovery.  If they support it:
```
cleemullins, line 176: As a random point, I really like this section and the features it describes. Are there libraries to help with this, or is it all an exercise to the reader?  This seems like something that deserves a speclet (or even a DRAFT RFC!) on it's own.

cleemullins (commented):
This is really great. Awesome to see this moving forward.

azure/Guidelines.md
```diff
@@ -0,0 +1,243 @@
...
+
+#### Special case for pre-release and beta APIs
+
+Pre-release and beta APIs are not covered by the normal API deprecation policy. Each team providing a preview API **SHOULD** communicate to customers what the policy is going to be for support and deprecation, even if that policy is “we may remove this at any time”. The Azure REST API Guidelines cover pre-release API versions. To summarize that section, they should be marked with a version tag like `2013-03-21-Preview`.
+
+Though services may set their own deprecation policy for pre-release APIs, they should monitor these endpoints closely and consider following the normal deprecation policy and process. ** *Customers have suffered downtime because of deprecation of preview APIs* **.
+
+<!-- Links -->
+[1]: https://github.com/microsoft/api-guidelines
+[RFC2557]: http://www.ietf.org/rfc/rfc2557.txt
```
cleemullins: https

azure/Guidelines.md
```diff
@@ -0,0 +1,243 @@
...
+
+<!-- Azure ARM Links -->
+[2]: https://aka.ms/armwiki
+[3]: https://github.com/Azure/azure-resource-manager-rpc
+
+<!-- Open API Spec -->
+[OpenAPI Specification]: https://github.com/Azure/adx-documentation-pr/wiki/Getting-started-with-OpenAPI-specifications
+
+<!-- Versioning Guidelines -->
+[6]: http://support.microsoft.com/gp/azure-cloud-lifecycle-faq
```
cleemullins: https. :)

azure/Guidelines.md
adrianhall: Yes.

azure/Guidelines.md
adrianhall: ack

azure/Guidelines.md
adrianhall: Yes. (although that was a direct copy from the guidelines)

azure/Guidelines.md
adrianhall: Adjusting throughout

azure/Guidelines.md
adrianhall: It's certainly a topic that can be visited during the update meetings that are planned.

azure/Guidelines.md
adrianhall: I wrote a javascript library for this a couple of years ago, but never got around to putting it into practice.  The reality is that I locked the code down to a specific API and then depended on the contract not changing.

azure/Guidelines.md
adrianhall: I didn't like the way this was worded - so I word-smithed.  I think it displays intent better now.

azure/Guidelines.md
adrianhall: Common Engineerign Criteria :-) - got it.

azure/Guidelines.md
KrzysztofCwalina: I think a central Azure-wide group should define, review, and approve the compatibility bar. Incompatibilities hurt the Azure brand (i.e. all services), but benefit (and that is debatable too) only/mainly the individual service doing the break.

azure/Guidelines.md
```diff
@@ -0,0 +1,241 @@
...
+
+At a high level, changes to the contract of an API constitute a breaking change. Changes that impact backwards compatibility of an API is also considered a breaking change. Teams MAY define backwards compatibility as their business needs require. For example, Azure defines the addition of a new JSON field in a response to be not backwards compatible. Anything that would violate the Principle of Least Astonishment is considered a breaking change in Azure. Below are some concrete examples of what constitutes a breaking change. In the below breaking change scenarios, the API version must be changed.
+
+#### Existing property is removed
+
+If a property called `foo` that was present in v1 of the API needs to be removed, it must be done in a newer API version.
```
KrzysztofCwalina, line 107: I wish this section started with a strong statement describing why compatibility is important to gaining trust with enterprise customers and how we will be striving for 100% compact. Right now, it reads like a prescription for trivial breaking changes, like removing a "foo" field. If you want to use an example of a breaking change, I would suggest examples related to closing a security hole, privacy issue, or geo-political issue, which by the way are the only API breaking changes we should consider. These changes improve trust we build with customers, and so they offset the loss of trust caused by breaks.

johngossman: Agreed, We have such guidance currently in another doc, and want to merge that either here or up one level in the Microsoft guidelines in a future PR.

azure/Guidelines.md
darrelmiller: The Graph SDKs are able to avoid this problem because unknown properties retrieved from the server are round tripped.  This follows the guidelines defined by the [Atom publishing protocol](https://bitworking.org/projects/atom/rfc5023#rfc.section.9.3).  It is worth mentioning that Microsoft Graph does not make extensive use of PUT.

azure/Guidelines.md
```diff
@@ -0,0 +1,241 @@
...
+### Additional guidance for Azure Resource Manager resource providers
+
+Teams building ARM Resource Providers (RPs) MUST follow the additional guidance in the ARM Resource Provider Contract (RPC) and related documents. These documents can be found here.
+
+* [Azure Resource Manager Wiki][2] (Internal only)
+* [Azure Resource Provider Contract][3]
+
+ARM RPs are a Common Engineering Criteria (CEC) requirement for Azure Services and ARM RP review is another mandatory review. Some of the guidance overlaps with general API review, but passing one review will generally make the other one go very quickly.
+
+## Use OpenAPI Specification to describe API
```
darrelmiller:
```suggestion
## Use an OpenAPI definition to describe API
```
The OpenAPI Specification is [this](http://spec.openapis.org/oas/v3.0.3) document.  The document used to describe an API is an `OpenAPI Definition`.

azure/Guidelines.md
```diff
@@ -0,0 +1,241 @@
...
+
+All Services **MUST** provide an [OpenAPI Specification] that describes their service. The OpenAPI Specification is a key element of the Azure SDK plan and essential to improving the usability and discoverability of services.
```
darrelmiller:
```suggestion
All Services **MUST** provide an [OpenAPI Definition] that describes their service. The OpenAPI Specification is a key element of the Azure SDK plan and essential to improving the usability and discoverability of services.
```

azure/Guidelines.md
```diff
@@ -0,0 +1,241 @@
...
+
+#### Asynchronous operations
+
+The Microsoft REST API guidelines for Long Running Operations are an updated, clarified and simplified version of the Asynchronous Operations guidelines from the 2.1 version of the Azure API guidelines. Unfortunately, to generalize to the whole of Microsoft and not just Azure, the HEADER used in the operation was renamed from `Azure-AsyncOperation` to `Operation-Location`. Services **SHOULD** support both `Azure-AsyncOperation` and `Operation-Location` HEADERS, even though they are redundant so that existing SDKs and clients will continue to operate. Clients that call these services **SHOULD** look for both HEADERS and prefer the `Operation-Location` version. Both HEADERS **MUST** return the same value.
```
darrelmiller, line 14: Has this been finalized?  The Graph guidance has moved away from recommending the use of this custom header.

azure/Guidelines.md
```diff
@@ -0,0 +1,241 @@
...
+
+   ```
+   GET https://blobstore.azure.net/contoso.com/account1/container1/blob2
+   ```
+
+2. That request is returned with the `Content-Location` header set to the direct endpoint.  See [RFC2557]:
+
+   ```
+   200 OK
+   Content-Location: https://contoso-dot-com-account1.blobstore.azure.net/container1/blob2
```
darrelmiller, line 73: Is the Content-Location header identifying the content in response payload, or is the user expected to de-reference the URL in the Content-Location to access the blob?
If it is the latter, then that is not consistent with how [Content-Location](https://tools.ietf.org/html/rfc7231#section-3.1.4.2) is defined.

adrianhall: This is a copy of the existing API guidelines, so it's been in there (like this) for a while.  @johanste or @johngossman - care to comment here?  I'm happy to update to correct an error here.

azure/Guidelines.md
adrianhall: In reading this the first time, I agreed - we need a more meta "what will be considered for a breaking change" and then anything else is automatically not a breaking change.  I think that wording can be a follow-on after an API review board meeting.

azure/Guidelines.md
johngossman: @darrelmiller I don't understand the question

azure/Guidelines.md
johanste: https://github.com/Azure/azure-resource-manager-rpc

azure/Guidelines.md
johanste: This is a consequence of https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#61-ignore-rule

azure/Guidelines.md
johanste: @cleemullins, are you asking about libraries to help the service implement the behavior, or for guidance/libraries that help clients dynamically discover and light up features based on version availability?

azure/Guidelines.md
```diff
@@ -0,0 +1,241 @@
...
+
+### Additional guidance for Azure Resource Manager resource providers
+
+Teams building ARM Resource Providers (RPs) MUST follow the additional guidance in the ARM Resource Provider Contract (RPC) and related documents. These documents can be found here.
+
+* [Azure Resource Manager Wiki][2] (Internal only)
+* [Azure Resource Provider Contract][3]
+
+ARM RPs are a Common Engineering Criteria (CEC) requirement for Azure Services and ARM RP review is another mandatory review. Some of the guidance overlaps with general API review, but passing one review will generally make the other one go very quickly.
```
johanste: Aren't CEC requirements replaces with something else?

azure/Guidelines.md
```diff
@@ -0,0 +1,241 @@
...
+
+## Use an OpenAPI definition to describe API
+
+All Services **MUST** provide an [OpenAPI Definition] that describes their service. The OpenAPI Specification is a key element of the Azure SDK plan and essential to improving the usability and discoverability of services.
```
johanste: documentation is probably a higher order bit than discoverability.

azure/Guidelines.md
```diff
@@ -0,0 +1,241 @@
...
+## URL structure
+
+In addition to the URL structure guidance in the Microsoft REST API guidelines, Azure has specific guidance about service exposure for multi-tenant services
+
+### URL structure
+
+All services **MUST** expose their service to developers via the following URL pattern:
+
+```
+https://<service>.azure.net/<unit-of-multi-tenancy>/<service-defined-root>
```
johanste: Should we call out the fact that different cloud instances (e.g national clouds) have different domains as well (i.e. .azure.net is subject to change between clouds)

azure/Guidelines.md
johanste: @darrelmiller, this is not an alternate redirect implementation. The response body should (in the case of GET) include the data as per the specification. The `Content-Location` can be used for subsequent requests to avoid bouncing off the common front door etc. At least that's how I've read the specification. Please note that I don't think I've ever actually gotten a question on this in an API review.

azure/Guidelines.md
johngossman: Let's call it Azure Fundamentals, which is both accurate and sounds generic. Win win.

azure/Guidelines.md
johanste: I agree that we should change the wording of this. It seems to be a copy/paste mistake from the general guidelines at some point.

There *are* situations where things get thorny - for example when an azure service is acting as a proxy for some other service in part of their APIs.

azure/Guidelines.md
johanste: Yes. I believe this document was originally written specifically for resource providers/RPC. Which both means that it was more specific than general guidance (both based on specific issues that we wanted to prevent before a team even came to the board as well as things that are extra problematic for resource providers). In other words, it was a collection of case law rather than a holistic document.

azure/Guidelines.md
johanste: And please note that we've had instances where a service would reject a call with a read-only property that a client didn't know, but still tried to round-trip. The requirement for client behavior has to be in lock step with how a service is to treat unknown or read-only properties.

azure/Guidelines.md
johanste: Anecdotally, we've shipped client libraries before the service was publicly available (it had not left private preview). It has not ended well. We are definitely shipping libraries before the api version is available in all clouds, however (stack being the prime example of where it will never be possible to be 100% in lock step with public azure)

azure/Guidelines.md
```diff
@@ -0,0 +1,241 @@
...
+
+The following changes are considered backwards compatible and hence non-breaking.
+
+#### Adding new APIs to an existing service
+
+When a new resource types is added, it does not require API version to be updated for existing types.
+
+#### Bug fixes to existing API
+
+Bug fixes to existing API which don’t fall into one of the above categories of breaking changes as described above are fine.
```
johanste, line 164: This implies that the list of breaking changes above is exhaustive. It is not.

azure/Guidelines.md
```diff
@@ -0,0 +1,241 @@
...
+
+This could result in failures which would have earlier succeeded. Even if the rules become less strict, clients relying on earlier name constraints to perform local validation will fail.
+
+### Non-Breaking Changes
+
+The following changes are considered backwards compatible and hence non-breaking.
+
+#### Adding new APIs to an existing service
+
+When a new resource types is added, it does not require API version to be updated for existing types.
```
johanste, line 160: This is ARM/RPC specific (based on the use of the term "resource type")

azure/Guidelines.md
```diff
@@ -0,0 +1,241 @@
...
+Example response:
+
+```
+200 OK
+api-supported-versions: 2011-08,2012-02,1.1,2.0
+api-deprecated-versions: 2009-04,1.0
+Content-Length: 0
+```
+
+Clients that use version discovery are expected to cache version information. Since there’s a year of lead time after an API version shows in the `api-deprecated-versions` before it’s removed, checking once a week should provide sufficient lead time to client authors or operators. In the rare case where a server rolls back a version that clients are already using, the service will reject requests because they are ahead of the latest version supported. Whenever a client sees a `version-too-new` error, it should re-execute its version discovery procedure.
```
johanste, line 202: Is it still a minimum of one year?

azure/Guidelines.md
```diff
@@ -0,0 +1,241 @@
...
+
+## API deprecation policy
+
+Disabling a runtime REST API that customers are dependent on of course has the potential of breaking their applications or services, perhaps even mission critical services.  But inevitably our APIs will become obsolete and the cost of supporting them and operating the servers on which they run will require us to deprecate and shut them down. We have a public policy that describes how we will inform customers that deprecation is coming and help them move their applications off these services and on to their replacements.
+
+### Policy
+
+Azure does not have a single SLA for how long we will support all services. However, we have published expectations such as [the Azure Cloud Lifecycle FAQ][6].  The most relevant section of the document:
+
+> Azure Cloud Services will support no fewer than the latest two SDK versions for deploying new Cloud Services. Microsoft will provide notification 12 months before retiring a SDK in order to smooth the transition to a supported version.
```
johanste: I don't think this is very relevant anymore (it refers to the RDFE cloud services and runtimes)

azure/Guidelines.md
```diff
@@ -0,0 +1,241 @@
...
+
+## API deprecation policy
+
+Disabling a runtime REST API that customers are dependent on of course has the potential of breaking their applications or services, perhaps even mission critical services.  But inevitably our APIs will become obsolete and the cost of supporting them and operating the servers on which they run will require us to deprecate and shut them down. We have a public policy that describes how we will inform customers that deprecation is coming and help them move their applications off these services and on to their replacements.
+
+### Policy
+
+Azure does not have a single SLA for how long we will support all services. However, we have published expectations such as [the Azure Cloud Lifecycle FAQ][6].  The most relevant section of the document:
```
johanste: This link refers to the RDFE style cloud services runtime and client libraries. While it is *an* example, I don't think we want to make it front and center. We should probably refer to the [modern life cycle policies](https://support.microsoft.com/en-us/help/30881).

johngossman: Yes. Thank you. That's the link we need.

azure/Guidelines.md
adrianhall: Agreed - I'll add something here.

azure/Guidelines.md
adrianhall: I think we want to completely restructure the breaking change section as a first order of business .

azure/Guidelines.md
adrianhall: Will replace.

azure/Guidelines.md
adrianhall: I've updated it to the appropriate paragraph.

azure/Guidelines.md
```diff
@@ -0,0 +1,242 @@
...
+Teams building ARM Resource Providers (RPs) MUST follow the additional guidance in the ARM Resource Provider Contract (RPC) and related documents. These documents can be found here.
+
+* [Azure Resource Manager Wiki][2] (Internal only)
+* [Azure Resource Provider Contract][3]
+
+ARM RPs are Azure Fundamentals requirement for Azure Services and ARM RP review is another mandatory review. Some of the guidance overlaps with general API review, but passing one review will generally make the other one go very quickly.
+
+## Use an OpenAPI definition to describe API
+
+All Services **MUST** provide an [OpenAPI Definition] that describes their service. The OpenAPI Specification is a key element of the Azure SDK plan and essential to improving the documentation, usability and discoverability of services.
```
annelo-msft: Is it worth calling out that [specific extensions](https://github.com/Azure/autorest/blob/master/docs/extensions/readme.md) can be used for code-generation?  Or is that too low-level a detail for this guideline?

azure/Guidelines.md
```diff
@@ -0,0 +1,242 @@
...
+
+## URL structure
+
+In addition to the URL structure guidance in the Microsoft REST API guidelines, Azure has specific guidance about service exposure for multi-tenant services
```
annelo-msft: Would it make sense to link to the specific section in the MS REST API guidelines here (and potentially elsewhere), so that readers don't need to search around the guidelines doc to find this?

azure/Guidelines.md
annelo-msft: I'd love to get clarification on this time frame - it comes up often.

azure/Guidelines.md
```diff
@@ -0,0 +1,242 @@
...
+## History
+
+| Date | Version | Notes |
+| 2020-Mar-31 | v3.1 | 1st public release of the Azure REST API Guidelines|
+
+## Introduction
+
+The Azure REST API guidelines are an extension of the [Microsoft REST API guidelines][1]. Readers of this document are assumed to be also reading the [Microsoft REST API guidelines][1] and be familiar with them.  Azure guidance is a superset of the Microsoft API guidelines and services should follow them *except* where this document outlines specific differences or exceptions to those guidelines. This document does contain additional Azure-specific guidance and additional details.
+
+#### Asynchronous operations
```
annelo-msft: This feels like a section that should move outside the Introduction.

annelo-msft: Also, small nit: would it make sense to name this section the same way it appears in the MS REST API Guidelines, i.e. Long running operations?  It would be nice for readers to be able to map the ideas in these guidelines back to the MS guidelines with ease.

azure/Guidelines.md
```diff
@@ -0,0 +1,242 @@
...
+### Additional guidance for Azure Resource Manager resource providers
+
+Teams building ARM Resource Providers (RPs) MUST follow the additional guidance in the ARM Resource Provider Contract (RPC) and related documents. These documents can be found here.
+
+* [Azure Resource Manager Wiki][2] (Internal only)
+* [Azure Resource Provider Contract][3]
+
+ARM RPs are Azure Fundamentals requirement for Azure Services and ARM RP review is another mandatory review. Some of the guidance overlaps with general API review, but passing one review will generally make the other one go very quickly.
+
+## Use an OpenAPI definition to describe API
```
annelo-msft: Another nit: this section header is named inconsistently with other section headers.  Perhaps rename if to "API Definition" or similar?

image

azure/Guidelines.md
darrelmiller: @johanste Thanks.  I was curious because I had also never seen that pattern come up before.

annelo-msft:
The structure of this document is possibly something to address later, but as a reader of these guidelines, I'd love it if we could more clearly draw parallels back to the MS REST API Guidelines -- possibly with common formatting to make it easy to identify visually in the document.  It would be great if we could say in each section related to the MS Guidelines -- (1) this is the link to the section of the guidelines we're referring to (2) this is a correction, or a clarification of which option Azure uses, etc.  In addition, it might be nice to group areas that aren't directly related to the MS Guidelines (it feels like the OpenAPI requirement is this?).  Addressing this in an early stage might also help frame future discussions as well.

azure/Guidelines.md
adrianhall: Agreed.  I'll promote it to its own section.

azure/Guidelines.md
adrianhall: Agreed.

azure/Guidelines.md
adrianhall: I think thats a good addition.

azure/Guidelines.md
adrianhall: Assuming the Microsoft guidelines don't change the header, this is a good thing.  Given the speed at which guidelines change, I don't think this is a problem.

azure/Guidelines.md
adrianhall: Right now, I don't think the guidance of one year has officially changed.  @johngossman - do you have any insight here?

gareth-a-jones (approved):

PR #192, opened 2020-04-10
Azure breaking changes update

adrianhall:


azure/Guidelines.md
```diff
@@ -94,74 +94,47 @@ PUT https://blobstore.azure.com/foo.com/acct1/c1/b2?api-version=2014-12-07
...
-#### New required property added to request
+* An existing property is removed.
+* A new property is added to an existing response.
+* A new required property is added to an existing request.
+* A property name is changed.
+* A property type is changed.
+* The default value of a property is changed.
+* The allowed values for an enum is changed.
+* An API is removed.
+* An API is removed.
```
assafi: nit: duplication

azure/Guidelines.md
```diff
@@ -94,74 +94,47 @@ PUT https://blobstore.azure.com/foo.com/acct1/c1/b2?api-version=2014-12-07
...
+At a high level, any change to the contract of an API constitutes a breaking change. The following is a non-exhaustive list of breaking changes:
 
-#### New required property added to request
+* An existing property is removed.
+* A new property is added to an existing response.
+* A new required property is added to an existing request.
+* A property name is changed.
+* A property type is changed.
+* The default value of a property is changed.
+* The allowed values for an enum is changed.
```
assafi: If they are only expanded, is it still a breaking change? Perhaps refine by saying "renamed and/or removed".

azure/Guidelines.md
```diff
@@ -94,74 +94,47 @@ PUT https://blobstore.azure.com/foo.com/acct1/c1/b2?api-version=2014-12-07
...
+#### Examples of breaking changes
 
-If a new property/field is added to the response of an API, the GET-PUT pipeline will be broken. Consider the case where a customer updates the value of a new property "A" from the Azure portal. Another customer does a GET of this resource using the SDK. The SDK will ignore the property since it does not understand it. From the SDK, the customer does a PUT using the model that was returned from the GET. This will overwrite the change made by the first customer from the portal.
+At a high level, any change to the contract of an API constitutes a breaking change. The following is a non-exhaustive list of breaking changes:
 
-#### New required property added to request
+* An existing property is removed.
+* A new property is added to an existing response.
+* A new required property is added to an existing request.
+* A property name is changed.
```
assafi: May want to also add here that case changes are also included

azure/Guidelines.md
```diff
@@ -94,74 +94,47 @@ PUT https://blobstore.azure.com/foo.com/acct1/c1/b2?api-version=2014-12-07
 POST https://blobstore.azure.com/foo.com/acct1/c1/b2?api-version=2015-12-07
 ```
 
-### Breaking changes in Azure
+### Breaking changes
```
assafi: Thanks Adrian for compiling this, I think it’s really great and would help a lot of us especially those that iterate quite a lot on our APIs.

One tactical comment I would make is that we should consider tie the breaking changing definition with the tools that are defined in the guidelines to deal with them. The guidelines offer a SemVer versioning which includes Major.Minor. This implies that there are Major breaking changes, as well as Minor ones. Should we also say something of the form “these changes are major and these are minor”? 

I think most developers are more apprehensive about major versions upgrades than minor ones, but we have some services that are on their 8th and 12th major API versions already.

azure/Guidelines.md
adrianhall: If previously, you had a finite (small) set of possibilities, this may be encoded as an enum in an SDK.  Adding a new possibility would break the decoding of the response in this case.   So, yes - it is a breaking change.  It may be one we are willing to live with, though - that's a good discussion to have.

azure/Guidelines.md
adrianhall: The major thrust here is "breaking changes (except security, regulatory and API deprecation reasons) are not allowed, but if you do them, then it's a major version bump.  Everything else is a minor version bump.

johngossman: This is the gist of the thread I started. I don't know how services can evolve if adding properties have to meet that security/reg/deprecation bar

azure/Guidelines.md
assafi: In cognitive services we often improve our capabilities without changing the API. For example, a new language detection model can detect a new language. While we have optional model-version query parameters to make sure the behavior can be controlled by the customers, upgrading an API version would really hinder the ability to evolve. 

Another example, would be the set of error codes returned. It's virtually impossible to try and predict all the possible things that can go wrong with a service on the day you release the API.

azure/Guidelines.md
adrianhall: If it is a required addition, then it is absolutely a breaking change.  An optional field with a default value would be ok.  

I think this is something that needs to be discussed during the live meeting.

adrianhall: ... I've just added "a new optional property is added to an existing request" as an example of a non-breaking change.  I'm not sure that's the case (round-tripping concerns), but it's a starting point on this.

azure/Guidelines.md
```diff
@@ -88,80 +86,53 @@ All Azure APIs **MUST** use explicit versioning. The Microsoft REST API guidelin
...
-
-#### New required property added to request
+### API Changes that require a version change
 
-If a new property is made required in the request body, clients will have no way to set this and the request will fail.
+Azure services **MUST** update the version number of their API whenever there is a change to the API, no matter how small.  Customers will "lock the API version" so that their code does not fail when the service introduces new features.  They rely on the fact that an API version is a contract with the services that will never change.
 
-#### Property name has changed
+A breaking change is any change in the API that may cause client or service code making the API call to fail. Obvious examples of such a change are the removal of an endpoint, adding or removing a required field or changing the format of the body (from XML to JSON for example). Even though we recommend clients ignore new fields, there are many libraries and clients that fail when new fields are introduced.
```
anthony-c-martin: When reviewing ARM APIs, we really have two concepts of a 'breaking change', and differentiating between the two can be quite confusing:
1. Client breaks in an obvious way. This guide covers that case well.
2. A change is introduced in a way that neither old or new client breaks, but it becomes possible for a PUT with a newer version of the client to push some additional state to the server, which will be silently reverted if an older version of the client does a GET + PUT.

Since we have expectations of idempotency in ARM, even the introduction of optional request properties is technically a potentially 'breaking change' as per (2) - although it is tends to be subjective for practical reasons. I know that this guide isn't ARM-specific - just wonder if it's worth explicitly calling out that this scenario isn't considered a breaking change as per these guidelines, but some services may consider as such?

azure/Guidelines.md
```diff
@@ -88,80 +86,53 @@ All Azure APIs **MUST** use explicit versioning. The Microsoft REST API guidelin
...
-#### Existing property is removed
-
-If a property called `foo` that was present in v1 of the API needs to be removed, it must be done in a newer API version.
-
-#### New property added to response
-
-If a new property/field is added to the response of an API, the GET-PUT pipeline will be broken. Consider the case where a customer updates the value of a new property "A" from the Azure portal. Another customer does a GET of this resource using the SDK. The SDK will ignore the property since it does not understand it. From the SDK, the customer does a PUT using the model that was returned from the GET. This will overwrite the change made by the first customer from the portal.
-
-#### New required property added to request
+### API Changes that require a version change
```
annelo-msft, line 97: With this change in terminology, is it worth looking at other places we use the term "breaking change" and evaluating whether this is the right thing in those places?  It'll be good to have consistency across this document in how we call these.

azure/Guidelines.md
```diff
@@ -88,80 +86,53 @@ All Azure APIs **MUST** use explicit versioning. The Microsoft REST API guidelin
...
-#### Property name has changed
+A breaking change is any change in the API that may cause client or service code making the API call to fail. Obvious examples of such a change are the removal of an endpoint, adding or removing a required field or changing the format of the body (from XML to JSON for example). Even though we recommend clients ignore new fields, there are many libraries and clients that fail when new fields are introduced.
 
-Note that this is implied by the requirement that adding/and removing properties are breaking changes, but in some ways worse, since it leads to the possibility of reusing a property name. Even with an API version change, this change is discouraged because it creates documentation and cognitive challenges.
+There are three reasons why a service may issue a breaking change:
 
-#### Property type has changed
+1. To remove a security vulnerability.
+2. To comply with regulatory requirements.
+3. To deprecate a feature of the service.
```
annelo-msft: I'm still a little unclear on this guidance, and I'm wondering if restructuring it a bit would make it easier to learn from this document.  

The big thing I'd like to clarify is how we use the term "breaking change" or whether we use it at all -- I know some are in favor of calling these "service evolutions" instead, because they won't technically break customers if you correctly follow the deprecation policy.  We seem to have the following meanings of breaking co-existing here:
- all changes are "breaking" (and therefore require an updated version),
- changes that impact compatibility with prior APIs are "breaking", but some additions don't impact compatibility so are "non-breaking"
- the changes that impact backward compatibility can be done as long as they are not actually "breaking" (i.e. are service evolutions that follow the deprecation policy).
- changes that will actually "break" customers (i.e. affect backward compatibility with prior API versions) can be done only for security and regulatory reasons, subject to REST API Board approval.  (Question: these can affect existing API versions, so would be done without a version bump?)

If this is the case, perhaps restructuring the guidance something like the following would make it a little more clear?

1. All changes to the API require a version change (because all changes are breaking, if this is how we choose to define breaking)
2. Some changes are allowed between versions, namely, those that don't impact backward compatibility
    a. Large additions to the API require major version updates
    b. Small additions to the APIs require minor version updates
3. Changes that would impact backward compatibility with any prior versions are allowed only under the three conditions, and subject to API Board approval
    a. for conditions 1 & 2 (security & regulatory requirements), services may actually break compatibility within an API version, but this is an incredibly high bar.
    b. for condition 3 (deprecation), you must follow the API deprecation policy below so that no customer is broken.

azure/Guidelines.md
```diff
@@ -88,80 +86,53 @@ All Azure APIs **MUST** use explicit versioning. The Microsoft REST API guidelin
...
+Azure services **MUST** update the version number of their API whenever there is a change to the API, no matter how small.  Customers will "lock the API version" so that their code does not fail when the service introduces new features.  They rely on the fact that an API version is a contract with the services that will never change.
 
-#### Property name has changed
+A breaking change is any change in the API that may cause client or service code making the API call to fail. Obvious examples of such a change are the removal of an endpoint, adding or removing a required field or changing the format of the body (from XML to JSON for example). Even though we recommend clients ignore new fields, there are many libraries and clients that fail when new fields are introduced.
 
-Note that this is implied by the requirement that adding/and removing properties are breaking changes, but in some ways worse, since it leads to the possibility of reusing a property name. Even with an API version change, this change is discouraged because it creates documentation and cognitive challenges.
+There are three reasons why a service may issue a breaking change:
```
markcowl, line 101: Are these meant to be reasons why a change could be made without changing the api-version, or are api-version changes required, even in these cases?

azure/Guidelines.md
markcowl: Is this a reason not to model a property as an enum?  I think it would be useful to have a distinction between 'known good values' and 'an exhaustive enumeration of possibilities' - where the latter is an enum.

azure/Guidelines.md
```diff
@@ -88,80 +86,53 @@ All Azure APIs **MUST** use explicit versioning. The Microsoft REST API guidelin
...
-#### Allowed values for an enum have changed
-Enum “foo” had allowed values as “val1” and “val2” in v1 of API. Now, the values accepted by the service are “val1”, “val2” and “val3”.  The client will fail to de-serialize if “val3” comes back in the response.
+* An existing property is removed.
+* A new property is added to an existing response.
+* A new required property is added to an existing request.
+* A property name is changed (including case changes).
+* A property type is changed.
+* The default value of a property is changed.
+* The allowed values for an enum is changed.
+* An API is removed.
```
markcowl: API = operation? path? (assume either)

azure/Guidelines.md
```diff
@@ -88,80 +86,53 @@ All Azure APIs **MUST** use explicit versioning. The Microsoft REST API guidelin
...
+* The behavior of an existing API is changed.
+* The error contract has changed.
```
southpolesteve: Does this include the string contents of an error message? Does changing an error message count as a breaking change? For Cosmos, these are frequently iterated on to provide better debugging information to the user. I'm fairly sure we don't consider these changes breaking.

southpolesteve, line 122: Does this include the string contents of an error message? Does changing an error message count as a breaking change? For Cosmos, these are frequently iterated on to provide better debugging information to the user. I'm fairly sure we don't consider these changes breaking.

azure/Guidelines.md
```diff
@@ -88,80 +86,53 @@ All Azure APIs **MUST** use explicit versioning. The Microsoft REST API guidelin
...
-Enum “foo” had allowed values as “val1” and “val2” in v1 of API. Now, the values accepted by the service are “val1”, “val2” and “val3”.  The client will fail to de-serialize if “val3” comes back in the response.
+* An existing property is removed.
+* A new property is added to an existing response.
+* A new required property is added to an existing request.
+* A property name is changed (including case changes).
+* A property type is changed.
+* The default value of a property is changed.
+* The allowed values for an enum is changed.
+* An API is removed.
+* The behavior of an existing API is changed.
```
southpolesteve: I would like to see more clarity, some examples here. One immediate example comes to mind. Cosmos query using ORDER BY does not currently return results where the ORDER BY property is undefined. Customers often find this surprising.

There are some people that want to change this behavior. Is this a breaking change? How would we introduce it? This is only detectable as runtime behavior. Same API contract.

azure/Guidelines.md
```diff
@@ -88,80 +86,53 @@ All Azure APIs **MUST** use explicit versioning. The Microsoft REST API guidelin
...
 GET https://blobstore.azure.com/foo.com/acct1/c1/blob2?api-version=1.0
 PUT https://blobstore.azure.com/foo.com/acct1/c1/b2?api-version=2014-12-07
 POST https://blobstore.azure.com/foo.com/acct1/c1/b2?api-version=2015-12-07
 ```
 
-### Breaking changes in Azure
-
-A breaking change is any change in the API that may cause client or service code making the API call to fail. Obvious examples of such a change are the removal of an endpoint, adding or removing a required field or changing the format of the body (from XML to JSON for example).
-
-Even though we recommend clients ignore new fields, there are many libraries and clients that fail when new fields are introduced. Azure services **MUST** update the version number of their API even when adding optional fields. In fact, servers should be as strict as possible. Ignoring a field can result in the API accepting content that containered a typo or an element at the wrong level of nesting. If this missing field changes the semantics (for example, we have seen cases where security settings were misplaced and ignored, leaving the resources more exposed than intended) this can be a huge and hard to discover error.
```
gareth-a-jones, line 101: I want to challenge this piece of accepted wisdom.   At the very least, if it's going to state libraries, it should cite some popular ones that behave like this.   This one statement drives so much decision making and given the experience we have in other groups at Microsoft, I simply don't buy it.  It may have been true at the dawn of Azure, but I don't believe it is true now.   And given the customer pain in having to track and chase version changes caused by this when the large majority of change is additive and wouldn't otherwise require a version change at all, there is a business decision to be made between pain for a tiny minority of customers vs pain for a large majority thereof.

azure/Guidelines.md
gareth-a-jones: Here's part of the problem - we need to get Azure off the ultra-dated concept of PUT as a typical change mechanism.  The industry has moved on and PATCH is the accepted standard for updates now.  This problem of downlevel reversion simply goes away.   How to we get Azure APIs to evolve away from the typical use of PUT for updates?

azure/Guidelines.md
gareth-a-jones: In Graph we have have the explicit notion of an extensible enum, where customers have to opt in with a header if they are capable of dealing with a changing set of values.  If they don't opt in, then they get a marker value called 'unkownfuturevalue' returned instead of the new value.  So simpler customers get an easy experience with a fixed set plus a marker.  More sophisticated customers get access to the full set of values.   ALl customers do have to deal with unkownfuturevalue, but that is part oft he enum set from day one.

azure/Guidelines.md
gareth-a-jones: We've often used the razor of 'Change from not working to working is an OK change", cos only an insignificant amount of customers rely on something not working.

azure/Guidelines.md
gareth-a-jones: Per the top-level API guidelines on errors, the message text is just human readable, not machine readable, therefore you can make any change you like.   However, the set of codes you return in top-levle messages is a fixed set for which adding new ones is a breaking change.  Adding an innerError to an existing error case is not a breaking change and allows for adding more granular information in a machine-readable form.

azure/Guidelines.md
```diff
@@ -88,80 +86,64 @@ All Azure APIs **MUST** use explicit versioning. The Microsoft REST API guidelin
...
+| Integer widens         | Evolutionary | Breaking     |
+| Integer narrows        | Breaking     | Evolutionary |
+| Add new value to enum  | Evolutionary | Breaking     |
+| Remove value from enum | Breaking     | Breaking     |
+| Optional to required   | Breaking     | Breaking     |
+| Required to optional   | Evolutionary | Breaking     |
+
+Breaking changes require prior approval of the Azure REST API review board. In the case of deprecation, follow the API deprecation policy (below).  If the service is using SemVer for versioning, breaking changes constitute a major version change.
+
+Evolutionary changes do not require prior approval (but still need a version bump).  If the service is using SemVer for versioning, evolutionary changes constitute a minor version change.
```
southpolesteve, line 125: Is it possible to have a patch level version change or is `Major.Minor` the only format covered under the guidelines?

southpolesteve: Looking through the larger document, I don't see the `Major.Minor` explicitly defined, but that is expected.

azure/Guidelines.md
southpolesteve: Trying to find some middle ground... what if we allowed clients to opt-in to picking up evolutionary changes without a version bump?

`GET https://blobstore.azure.com/foo.com/acct1/c1/blob2?api-version=1.x`

Resolves to the latest 1.x version of the API. It doesn't solve service side concerns but does allow customers to opt into less conservative behavior.

gareth-a-jones: That's pretty interesting - we are very used to doing ~ or ^ as version prefixes to semver in our NPM packages in order to say, hey I want to pick newer stuff up until there's a major change.

gareth-a-jones: And I am ALL for customers getting to opt-in to te level of sophistication they happen to think about as a client.

azure/Guidelines.md
adrianhall: To resolve this, I've added an explicit mention of the extensible enum.

azure/Guidelines.md
adrianhall: The current document states that only date-versioning is allowed.  However, some service groups have decided at a business level to adopt major.minor (e.g. cognitive services).  This is basically saying "don't use semver, but if you do - here are the rules"

johngossman: The current document recommends M.m versions and "group versions" which are date based

https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#121-versioning-formats

azure/Guidelines.md
```diff
@@ -84,84 +82,91 @@ https://<tenant-id>-<service-defined-root>.<service>.azure.net
...
 
-#### Behavior of existing API has changed
+There are a few mechanisms that can reduce breaking changes and their effects on our customers.
 
-There is a functional change in what the API was doing. This is a complex issue because it sometimes is not an easy option to maintain the old behavior even on an old API version. It also is very confusing to end users even when version is update and documented. Behavior changes need to be well justified and discussed on a case-by-case basis.
+#### Use PATCH instead of PUT
```
```suggestion
#### Use PATCH instead of PUT for updates.
```

azure/Guidelines.md
```diff
@@ -84,84 +82,103 @@ https://<tenant-id>-<service-defined-root>.<service>.azure.net
...
-Property `foo was a boolean in v1 but is changed to a string. A client using the existing API version tries to set it as a boolean, but the service will fail since its now expecting a string. So, the API version must be updated.
+#### Changing the API without changing the version
 
-#### Property default value has changed
+There are a limited set of situations where changing the API is permissable without a version bump:
 
-If a property is optional and the service provides a default value, changing that default requires an updated API version.
+1. Adding a new (optional) value to an extensible enum.
+2. Adding a new (optional) query parameter to adjust the output (for example, adding filter options to a list operation).
+3. Adding optional computed read-only output values that are generated based on the new (optional) query parameters.
```
ravbhatnagar: @adrianhall - can you please provide some more context on why an optional writable property is considered breaking but a read only computed property is not a breaking change? The strict clients will fail to deserialize? am i missing something?

azure/Guidelines.md
```diff
@@ -84,84 +82,103 @@ https://<tenant-id>-<service-defined-root>.<service>.azure.net
...
 
-If a new property is made required in the request body, clients will have no way to set this and the request will fail.
+A _breaking change_ is any change in the API that may cause client or service code making the API call to fail. Obvious examples of such a change are the removal of an endpoint, adding or removing a required field or changing the format of the body (from XML to JSON for example). Even though we recommend clients ignore new fields, there are many libraries and clients that fail when new fields are introduced. Removing an endpoint from an API is always a _breaking change_.  Adding a new endpoint is always an _evolutionary change_.  Changes to properties may be _evolutionary_ or _breaking_ depending on the type of change and whether the change is to an input parameter or output parameter:
```
ravbhatnagar, line 107: not sure if adding new apis on existing types fall in the bucket of a change needing new api-version or not?

ravbhatnagar: Do we have data on how many calls (what percentage, and which customers) we see from these strict libraries and clients which fail to deserialize upon seeing new fields? Just trying to understand if we are optimizing for 90% case or 10%. May be excluding the clients/libraries we own (as if needed we can perhaps fix/change stuff on things we own)

 I would like to reiterate, based on my investigation, GCP and AWS both allow addition of optional properties in the same api-version. GCP has this publicly documented, and for AWS because they dont publicly document APIs/guidance, the knowledge comes via people who have worked in AWS. So it may be in inaccurate and happy to be corrected on that. More so, even our other large API platform (Graph) allows for properties to be added on the same api-version. 

I completely understand that this (requiring api-version bump on even property adds) is perhaps the safest thing to do. And will ensure that no one will ever be broken. But, is this too idealist of a view? 

Because if we look on the other side, it does cause a lot of challenges - 
1) either customer gets an explosion of api-versions (since every small change will be a new api-version) and had to spend considerable time and effort before adopting a new version 
2) slow down new capabilities being added on the service side (by requiring a fixed cadence) 
3) The problem of major change vs. minor change still does not get solved by this guidance as Azure API versioning is date based on the service side. So customer has no way of knowing what is major vs what is minor. And most likely treats everything as major, which slows customer adoption of newer stuff. 
4. it will be a nightmare enforcing this change/guidance (but still a problem we can try to address). Many times, swagger is the last thing that service teams update. And the change has already been deployed and made its way to portal and likely customers are already using it. 
Apologies for pushing back on this over and over again, but lets make a decision based on data and after understanding the flip side/challenges it will cause. @johngossman @garethj-msft @adrianhall

azure/Guidelines.md
```diff
@@ -84,84 +82,103 @@ https://<tenant-id>-<service-defined-root>.<service>.azure.net
...
 
-Resource parameter names change from `/resourceType1/{resourceType1_name}` to `/resourceType1/{resourceType1_id}`. This will impact code generation.
+* User1 creates a resource with version v2, using a new optional parameter.
+* Later, User2 wants to update the resource using unrelated settings.  Using version v1, User2 issues a GET, does the changes, and then issues a PUT to replace the resource definition.
 
-#### Resource naming rules should not change
+In this case, the optional parameter is lost because of the replace semantics.  The optional parameter only exists on API version v2, and not on version v1.
 
-This could result in failures which would have earlier succeeded. Even if the rules become less strict, clients relying on earlier name constraints to perform local validation will fail.
+Service teams SHOULD prefer and recommend PATCH operations for updating resources.
```
ravbhatnagar, line 169: This is great guidance and something we have been strongly recommending to teams (specially new ones). But getting to 80% coverage here is a multi year effort. So, what should be the guidance in the meantime? Overwrite the goal state? Guidance from ARM team has been to 400 bad request the PUT call from older version if the resource on the server has properties which were set by a newer version and only available to that version. This will prevent unintentional overwrite. And moreover, today ARM templates cant invoke PATCH. So, customers problems are not holistically solved with this guidance. We need to do something more alongside this guidance.

azure/Guidelines.md
```diff
@@ -84,84 +82,103 @@ https://<tenant-id>-<service-defined-root>.<service>.azure.net
...
 
-There is a functional change in what the API was doing. This is a complex issue because it sometimes is not an easy option to maintain the old behavior even on an old API version. It also is very confusing to end users even when version is update and documented. Behavior changes need to be well justified and discussed on a case-by-case basis.
+Even changes that are evolutionary can cause problems.  For instance, let's say that a service adds a new feature via a new endpoint in the API.  The SDK gets updates to support this new API, but the service does not bump the version number.  Since the roll out of the new feature is not atomic, there is a period of time (potentially months long) where the feature is available in some regions but not others.  A customer has the potential for attempting to use the feature in two different regions and having it work in one region but not the other, despite the two regions supporting the same version number.  This is only made worse when we consider Azure Stack, which can be upwards of a year behind the public cloud offerings.
```
ravbhatnagar, line 154: For the region scenario you described below, the service is expected to be rolled out to all regions it supports following SDP. If its not available in all regions, one case might be that the feature is being flighted (using Azure Feature Exposure COntrol). And in that case its only available to certain regions. But for this case, only selected customers get to use the feature and aware of regions where that capability is available. And even in the generic case if it does happen, then it wouldnt be a breaking change to the application. In the error message, The customer will be informed about the regions that the capability is available in and guided to target one of the allowed regions. Its standard discoverability in Azure exposed through metadata APIs which customers can use to know which resource types support which api-versions/locations. So, not sure if this example is completely accurate.

azure/Guidelines.md
ravbhatnagar: There is still a chance that some service team may cause a livesite by not properly handling new properties in the same api-version. Is there something we can do catch such instances- test runners? canary coverage? 
I know few teams have handled it poorly which resulted in livesite, but should we go the other extreme and have strict guidance which have other disadvantages, or explore other ways to mitigate such cases through proper test coverage and reporting.

azure/Guidelines.md
ravbhatnagar: IN addition, as mentioned on the call, having 2 and 3 points above will result in sub-optimal api design. These are service generated properties, and now customer will have to understand which query param to use to get a particular value. And it will also result in an explosion of query params if for every new read only property, a new query param gets introduced.

JeffreyRichter: @adrianhall: In addition to the wording you're going to add about not allowing this for multi-cloud services, I also think wording should be added that the service must be updated in ALL datacenters before the new query parameter is advertised to customers. This removes the race condition of a customer passing the new parameter against a version of the service that doesn't support the query parameter.

azure/Guidelines.md
anthony-c-martin: @ravbhatnagar - could you clarify on what you mean by 'not properly handling new properties in the same api-version'?

To me it feels like there are two options for the RP implementation, with the following drawbacks:
1. **PUT-as-PATCH**: Template deployments become non-deterministic/repeatable.
2. **Strict PUT semantics**: An older version of the SDK can inadvertently perform a destructive update.

johanste: There is a middle ground that is PUT as "replace everything the client should know about in the given API version and keep the rest as-is". Which is a variation of PUT-as-PATCH.

azure/Guidelines.md
adrianhall: Ack - added language around this.

azure/Guidelines.md
adrianhall: If the service is updated with the same API version number, and different implementations of the same version are available in different regions, then this is absolutely a valid scenario - despite flighting.  Flighting only works if there are different API versions for different functionality.

azure/Guidelines.md
adrianhall: Ack on the change - done.

azure/Guidelines.md
adrianhall: ARM has its own review board - we should be moving the whole org into the direction of allowing PATCH, including ARM.  But it's a multi-year effort, as you said.  We'll add guidance as we move along.  I'll add it to the backlog.

azure/Guidelines.md
ravbhatnagar: Anthony - There are two options in my mind (and RP can chose) 
1. RP can explicitly disallow update to a resource which was created using v2 if the request comes from v1. Specially if the resource has properties set which are not exposed to v1. 400 bad request with a nice error message indicating that the resource was created with a new version, please move to latest version to update it. 
2. As Johan was saying, and this is deviating from what true PUT definition, but an option where I see a path to bringing some consistency in behavior across services. Else, in other options it will be next to impossible to get existing RPs change the behavior and fix PUTCH. So, yes, Microsoft Azure invents a new definition of PUT (on similar lines as Johan was saying). But atleast we have a path to achieve consistency across 200+ Azure services. Something to noodle on (there are definitely challenges/concerns I am aware of, but willing to explore further)

azure/Guidelines.md
```diff
@@ -56,26 +54,26 @@ In addition to the required format above, services **MAY** also choose to expose
 
 The format of the root of the direct endpoint **MUST** be as follows:
 
-```
+```text
 https://<tenant-id>-<service-defined-root>.<service>.azure.net
 ```
 
 1. A request is made to the default end point (GET or HEAD).  For example:
```
JeffreyRichter: end point should be 1 word

azure/Guidelines.md
```diff
@@ -84,84 +82,114 @@ https://<tenant-id>-<service-defined-root>.<service>.azure.net
 
 All Azure APIs **MUST** use explicit versioning. The Microsoft REST API guidelines offer different options on how to specify an API version and guidance on what constitutes a breaking change.  This section of the Azure API guidelines describes updates those guidelines to ensure consistency between Azure services across Azure Stack, public Azure, and sovereign clouds.
```
JeffreyRichter: typo: "describes updates TO  those guidelines"

azure/Guidelines.md
```diff
@@ -84,84 +82,114 @@ https://<tenant-id>-<service-defined-root>.<service>.azure.net
...
+   ],
+   "x-ms-enum": {
+      "name": "createdByType",
+      "modelAsString": true
+   }
+}
+```
 
-Bug fixes to existing API which don’t fall into one of the above categories of breaking changes as described above are fine.
+If you can foresee that an enum will be extended in the future, model the enum as an extensible enum.
```
JeffreyRichter: I"d rephrase this: If you are positive that the enum's symbol set will NEVER change over time, then use model as an enum otherwise, model the enum as a string.

azure/Guidelines.md
```diff
@@ -185,14 +213,14 @@ Azure services **SHOULD** support API version discovery.  If they support it:
 
 Example request to discover versions (blob storage container list API):
```
JeffreyRichter: discover versions of what? supported API versions? versions of a blob?

PR #205, opened 2020-05-30
Update Guidelines.md

y-code:
Fix rendering of the group-version correspondence table in the example in `12.1.1 Group versioning`. The table is interpreted as code blocks because of the four-space at the beginning of each line. Removing the leading spaces is one way to correct it, but for readability in text, I corrected it with leading `|`s instead.


gareth-a-jones:
Thank you Yas - could I ask you to point this to next rather than master please?

y-code:
It turned out that the part I fixed has already been fixed in vNext in the same manner. I'm going to close this PR.

PR #207, opened 2020-06-18
Updated header casing bullet in section 17.2

JeffreyRichter:


Guidelines.md
```diff
@@ -2162,7 +2162,7 @@ Services SHOULD use verbose naming patterns and SHOULD NOT use abbreviations oth
 ### 17.2. Casing
 - Acronyms SHOULD follow the casing conventions as though they were regular words (e.g. Url).
 - All identifiers including namespaces, entityTypes, entitySets, properties, actions, functions and enumeration values SHOULD use lowerCamelCase.
-- HTTP headers are the exception and SHOULD use standard HTTP convention of Capitalized-Hyphenated-Terms.
+- Ideally, HTTP headers are sent over hte wire using lowercase letters a the HTTP/2 specification mandates this ([RFC](https://http2.github.io/http2-spec/#HttpHeaders)). For backwards compatibility, services should treat request headers in a case-insensitive fashion. New services should send response headers in all lowercase. For legacy services (which do not send lowercase response headers), the client should process the response headers in a case-insensitive fashion. If a client converts headers into a map/dictionary where the lanaguge does not support looking up keys in a case-insensitive way, the client library should lowercase the header keys when adding them to the map/dictionary.
```
Petermarcu: hte -> the. a -> as.

gareth-a-jones: Hi Jeffrey - thanks for the contribution.  We're rather behind on reviewing PRs at this point, but we appreciate it nonetheless.

As well as Peter's comments above, can I ask that you express this change in normative language please.   For example, I would have expected the backwards compatibility clause to be a MUST rather than a SHOULD, and capitalized regardless.  

Also, it appears the link to the HTTP2 spec has become outdated.    I believe this is now the correct link. https://httpwg.org/specs/rfc7540.html#HttpHeaders

I'm not sure I agree with the proposal for new services to return headers in lower case.   It seems to me for optimum compatibility with marginally compliant older client implementations, this should be protocol dependent.   If serving over Http/1.x services should return using the Capitalized-Hyphenated style whereas Http/2 should use lowercase-hyphenated.   

Client behavioral guidance should be broken out from the main text, as these are not primarily client guidelines.

PR #208, opened 2020-07-08
Azure v1 service advice

adrianhall:
V1 advice for Azure services.

### DIFF
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -32,7 +32,7 @@ Developing a new service requires the development of at least 1 (management plan
   * We recommend a minimum of 2 preview versions prior to your first GA release.  
   * Your API should be in preview for a minimum of 3 months to gain customer insights on usage.
 
-Breaking changes are a source of concern during initial review.  Without a history, the review process attempts to identify patterns that may result in breaking changes later on.  For instance:
+Preventing future breaking changes is a source of concern during initial review.  Without a history, the review process attempts to identify patterns that may result in breaking changes later on.  For instance:
 
 * Return collections with server-side paging, even if your resource does not currently need paging.  This avoids a breaking change when your service expands.
 * Think about how the developer can reason about the collection organization.  Filtering is a common customer request.
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -2,8 +2,10 @@
 
 ## History
 
-| Date | Version | Notes |
-| 2020-Mar-31 | v3.1 | 1st public release of the Azure REST API Guidelines|
+| Date        | Version | Notes                                               |
+| ----------- | ------- | --------------------------------------------------- |
+| 2020-Mar-31 | v3.1    | 1st public release of the Azure REST API Guidelines |
+| 2020-Jul-31 | v3.2    | Added service advice for initial versions           |
 
 ## Introduction
 
@@ -17,6 +19,33 @@ Teams building ARM Resource Providers (RPs) MUST follow the additional guidance 
 * [Azure Resource Provider Contract][3]
 
 ARM RPs are Azure Fundamentals requirement for Azure Services and ARM RP review is another mandatory review. Some of the guidance overlaps with general API review, but passing one review will generally make the other one go very quickly.
+
+### Advice for new services
+
+Developing a new service requires the development of at least 1 (management plane) API and potentially one or more additional (data plane) APIs.  
+
+* Think about naming from the context of a **developer experience**.  
+  * Start with the "things" your API manipulates, then think about the operations that a developer needs to do to these "things".
+  * Avoid the use of generic names like "Object", "Job", "Task", "Operation" (for example - the list is not exhaustive).
+  * What happens to the names when the focus of the service expands?  It may be worth starting with a less generic name to avoid a breaking change later on.
+* Use previews to get the shape right.  There are different notification, breaking change, and lifetime requirements on preview API versions.
+  * We recommend a minimum of 2 preview versions prior to your first GA release.  
+  * Your API should be in preview for a minimum of 3 months to gain customer insights on usage.
+
+Breaking changes are a source of concern during initial review.  Without a history, the review process attempts to identify patterns that may result in breaking changes later on.  For instance:
+
+* Return collections with server-side paging, even if your resource does not currently need paging.  This avoids a breaking change when your service expands.
+* Use extensible enumerations unless you are completely sure that the enumeration will never expand.  Extensible enumerations are modeled as strings - expanding an extensible enumeration is not a breaking change.
+* Implement [conditional requests](https://tools.ietf.org/html/rfc7232) early.  This allows you to support concurrency, which tends to be a concern later on.
+* If your API specified access conditions to another resource, think about how to represent that model polymorphically.  For example, you may be using a SQL Azure connection now, but extend to Cosmos DB, Azure Data Lake, or Redis Cache later on.  Think about how you can specify that resource in a non-breaking manner.
+* Be concerned about data widths of numeric types.  Wider data types (e.g. 64-bit vs. 32-bit) are more future-proof.
+* Think about how the interface will be represented by an SDK.  For example, JavaScript can only support numbers up to 2<sup>53</sup>, so relying on the full width of a 64-bit number should be avoided.
+* Implement (and encourage) the use of PATCH for resource modifications.  The PATCH operation should be able to modify any mutable property on the resource.
+
+Additionally, for management APIs:
+
+* Follow the advice in the [Azure Resource Manager Wiki][2] (internal only).
+* Use [RPaaS](https://armwiki.azurewebsites.net/rpaas/overview.html) (internal only) to implement the Azure Resource Provider.
 
 ## API definition
 
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -22,27 +22,42 @@ ARM RPs are Azure Fundamentals requirement for Azure Services and ARM RP review 
 
 ### Advice for new services
 
-Developing a new service requires the development of at least 1 (management plane) API and potentially one or more additional (data plane) APIs.  
+Developing a new service requires the development of at least 1 (management plane) API and potentially one or more additional (data plane) APIs.  When reviewing v1 service APIs, we see common advice provided during the review.
+
+> A **management plane** API is implemented through the Azure Resource Manager (ARM) and is used by subscription administrators.  A **data plane** API is used by developers to implement applications.  Rarely, a subset of operations may be useful to both administrators and users, in which case it should appear in both APIs.
 
 * Think about naming from the context of a **developer experience**.  
   * Start with the "things" your API manipulates, then think about the operations that a developer needs to do to these "things".
+  * Keep the verbs present-tense.  Avoid the use of past or future tense in most cases.
   * Avoid the use of generic names like "Object", "Job", "Task", "Operation" (for example - the list is not exhaustive).
   * What happens to the names when the focus of the service expands?  It may be worth starting with a less generic name to avoid a breaking change later on.
+* Think about the code that a customer will write both before and after the REST API call.  How will a developer use this API in the canonical use case?
+  * Consider multiple languages, and include at least one dynamically typed language (for example, Python or JavaScript) and one statically typed language (for example, Java or C#).
 * Use previews to get the shape right.  There are different notification, breaking change, and lifetime requirements on preview API versions.
-  * We recommend a minimum of 2 preview versions prior to your first GA release.  
-  * Your API should be in preview for a minimum of 3 months to gain customer insights on usage.
+  * We recommend a minimum of 2 preview versions prior to your first GA release.  However, there is no hard rule for previews.  
+  * You should gather feedback from your customers and iterate until the API is useful.  Actively solicit feedback from your preview customers.
 
-Preventing future breaking changes is a source of concern during initial review.  Without a history, the review process attempts to identify patterns that may result in breaking changes later on.  For instance:
+Preventing future breaking changes is a source of concern during initial review.  Without a history, the review process attempts to identify patterns that may result in breaking changes later on.  For instance, think about the following:
 
-* Return collections with server-side paging, even if your resource does not currently need paging.  This avoids a breaking change when your service expands.
-* Think about how the developer can reason about the collection organization.  Filtering is a common customer request.
-* Prefer allowing the developer to specify the ID of the resource.  A developer will commonly want to download a specific resource by name.
+* [Think about idempotency](https://stripe.com/blog/idempotency).  In a distributed cloud, each HTTP call must be idempotent.  You must be resilient in the face of failure.  A developer must rely on idempotency to build fault-tolerant systems.
+  * The HTTP specification requires that GET, PUT, DELETE, and HEAD be idempotent.
+  * Prefer allowing the developer to use PUT or PATCH to create a resource with a user-specified name or ID.  A developer will commonly want to download a specific resource by name or ID.  This provides the developer with an idempotent mechanism for creating resources.
+* Collections are a common source of review comments:
+  * Return collections with server-side paging, even if your resource does not currently need paging.  This avoids a breaking change when your service expands.
+  * A collection should return a homogeneous collection (single type).  Do not return heterogeneous collections unless there is a really good reason to do so.  If you feel heterogeneous collections are required, discuss the requirement with an API reviewer prior to implementation.
+  * Think about how the developer can reason about the collection organization.  Filtering is a common customer request.
+* Avoid polymorphism.  An endpoint should work with a single type to avoid problems during SDK creation.  Remember that a change to the model is a breaking change.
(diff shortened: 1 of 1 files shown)


JeffreyRichter (approved):

azure/Guidelines.md
```diff
@@ -18,6 +20,35 @@ Teams building ARM Resource Providers (RPs) MUST follow the additional guidance
...
 ARM RPs are Azure Fundamentals requirement for Azure Services and ARM RP review is another mandatory review. Some of the guidance overlaps with general API review, but passing one review will generally make the other one go very quickly.
 
+### Advice for new services
+
+Developing a new service requires the development of at least 1 (management plane) API and potentially one or more additional (data plane) APIs.  
+
+* Think about naming from the context of a **developer experience**.  
+  * Start with the "things" your API manipulates, then think about the operations that a developer needs to do to these "things".
+  * Avoid the use of generic names like "Object", "Job", "Task", "Operation" (for example - the list is not exhaustive).
+  * What happens to the names when the focus of the service expands?  It may be worth starting with a less generic name to avoid a breaking change later on.
```
chrisspre, line 33: Maybe it is my english but isn't it the other way around: more generic name leads to fewer breaking changes?

azure/Guidelines.md
```diff
@@ -18,6 +20,35 @@ Teams building ARM Resource Providers (RPs) MUST follow the additional guidance
...
+* Use previews to get the shape right.  There are different notification, breaking change, and lifetime requirements on preview API versions.
+  * We recommend a minimum of 2 preview versions prior to your first GA release.  
+  * Your API should be in preview for a minimum of 3 months to gain customer insights on usage.
+
+Breaking changes are a source of concern during initial review.  Without a history, the review process attempts to identify patterns that may result in breaking changes later on.  For instance:
+
+* Return collections with server-side paging, even if your resource does not currently need paging.  This avoids a breaking change when your service expands.
+* Think about how the developer can reason about the collection organization.  Filtering is a common customer request.
+* Prefer allowing the developer to specify the ID of the resource.  A developer will commonly want to download a specific resource by name.
```
chrisspre: should this say id/name in both places : "... to specify the id/name of ... " and "... download a specific resource by id/name

azure/Guidelines.md
```diff
@@ -18,6 +20,35 @@ Teams building ARM Resource Providers (RPs) MUST follow the additional guidance
...
+* Use extensible enumerations unless you are completely sure that the enumeration will never expand.  Extensible enumerations are modeled as strings - expanding an extensible enumeration is not a breaking change.
```
chrisspre, line 50: Is it necessary to mentioned that even with a extensible enumeration / string there is a small likelihood that a client might not be able to deal with new values ?

azure/Guidelines.md
```diff
@@ -18,6 +20,35 @@ Teams building ARM Resource Providers (RPs) MUST follow the additional guidance
...
+* Implement [conditional requests](https://tools.ietf.org/html/rfc7232) early.  This allows you to support concurrency, which tends to be a concern later on.
+* If your API specified access conditions to another resource, think about how to represent that model polymorphically.  For example, you may be using a SQL Azure connection now, but extend to Cosmos DB, Azure Data Lake, or Redis Cache later on.  Think about how you can specify that resource in a non-breaking manner.
+* Be concerned about data widths of numeric types.  Wider data types (e.g. 64-bit vs. 32-bit) are more future-proof.
+* Think about how the interface will be represented by an SDK.  For example, JavaScript can only support numbers up to 2<sup>53</sup>, so relying on the full width of a 64-bit number should be avoided.
+* Implement (and encourage) the use of PATCH for resource modifications.  The PATCH operation should be able to modify any mutable property on the resource.
```
TomMilos: With more complicated schemas basic PATCH can be ambiguous (e.g. when dealing with arrays of objects). One approach to address that is using JSON Patch (https://tools.ietf.org/html/rfc6902) - should we consider adding it to the guidelines as a recommended/approved approach?

azure/Guidelines.md
```diff
@@ -18,6 +20,35 @@ Teams building ARM Resource Providers (RPs) MUST follow the additional guidance
...
+
+* Think about naming from the context of a **developer experience**.  
+  * Start with the "things" your API manipulates, then think about the operations that a developer needs to do to these "things".
+  * Avoid the use of generic names like "Object", "Job", "Task", "Operation" (for example - the list is not exhaustive).
+  * What happens to the names when the focus of the service expands?  It may be worth starting with a less generic name to avoid a breaking change later on.
+* Use previews to get the shape right.  There are different notification, breaking change, and lifetime requirements on preview API versions.
+  * We recommend a minimum of 2 preview versions prior to your first GA release.  
+  * Your API should be in preview for a minimum of 3 months to gain customer insights on usage.
+
+Breaking changes are a source of concern during initial review.  Without a history, the review process attempts to identify patterns that may result in breaking changes later on.  For instance:
```
JeffreyRichter: While previewing v1 of a service, breaking changes are not a concern. After v1 has reached general availability, breaking changes should NOT be done. If you feel that a breaking change is required, then a board review is required.

azure/Guidelines.md
```diff
@@ -18,6 +20,35 @@ Teams building ARM Resource Providers (RPs) MUST follow the additional guidance
 
 ARM RPs are Azure Fundamentals requirement for Azure Services and ARM RP review is another mandatory review. Some of the guidance overlaps with general API review, but passing one review will generally make the other one go very quickly.
 
+### Advice for new services
+
+Developing a new service requires the development of at least 1 (management plane) API and potentially one or more additional (data plane) APIs.  
+
+* Think about naming from the context of a **developer experience**.  
+  * Start with the "things" your API manipulates, then think about the operations that a developer needs to do to these "things".
```
assafi, line 30: @johanste should we have some guideline around the use of verbs/nouns in REST APIs, similar to Powershell best practices?

azure/Guidelines.md
```diff
@@ -18,6 +20,35 @@ Teams building ARM Resource Providers (RPs) MUST follow the additional guidance
...
+  * Avoid the use of generic names like "Object", "Job", "Task", "Operation" (for example - the list is not exhaustive).
+  * What happens to the names when the focus of the service expands?  It may be worth starting with a less generic name to avoid a breaking change later on.
+* Use previews to get the shape right.  There are different notification, breaking change, and lifetime requirements on preview API versions.
+  * We recommend a minimum of 2 preview versions prior to your first GA release.  
+  * Your API should be in preview for a minimum of 3 months to gain customer insights on usage.
+
+Breaking changes are a source of concern during initial review.  Without a history, the review process attempts to identify patterns that may result in breaking changes later on.  For instance:
+
+* Return collections with server-side paging, even if your resource does not currently need paging.  This avoids a breaking change when your service expands.
```
JeffreyRichter: I feel that this is a bit too strict. It may well be known that some API will never return a large collection and therefore a non-paging API is better. For example, Storage allows up to 5 access policies on a container: https://docs.microsoft.com/en-us/rest/api/storageservices/set-container-acl. Or some APIs might return # of countries that are supported (max countries today is 195).

azure/Guidelines.md
```suggestion
Preventing future breaking changes is a source of concern during initial review.  Without a history, the review process attempts to identify patterns that may result in breaking changes later on.  For instance:
```

azure/Guidelines.md
```diff
@@ -18,6 +20,35 @@ Teams building ARM Resource Providers (RPs) MUST follow the additional guidance
 
 ARM RPs are Azure Fundamentals requirement for Azure Services and ARM RP review is another mandatory review. Some of the guidance overlaps with general API review, but passing one review will generally make the other one go very quickly.
 
+### Advice for new services
+
+Developing a new service requires the development of at least 1 (management plane) API and potentially one or more additional (data plane) APIs.  
+
+* Think about naming from the context of a **developer experience**.
```
adrianhall, line 29: [RyanSweet] Think about the code that the customer will write (in at least two languages - a dynamic language like Python or JavaScript, and a static language like Java or C#) both before and after the REST API call.  How will a developer use this API in the standard use case?

azure/Guidelines.md
tg-msft: I think having a blurb about forward-compat and clients would be great to include because I've seen a lot of folks surprised by it.

azure/Guidelines.md
```diff
@@ -18,6 +20,35 @@ Teams building ARM Resource Providers (RPs) MUST follow the additional guidance
...
+  * Avoid the use of generic names like "Object", "Job", "Task", "Operation" (for example - the list is not exhaustive).
+  * What happens to the names when the focus of the service expands?  It may be worth starting with a less generic name to avoid a breaking change later on.
+* Use previews to get the shape right.  There are different notification, breaking change, and lifetime requirements on preview API versions.
+  * We recommend a minimum of 2 preview versions prior to your first GA release.  
+  * Your API should be in preview for a minimum of 3 months to gain customer insights on usage.
+
+Breaking changes are a source of concern during initial review.  Without a history, the review process attempts to identify patterns that may result in breaking changes later on.  For instance:
+
+* Return collections with server-side paging, even if your resource does not currently need paging.  This avoids a breaking change when your service expands.
+* Think about how the developer can reason about the collection organization.  Filtering is a common customer request.
```
JeffreyRichter: A collection should return a homogeneous collection; not a heterogeneous collection. Here's an example for an API that returns directories & files which is very hard to iterate over: https://docs.microsoft.com/en-us/rest/api/storageservices/list-directories-and-files#response-body

azure/Guidelines.md
assafi: Should we expend this and explain the notion of Idempotency?

azure/Guidelines.md
JeffreyRichter: Yes this is true. I'd to use the term "enumerations" for a forever set of fixed values. And never use the term "enumerations" for non-fixed values; just use strings.

azure/Guidelines.md
adrianhall: Try to keep verbs in the present (not past or future)

azure/Guidelines.md
```diff
@@ -18,6 +20,35 @@ Teams building ARM Resource Providers (RPs) MUST follow the additional guidance
 
 ARM RPs are Azure Fundamentals requirement for Azure Services and ARM RP review is another mandatory review. Some of the guidance overlaps with general API review, but passing one review will generally make the other one go very quickly.
 
+### Advice for new services
+
+Developing a new service requires the development of at least 1 (management plane) API and potentially one or more additional (data plane) APIs.  
+
+* Think about naming from the context of a **developer experience**.  
+  * Start with the "things" your API manipulates, then think about the operations that a developer needs to do to these "things".
+  * Avoid the use of generic names like "Object", "Job", "Task", "Operation" (for example - the list is not exhaustive).
```
adrianhall, line 32: Add a taxonomy of acceptable generic names and their meaning

azure/Guidelines.md
```diff
@@ -18,6 +20,35 @@ Teams building ARM Resource Providers (RPs) MUST follow the additional guidance
 
 ARM RPs are Azure Fundamentals requirement for Azure Services and ARM RP review is another mandatory review. Some of the guidance overlaps with general API review, but passing one review will generally make the other one go very quickly.
 
+### Advice for new services
+
+Developing a new service requires the development of at least 1 (management plane) API and potentially one or more additional (data plane) APIs.
```
assafi: Add general advise: plan for multiple previews. Gather feedback and iterate. Don't assume a single preview -> GA. 
No feedback != No required changes. Actively solicit feedback.

azure/Guidelines.md
adrianhall: Think about idempotency.  In a distributed cloud, each HTTP call must be idempotent.  You must be resilient in the face of failure.  This is the only way that developers can build fault-tolerant systems. (Link to document - Jeffrey Richter to provide)

azure/Guidelines.md
adrianhall: Arguments for strict: Consistency - all collections act the same.  
Arguments against strict: Customer must write a loop

Replace "For instance:" => "For instance, think about the following:"

azure/Guidelines.md
adrianhall: Avoid Polymorphism (not just in collections - but the API in general).  It makes designing APIs for statically typed languages like C# and Java very hard.

azure/Guidelines.md
adrianhall: Developers can prefer PUT/PATCH to specify the name/id of the entity, rather than POST without a name/id.

azure/Guidelines.md
```diff
@@ -18,6 +20,35 @@ Teams building ARM Resource Providers (RPs) MUST follow the additional guidance
...
+
+Developing a new service requires the development of at least 1 (management plane) API and potentially one or more additional (data plane) APIs.  
+
+* Think about naming from the context of a **developer experience**.  
+  * Start with the "things" your API manipulates, then think about the operations that a developer needs to do to these "things".
+  * Avoid the use of generic names like "Object", "Job", "Task", "Operation" (for example - the list is not exhaustive).
+  * What happens to the names when the focus of the service expands?  It may be worth starting with a less generic name to avoid a breaking change later on.
+* Use previews to get the shape right.  There are different notification, breaking change, and lifetime requirements on preview API versions.
+  * We recommend a minimum of 2 preview versions prior to your first GA release.  
+  * Your API should be in preview for a minimum of 3 months to gain customer insights on usage.
```
adrianhall: This should not be a minimum amount of time - rather that a representative number of developers have coded against the API.  Look for meaningful feedback from the preview developers.

azure/Guidelines.md
adrianhall: Recommendation is JSON Merge Patch (RFC 7396) over JSON Patch.

adrianhall (commented):
Add commentary about:
* Management plane vs. Data plane (Management plane is done by administrators via ARM, data plane by users)
* If it can be done by both, put the operation in both places.

azure/Guidelines.md
adrianhall: Per the normal review cycle, we start with less generic names, saving the generic names for more generalized cases later on.

azure/Guidelines.md
```diff
@@ -18,6 +20,50 @@ Teams building ARM Resource Providers (RPs) MUST follow the additional guidance
...
+  * Consider multiple languages, and include at least one dynamically typed language (for example, Python or JavaScript) and one statically typed language (for example, Java or C#).
+* Use previews to get the shape right.  There are different notification, breaking change, and lifetime requirements on preview API versions.
+  * We recommend a minimum of 2 preview versions prior to your first GA release.  However, there is no hard rule for previews.  
+  * You should gather feedback from your customers and iterate until the API is useful.  Actively solicit feedback from your preview customers.
+
+Preventing future breaking changes is a source of concern during initial review.  Without a history, the review process attempts to identify patterns that may result in breaking changes later on.  For instance, think about the following:
+
+* [Think about idempotency](https://stripe.com/blog/idempotency).  In a distributed cloud, each HTTP call must be idempotent.  You must be resilient in the face of failure.  A developer must rely on idempotency to build fault-tolerant systems.
```
RezaJooyandeh, line 42: Is [client request id](https://github.com/Azure/azure-sdk-for-net-pr/blob/feature/communication/sdk/communication/Azure.Communication.Configuration/tests/SessionRecords/Sample1_UserTokenClient/GetUserToken.json#L15) designed for this or should we have an explicit argument in which case should we make that standardize? (specially for POST requests).

azure/Guidelines.md
```diff
@@ -18,6 +20,50 @@ Teams building ARM Resource Providers (RPs) MUST follow the additional guidance
...
+
+> A **management plane** API is implemented through the Azure Resource Manager (ARM) and is used by subscription administrators.  A **data plane** API is used by developers to implement applications.  Rarely, a subset of operations may be useful to both administrators and users, in which case it should appear in both APIs.
+
+* Think about naming from the context of a **developer experience**.  
+  * Start with the "things" your API manipulates, then think about the operations that a developer needs to do to these "things".
+  * Keep the verbs present-tense.  Avoid the use of past or future tense in most cases.
+  * Avoid the use of generic names like "Object", "Job", "Task", "Operation" (for example - the list is not exhaustive).
+  * What happens to the names when the focus of the service expands?  It may be worth starting with a less generic name to avoid a breaking change later on.
+* Think about the code that a customer will write both before and after the REST API call.  How will a developer use this API in the canonical use case?
```
TomMilos, line 34: I'd consider adding a point about improving "discoverability" or "authoring" of the payload - i.e. structure it in a way that is self explanatory/intelli-sensible rather than require to consult documentation. Two examples:
1. Arrays of polymorphic objects - (usually) can be replaced with multiple arrays of objects of a single type (instead of ```{ fruits: [] }``` you can have  ```{ apples: [], oranges: [], bananas: [] }```. Frequently those would also be processed differently anyway (otherwise they wouldn't be different)
2. When having strings which can accept any value but your service predefines some well known values (e.g. perf counter names) model them as "extensible enums" - this way they will be present in the schema and generated SDKs

azure/Guidelines.md
adrianhall: Client Request ID is regenerated on each request, so it is not resilient for idempotency.

PR #212, opened 2020-08-04
Hotfix/casing

OmerCeyhan:


OmerCeyhan:
Refactor of innerError property as camel case

Issue #217, opened 2020-09-01
NotFound question


sublimemm:
Whats your position on returning a 200 with an empty body/collection when there's a collection endpoint and the results are empty? 

People tend to think this is obviously a 404 situation, but this leaves the client unable to differentiate between the wrong endpoint and the correct endpoint but no results.

Do you have the same opinion when it's an entity endpoint vs a collection endpoint? e.g., /myThing/123 is not found. /mything/123 is a valid endpoint, it's just that entity 123 is gone.

doug-ross:
Agreed. As someone who has done some deep Apache and IIS module work, 404's were intended to indicate a URI was invalid. That's very different than responding to a query that has a zero result set.

Larkenx:
Is there an appropriate 2xx response code for an empty response body that indicates no-content? 204 is close, but semantically relates to a PUT/POST on create-update-delete https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204

cwoodruff:
Actually the purpose of 404 response code is to notify the initiator that the resource addressed in the URL is not found. If the URL was not valid I think that is another response code. I think 400 Bad Request would be the best for a wrong URL.

doug-ross:
@cwoodruff  - agreed. Probably didn't state that clearly enough. If the URI specifies a resource that was not found, it should generate a 404.

gareth-a-jones:
We would always say that should return a 200 with an empty collection.   The call succeeded, the collection exists.  There's just nothing in it.

Given we always follow the convention of having the array itself be a 'value' property on a top-level object representing the collection itself, there's also the ability to return collection metadata that is valid regardless of whether there is anything in the collection, e.g. MaxItems.

cwoodruff:
@garethj-msft I was mistaken and I agree with you with a caveat. I have always architected my APIs to return a 204 when there is an empty response body. Now if you are returning an empty set and have an empty json enclosure within your response body then yes a 200 would be correct but if your API only returns a single resource and there is no content in the response body then a 204 would be the better response code to signal the API initiator that the body is empty and no attempt should be taken to examine it.

sublimemm:
@garethj-msft  yep, that's our thinking on the collection endpoints as well. 

What about on a regular entity endpoint e.g., /mything/123 where it is a valid url, but mything 123 doesn't exist?

doug-ross:
[...]

PR #218, opened 2020-09-14
Change per Issue #217 recommendations

sublimemm:
Service providers are unclear if they should return 404s when collection searches have no results. Per the advice in the issue 217 thread the recommendation is to NOT return a 404 in these cases and instead maintain consistency in their response and signifying no results as _value_  containing an empty array.

"We would always say that should return a 200 with an empty collection. The call succeeded, the collection exists. There's just nothing in it.

Given we always follow the convention of having the array itself be a 'value' property on a top-level object representing the collection itself, there's also the ability to return collection metadata that is valid regardless of whether there is anything in the collection, e.g. MaxItems." -- Gareth Jones

### DIFF
--- a/Guidelines.md
+++ b/Guidelines.md
@@ -998,6 +998,27 @@ If a server paginates an embedded collection, it MUST include additional continu
 
 **Recordset count:** Developers who want to know the full number of records across all pages, MAY include the query parameter _$count=true_ to tell the server to include the count of items in the response.
 
+**Empty Results** When a search is performed on a collection and the result set is empty you MUST respond with a valid response body and a 200 response code. 
+In this example the filters supplied by the client resulted in a empty result set. 
+The response body is return as normal and the _value_ attribute is set to a empty collection. 
+A client MAY be expecting metadata attributes like _MaxItems_ based on the format of your responses to similar calls which produced results. 
+You SHOULD maintain consistency in your API whenever possible. 
+
+```http
+GET https://api.contoso.com/v1.0/products?$filter=(name eq 'Milk' or name eq 'Eggs') and price lt 2.55
+Accept: application/json
+
+HTTP/1.1 200 OK
+Content-Type: application/json
+
+{
+  ...,
+  "MaxItems": 0,
+  "value": []
+}
+```
+
+
 ### 9.9. Compound collection operations
 Filtering, Sorting and Pagination operations MAY all be performed against a given collection.
 When these operations are performed together, the evaluation order MUST be:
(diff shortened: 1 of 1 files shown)


Guidelines.md
```diff
@@ -1006,6 +1007,27 @@ When these operations are performed together, the evaluation order MUST be:
 2. **Sorting**. The potentially filtered list is sorted according to the sort criteria.
 3. **Pagination**. The materialized paginated view is presented over the filtered, sorted list. This applies to both server-driven pagination and client-driven pagination.
 
+### 9.10. Empty Results
+When a search is performed on a collection and the result set is empty you MUST respond with a valid response body and a 200 response code.
```
gareth-a-jones: Please don't refer to this as a search as that's confusing with an actual $search operation which is usually very different from a filtered query on a collection.

Guidelines.md
```diff
@@ -1006,6 +1007,27 @@ When these operations are performed together, the evaluation order MUST be:
 2. **Sorting**. The potentially filtered list is sorted according to the sort criteria.
 3. **Pagination**. The materialized paginated view is presented over the filtered, sorted list. This applies to both server-driven pagination and client-driven pagination.
 
+### 9.10. Empty Results
+When a search is performed on a collection and the result set is empty you MUST respond with a valid response body and a 200 response code. 
+In this example the filters supplied by the client resulted in a empty result set. 
+The response body is return as normal and the _value_ attribute is set to a empty collection.
```
gareth-a-jones: 'returned' rather than 'return'

Guidelines.md
```diff
@@ -1006,6 +1007,27 @@ When these operations are performed together, the evaluation order MUST be:
...
+GET https://api.contoso.com/v1.0/products?$filter=(name eq 'Milk' or name eq 'Eggs') and price lt 2.55
+Accept: application/json
+
+HTTP/1.1 200 OK
+Content-Type: application/json
+
+{
+  ...,
+  "MaxItems": 0,
+  "value": []
```
gareth-a-jones, line 1027: lowercase 'm'

gareth-a-jones (changes requested):
Thanks - I've a couple of comments I'd like you to address.

Guidelines.md
sublimemm: how about the word filter instead of search?

sublimemm:
@garethj-msft Hey, are there any more changes you'd like to see before merging the pull request?

gareth-a-jones:
All good now - thanks for your contribution @sublimemm!

PR #243, opened 2021-06-13
Updates to the Azure Collections guidance

mkistler:
This PR contains the Azure-specific API guidance for Collections.

### DIFF
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -477,6 +477,210 @@ Overall, this is a very brittle design that leads to a poor developer experience
 
 ### Collections
 
+The standard REST pattern for accessing a collection of resources is with a _GET_ method on the collection URL.
+This operation is commonly called the "list operation" for the resource, since the primary content of the response
+is an array (list) of the resource type.  Query parameters on the list operation may be provided to control
+which resources are returned and in what order.
+
+Resource collections can often be arbitrarily large, increasing lookup time as well as the size of the responses being sent over the wire. Therefore, it is important that list operations implement pagination.
+
+:white_check_mark: **DO** provide a list operation for each resource type.
+
+:white_check_mark: **DO** structure the response to a list operation as an object with a top-level array property to contain the collection of resources.
+
+:ballot_box_with_check: **YOU SHOULD** use _value_ as the name of the top-level array property -- there are some allowable exceptions to this guidance.
+
+:white_check_mark: **DO** implement pagination of list operation responses unless there is no possibility of a collection exceeding a size appropriate for a single response.
+In particular, include a top-level property in the response named _nextLink_ that contains an opaque URL to return the next page of results when there are additional items in the collection.
+
+:no_entry: **DO NOT** include a _nextLink_ property in the response if there are no additional items in the collection.
+
+:white_check_mark: **DO** clearly document that resources may be skipped or duplicated across pages of a paginated collection unless the operation has made special provisions to prevent this.
+
+Clients MUST be resilient to collection data being either paged or nonpaged for any given request.
+
+Example:
+```json
+{
+  "value":[
+    { "id": "Item 1","price": 99.95,"sizes": null},
+    { … },
+    { … },
+    { "id": "Item 99","price": 59.99,"sizes": null}
+  ],
+  "nextLink": "{opaqueUrl}"
+}
+```
+
+#### Query options
+
+A number of query parameters that may be supported on a list operation to control which resources are returned and in what order. A summary of these query parameters is shown in the following table, with more detail given below.  All these query parameters are optional.
+
+| Parameter name | type | description |
+| -------------- | ---- | ----------- |
+| _filter_       | string | an expression on the resource type that selects the resources to be returned |
+| _orderby_      | array of string | a list of expressions that specify the order of the returned resources |
+| _skip_         | integer | an offset into the collection of the first resource to be returned |
+| _maxpagesize_  | integer | the maximum number of resources to include in a single response |
+
+:white_check_mark: **DO** treat these query parameter names as case-sensitve.
+
+:no_entry: **DO NOT** prefix any of these query parameter names with "$" (the convention in the [OData standard](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_QueryingCollections)).
+
+:no_entry: **DO NOT** define a _top_ query parameter to limit the number of results returned.  Use the _maxpagesize_ parameter for this purpose.
+
+:white_check_mark: **DO** apply the query options to the collection in the order shown in the table above.
+
+:white_check_mark: **DO** apply _select_ or _expand_ options after applying all the query options in the table above.
+
+#### filter
+
+:heavy_check_mark: **YOU MAY** support filtering of the results of a list operation with the _filter_ query parameter.
+
+The value of the _filter_ option is a Boolean expression which is evaluated for each resource in the collection, and only items where the expression evaluates to true are included in the response.
+
+:white_check_mark: **DO** omit all resources from the collection for which the _filter_ expression evaluates to false or to null, or references properties that are unavailable due to permissions.
+
+Example: return all Products whose Price is less than $10.00
+
+```http
+GET https://api.contoso.com/v1.0/products?filter=price lt 10.00
+```
+
+##### filter operators
+
+:ballot_box_with_check: **YOU SHOULD** support the following minimal set of operators in _filter_ expressions.
+
+Operator             | Description           | Example
+-------------------- | --------------------- | -----------------------------------------------------
+Comparison Operators |                       |
+eq                   | Equal                 | city eq 'Redmond'
+ne                   | Not equal             | city ne 'London'
+gt                   | Greater than          | price gt 20
+ge                   | Greater than or equal | price ge 10
+lt                   | Less than             | price lt 20
+le                   | Less than or equal    | price le 100
+Logical Operators    |                       |
+and                  | Logical and           | price le 200 and price gt 3.5
+or                   | Logical or            | price le 3.5 or price gt 200
+not                  | Logical negation      | not price le 3.5
+Grouping Operators   |                       |
+( )                  | Precedence grouping   | (priority eq 1 or city eq 'Redmond') and price gt 100
+
+:white_check_mark: **DO** respond with an error message as defined in the [Unsupported Requests](??) section if a client includes an operator in a _filter_ expression that is not supported by the operation.
+
+:white_check_mark: **DO** use the following operator precedence for supported operators when evaluating _filter_ expressions. Operators are listed by category in order of precedence from highest to lowest. Operators in the same category have equal precedence:
+
+| Group           | Operator | Description           |
+|:----------------|:---------|:----------------------|
+| Grouping        | ( )      | Precedence grouping   |
+| Unary           | not      | Logical Negation      |
+| Relational      | gt       | Greater Than          |
+|                 | ge       | Greater than or Equal |
+|                 | lt       | Less Than             |
+|                 | le       | Less than or Equal    |
+| Equality        | eq       | Equal                 |
+|                 | ne       | Not Equal             |
+| Conditional AND | and      | Logical And           |
+| Conditional OR  | or       | Logical Or            |
+
+##### Operator examples
+The following examples illustrate the use and semantics of each of the logical operators.
+
+Example: all products with a name equal to 'Milk'
+
+```http
+GET https://api.contoso.com/v1.0/products?filter=name eq 'Milk'
+```
+
+Example: all products with a name not equal to 'Milk'
+
+```http
+GET https://api.contoso.com/v1.0/products?filter=name ne 'Milk'
+```
+
+Example: all products with the name 'Milk' that also have a price less than 2.55:
+
+```http
+GET https://api.contoso.com/v1.0/products?filter=name eq 'Milk' and price lt 2.55
+```
+
+Example: all products that either have the name 'Milk' or have a price less than 2.55:
+
+```http
+GET https://api.contoso.com/v1.0/products?filter=name eq 'Milk' or price lt 2.55
+```
+
+Example: all products that have the name 'Milk' or 'Eggs' and have a price less than 2.55:
+
+```http
+GET https://api.contoso.com/v1.0/products?filter=(name eq 'Milk' or name eq 'Eggs') and price lt 2.55
+```
+
+#### orderby
+
+:heavy_check_mark: **YOU MAY** support sorting of the results of a list operation with the _orderby_ query parameter.
+
+The value of the _orderby_ parameter is a comma-separated list of expressions used to sort the items.
+A special case of such an expression is a property path terminating on a primitive property.
+
+Each expression in the _orderby_ parameter value may include the suffix "asc" for ascending or "desc" for descending, separated from the expression by one or more spaces.
+
+:white_check_mark: **DO** sort the collection in ascending order on an expression if "asc" or "desc" is not specified.
+
+:white_check_mark: **DO** sort NULL values as "less than" non-NULL values.
+
+:white_check_mark: **DO** sort items by the result values of the first expression, and then sort items with the same value for the first expression by the result value of the second expression, and so on.
+
+:white_check_mark: **DO** use the inherent sort order for the type of the property. For example, date-time values should be sorted chronologically and not alphabetically.
+
+:white_check_mark: **DO** respond with an error message as defined in the [Unsupported Requests](??) section if the client requests sorting by a property that is not supported by the operation.
+
+For example, to return all people sorted by name in ascending order:
+
+```http
+GET https://api.contoso.com/v1.0/people?$orderBy=name
+```
+
+
+For example, to return all people sorted by name in descending order and a secondary sort order of hireDate in ascending order.
+
+```http
+GET https://api.contoso.com/v1.0/people?$orderBy=name desc,hireDate
+```
+
+Sorting MUST compose with filtering such that:
+
+```http
+GET https://api.contoso.com/v1.0/people?filter=name eq 'david'&orderby=hireDate
+```
+
+will return all people whose name is David sorted in ascending order by hireDate.
+
+##### Considerations for sorting with pagination
+
+:white_check_mark: **DO** use the same filtering options and sort order for all pages of a paginated list operation response.
+
+##### skip
+
+:heavy_check_mark: **YOU MAY** allow clients to pass the _skip_ query parameter to specify an offset into collection of the first resource to be returned.
+
+:white_check_mark: **DO** define the _skip_ parameter as an integer with a default and minimum value of 0.
+
+##### maxpagesize
+
+:ballot_box_with_check: **YOU SHOULD** allow clients to pass the _maxpagesize_ query parameter to specify the maximum number of resources to include in the response.
+
+:white_check_mark: **DO** define the _maxpagesize_ parameter as an optional integer with a default value appropriate for the collection.
+
+:white_check_mark: **DO** make clear in documentation of the _maxpagesize_ parameter that the operation may choose to return fewer resources than the value specified in a single response.
+
+:white_check_mark: **DO** apply any _skip_ value specified by the client before applying the _maxpagesize_ to the collection.
+
+> MDK: Make sure that the guidance below is captured elsewhere then remove from here.
+> Note: If the server can't honor _skip_ or _maxpagesize_ (e.g. a negative value is specified), the server MUST return an error to the client informing about it instead of just ignoring the query options.
+> This will avoid the risk of the client making assumptions about the data returned.
+
 ### API Versioning
 
 Azure services need to change over time. However, when changing a service, there are 2 requirements:
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -486,18 +486,19 @@ Resource collections can often be arbitrarily large, increasing lookup time as w
 
 :white_check_mark: **DO** provide a list operation for each resource type.
 
-:white_check_mark: **DO** structure the response to a list operation as an object with a top-level array property to contain the collection of resources.
+:white_check_mark: **DO** structure the response to a list operation as an object with a top-level array field to contain the collection of resources.
 
-:ballot_box_with_check: **YOU SHOULD** use _value_ as the name of the top-level array property -- there are some allowable exceptions to this guidance.
+:ballot_box_with_check: **YOU SHOULD** at a minimum include the id field in each resource.  Also include the _ETag_ if supported for this resource type.
+
+:ballot_box_with_check: **YOU SHOULD** use _value_ as the name of the top-level array field -- there are some allowable exceptions to this guidance.
 
 :white_check_mark: **DO** implement pagination of list operation responses unless there is no possibility of a collection exceeding a size appropriate for a single response.
-In particular, include a top-level property in the response named _nextLink_ that contains an opaque URL to return the next page of results when there are additional items in the collection.
+In particular, include a top-level field in the response named _nextLink_ that contains an opaque absolute URL to return the next page of results when there are additional items in the collection.
 
-:no_entry: **DO NOT** include a _nextLink_ property in the response if there are no additional items in the collection.
+:no_entry: **DO NOT** include a _nextLink_ field in the response if there are no additional items in the collection.
+<!-- In other words, don't include `"nextLink": null` in the response to indicate the final page. I've seen it! -->
 
 :white_check_mark: **DO** clearly document that resources may be skipped or duplicated across pages of a paginated collection unless the operation has made special provisions to prevent this.
-
-Clients MUST be resilient to collection data being either paged or nonpaged for any given request.
 
 Example:
 ```json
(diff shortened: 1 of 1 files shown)


azure/Guidelines.md
```diff
@@ -455,6 +455,8 @@ Overall, this is a very brittle design that leads to a poor developer experience
```
mikekistler, line 477: Change the language to DO / DO NOT.

azure/Guidelines.md
```diff
@@ -455,6 +455,8 @@ Overall, this is a very brittle design that leads to a poor developer experience
 
 ### Collections
 
+Note: Query parameters described in this section named _$orderBy_, _$filter_, _$top_, _$skip_, _$count_, and _$maxpagesize_ MAY omit the leading "$". These names are used for consistency with the [OData standard](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_QueryingCollections).
```
mikekistler: DO NOT prefix with $.

Remove $ everywhere.

azure/Guidelines.md
```diff
@@ -463,6 +465,15 @@ Collections that support durable identifiers MAY support delta queries.
 ### 9.2. Serialization
 Collections are represented in JSON using standard array notation.
 
+The response to a collection query MUST be an object with a top-level array property that contains the items in the collection.  The name of the array property SHOULD be _value_.
+
+Example:
+```json
+{
+  "value": [...],
```
mikekistler: Might add nextLink

azure/Guidelines.md
```diff
@@ -553,13 +565,16 @@ PUT https://api.contoso.com/v1.0/people/123
 ```
 
 ### 9.6. Sorting collections
-The results of a collection query MAY be sorted based on property values.
-The property is determined by the value of the _$orderBy_ query parameter.
 
-The value of the _$orderBy_ parameter contains a comma-separated list of expressions used to sort the items.
+A collection query MAY allow its results to be sorted based on property values.
```
mikekistler: Use "list operation".

azure/Guidelines.md
```diff
@@ -553,13 +565,16 @@ PUT https://api.contoso.com/v1.0/people/123
...
+When sorting is supported, the operation MUST define a query parameter named _$orderBy_ to specify the sort ordering.
+
+The value of the _$orderBy_ parameter is a comma-separated list of expressions used to sort the items.
 A special case of such an expression is a property path terminating on a primitive property.
 
-The expression MAY include the suffix "asc" for ascending or "desc" for descending, separated from the property name by one or more spaces.
+If a client specifies more than one _$orderBy_ parameter in the query string, the service MUST return a "400 BadRequest" response with appropriate error message.
```
mikekistler: Don't need this.

azure/Guidelines.md
```diff
@@ -602,12 +617,16 @@ GET https://api.contoso.com/v1.0/people?$filter=name eq 'david'&$orderBy=hireDat
 Will return all people whose name is David sorted in ascending order by hireDate.
 
 #### 9.6.1. Interpreting a sorting expression
-Sorting parameters MUST be consistent across pages, as both client and server-side paging is fully compatible with sorting.
+Sorting parameters MUST be consistent across pages, as both client and server-side paging is fully compatible with sorting. The service MUST reject a request for a subsequent page of results if it specifies any sorting parameters that are inconsistent with the initial collection request.
```
mikekistler: Do ensure that subsequent pages use the same ordering as the original request.

azure/Guidelines.md
```diff
@@ -602,12 +617,16 @@ GET https://api.contoso.com/v1.0/people?$filter=name eq 'david'&$orderBy=hireDat
...
 
-If a service does not support sorting by a property named in a _$orderBy_ expression, the service MUST respond with an error message as defined in the Responding to Unsupported Requests section.
+If a service does not support sorting by a property named in a _$orderBy_ expression, the service MUST respond with an error message as defined in the Unsupported Requests section.
 
 ### 9.7. Filtering
-The _$filter_ querystring parameter allows clients to filter a collection of resources that are addressed by a request URL.
+A collection query MAY allow its results to be filtered.
+When filtering is supported, the operation MUST define a query parameter named _$filter_ to specify how the results are filtered.
+
+If a client specifies more than one _$filter_ parameter in the query string, the service MUST return a "400 BadRequest" response with appropriate error message.
```
mikekistler: Fix this.

azure/Guidelines.md
```diff
@@ -638,6 +657,8 @@ not                  | Logical negation      | not price le 3.5
 Grouping Operators   |                       |
 ( )                  | Precedence grouping   | (priority eq 1 or city eq 'Redmond') and price gt 100
 
+If an operation does not support an operator in a _$filter_ expression, the service MUST respond with an error message as defined in the Unsupported Requests section.
```
mikekistler: Generalize the guidance on unsupported query parameters in the URL or HTTP section, then reference here.

azure/Guidelines.md
```diff
@@ -697,14 +718,22 @@ There are two forms of pagination that MAY be supported by RESTful APIs.
 Server-driven paging mitigates against denial-of-service attacks by forcibly paginating a request over multiple response payloads.
 Client-driven paging enables clients to request only the number of resources that it can use at a given time.
 
-Sorting and Filtering parameters MUST be consistent across pages, because both client- and server-side paging is fully compatible with both filtering and sorting.
+Sorting and Filtering parameters MUST be consistent across pages, because both client- and server-side paging is fully compatible with both filtering and sorting. The service MUST reject a request for a subsequent page of results if it specifies any sorting or filtering parameters that are inconsistent with the initial collection request.
```
mikekistler: Address as above.

azure/Guidelines.md
```diff
@@ -697,14 +718,22 @@ There are two forms of pagination that MAY be supported by RESTful APIs.
...
 
 #### 9.8.1. Server-driven paging
-Paginated responses MUST indicate a partial result by including a continuation token in the response.
-The absence of a continuation token means that no additional pages are available.
+The response to collection request MUST indicate a partial result by including a property at the top level of the response body that contains the opaque URL that will return the next page of results.
+This property SHOULD be named _nextLink_.
```
mikekistler: "MUST" or "DO".

Field or property?  We should pick one and be consistent. Let's use field.

azure/Guidelines.md
```diff
@@ -697,14 +718,22 @@ There are two forms of pagination that MAY be supported by RESTful APIs.
...
+
+> MDK: should we mention the `x-ms-pageable` extension here?
```
mikekistler: Not here.  Should be elsewhere.

azure/Guidelines.md
```diff
@@ -697,14 +718,22 @@ There are two forms of pagination that MAY be supported by RESTful APIs.
...
+
+The absence of the _nextLink_ property means that no additional pages are available.
```
mikekistler: DO NOT include the _nextLink_ property on the last page.

azure/Guidelines.md
```diff
@@ -697,14 +718,22 @@ There are two forms of pagination that MAY be supported by RESTful APIs.
...
 
 Clients MUST treat the continuation URL as opaque, which means that query options may not be changed while iterating over a set of partial results.
 
+The operation MAY allow clients to specify the page size for server-driven paging with a query parameter named _$maxpagesize_.
+The _$maxpagesize_ query parameter MUST be defined as an integer and SHOULD have a defined maximum value.
+The service SHOULD honor this preference if the specified page size is smaller than the service's maximum page size.
+
 Example:
```
mikekistler, line 640: Have a section on query options.  You MAY support these, if you do support them, they must be named this, etc.

azure/Guidelines.md
```diff
@@ -717,14 +746,21 @@ Content-Type: application/json
...
+  "nextLink": "{opaqueUrl}"
 }
 ```
 
 #### 9.8.2. Client-driven paging
-Clients MAY use _$top_ and _$skip_ query parameters to specify a number of results to return and an offset into the collection.
+An operation MAY allow clients to request pagination of the collection results.
 
-The server SHOULD honor the values specified by the client; however, clients MUST be prepared to handle responses that contain a different page size or contain a continuation token.
+Clients use the _$top_ and _$skip_ query parameters to specify a number of results to return and an offset into the collection.
```
mikekistler: skip first, then top.

azure/Guidelines.md
```diff
@@ -717,14 +746,21 @@ Content-Type: application/json
...
+
+The _$top_ and _$skip_ parameters MUST be defined as integers. The _$skip_ parameter SHOULD define a default and minimum value of 0.
+Both _$top_ and _$skip_ MUST be defined as optional. When _$top_ is not specified, the service MUST return all items of the collection, starting from _$skip_ if specified, possibly paginated by server-driven pagination.  When _$skip_ is not specified, the service MUST return items from the start of the collection, up the number in _$top_ if specified, possibly paginated by server-driven pagination.
```
mikekistler: Crisp up.

azure/Guidelines.md
```diff
@@ -717,14 +746,21 @@ Content-Type: application/json
...
+
+The server SHOULD honor the values specified by the client; however, clients MUST be prepared to handle responses that are paginated by the server as indicated by the presence of a _nextLink_ property.
```
mikekistler: Server spin

azure/Guidelines.md
```diff
@@ -754,16 +790,17 @@ The server MUST supplement any specified order criteria with additional sorts (t
 Clients MUST be prepared to deal with these discrepancies.
 The server SHOULD always encode the record ID of the last read record, helping the client in the process of managing repeated/missing results.
```
mikekistler: DO document that this is the case.

Move this up to the general section -- intro to collections.

azure/Guidelines.md
```diff
@@ -754,16 +790,17 @@ The server MUST supplement any specified order criteria with additional sorts (t
...
 
-**Combining client- and server-driven paging:** Note that client-driven paging does not preclude server-driven paging.
```
mikekistler: Don't delete this -- it must be made clear.

Add an example.

azure/Guidelines.md
```diff
@@ -754,16 +790,17 @@ The server MUST supplement any specified order criteria with additional sorts (t
...
-If the page size requested by the client is larger than the default page size supported by the server, the expected response would be the number of results specified by the client, paginated as specified by the server paging settings.
+> MDK: Encode it where?  In the _nextLink_ URL?  What about for client-driven pagination?
 
-**Page Size:** Clients MAY request server-driven paging with a specific page size by specifying a _$maxpagesize_ preference.
-The server SHOULD honor this preference if the specified page size is smaller than the server's default page size.
+**Combining client- and server-driven paging:** Note that client-driven paging does not preclude server-driven paging.
+If the page size requested by the client in _$top_ is larger than the default or specified server page size, the service MUST return the number of results specified by _$top_, paginated as specified by the server paging settings.
```
mikekistler: Could we simply?  Don't have both top and maxpagesize?

mikekistler (changes requested):
Changes needed.

mkistler:
I've just pushed a major revision to the Collections guidance that I believe incorporates all the feedback from the first review.

There are a few issues still be resolved, but hopefully it will be "tweaking" from here on out.

cc: @markweitzel @johanste @jeffreyrichter @tg-msft

azure/Guidelines.md
```diff
@@ -477,6 +477,210 @@ Overall, this is a very brittle design that leads to a poor developer experience
...
+:white_check_mark: **DO** structure the response to a list operation as an object with a top-level array property to contain the collection of resources.
+
+:ballot_box_with_check: **YOU SHOULD** use _value_ as the name of the top-level array property -- there are some allowable exceptions to this guidance.
+
+:white_check_mark: **DO** implement pagination of list operation responses unless there is no possibility of a collection exceeding a size appropriate for a single response.
+In particular, include a top-level property in the response named _nextLink_ that contains an opaque URL to return the next page of results when there are additional items in the collection.
+
+:no_entry: **DO NOT** include a _nextLink_ property in the response if there are no additional items in the collection.
+
+:white_check_mark: **DO** clearly document that resources may be skipped or duplicated across pages of a paginated collection unless the operation has made special provisions to prevent this.
```
mkistler, line 501: Should we prefer "items" over "resources" here, elsewhere?

azure/Guidelines.md
```diff
@@ -477,6 +477,210 @@ Overall, this is a very brittle design that leads to a poor developer experience
...
+A number of query parameters that may be supported on a list operation to control which resources are returned and in what order. A summary of these query parameters is shown in the following table, with more detail given below.  All these query parameters are optional.
+
+| Parameter name | type | description |
+| -------------- | ---- | ----------- |
+| _filter_       | string | an expression on the resource type that selects the resources to be returned |
+| _orderby_      | array of string | a list of expressions that specify the order of the returned resources |
+| _skip_         | integer | an offset into the collection of the first resource to be returned |
+| _maxpagesize_  | integer | the maximum number of resources to include in a single response |
+
+:white_check_mark: **DO** treat these query parameter names as case-sensitve.
```
mkistler, line 528: - We have "orderby"  and "maxpagesize" (all lower-case) but also "nextLink" (lower camel case).  Are we being consistent?

azure/Guidelines.md
```diff
@@ -477,6 +477,210 @@ Overall, this is a very brittle design that leads to a poor developer experience
...
+#### Query options
+
+A number of query parameters that may be supported on a list operation to control which resources are returned and in what order. A summary of these query parameters is shown in the following table, with more detail given below.  All these query parameters are optional.
+
+| Parameter name | type | description |
+| -------------- | ---- | ----------- |
+| _filter_       | string | an expression on the resource type that selects the resources to be returned |
+| _orderby_      | array of string | a list of expressions that specify the order of the returned resources |
+| _skip_         | integer | an offset into the collection of the first resource to be returned |
+| _maxpagesize_  | integer | the maximum number of resources to include in a single response |
```
mkistler, line 526: I dropped _select_ and _expand_ from the table but do mention them below.

azure/Guidelines.md
```diff
@@ -477,6 +477,210 @@ Overall, this is a very brittle design that leads to a poor developer experience
 
 ### Collections
 
+The standard REST pattern for accessing a collection of resources is with a _GET_ method on the collection URL.
```
mkistler, line 480: URL patterns for collections should be covered elsewhere.  If not we'll need to cover them here.

azure/Guidelines.md
```diff
@@ -477,6 +477,210 @@ Overall, this is a very brittle design that leads to a poor developer experience
...
+This operation is commonly called the "list operation" for the resource, since the primary content of the response
+is an array (list) of the resource type.  Query parameters on the list operation may be provided to control
+which resources are returned and in what order.
+
+Resource collections can often be arbitrarily large, increasing lookup time as well as the size of the responses being sent over the wire. Therefore, it is important that list operations implement pagination.
```
johanste, line 485: We should probably distinguish between server driven pagination and client driven pagination.

azure/Guidelines.md
```diff
@@ -477,6 +477,210 @@ Overall, this is a very brittle design that leads to a poor developer experience
...
+
+:white_check_mark: **DO** provide a list operation for each resource type.
+
+:white_check_mark: **DO** structure the response to a list operation as an object with a top-level array property to contain the collection of resources.
```
johanste, line 489: Provide an example.

azure/Guidelines.md
```diff
@@ -477,6 +477,210 @@ Overall, this is a very brittle design that leads to a poor developer experience
...
+
+:ballot_box_with_check: **YOU SHOULD** use _value_ as the name of the top-level array property -- there are some allowable exceptions to this guidance.
+
+:white_check_mark: **DO** implement pagination of list operation responses unless there is no possibility of a collection exceeding a size appropriate for a single response.
+In particular, include a top-level property in the response named _nextLink_ that contains an opaque URL to return the next page of results when there are additional items in the collection.
```
```suggestion
In particular, include a top-level property in the response named _nextLink_ that contains an opaque absolute URL to return the next page of results when there are additional items in the collection.
```

azure/Guidelines.md
```diff
@@ -477,6 +477,210 @@ Overall, this is a very brittle design that leads to a poor developer experience
...
+
+:no_entry: **DO NOT** include a _nextLink_ property in the response if there are no additional items in the collection.
+
+:white_check_mark: **DO** clearly document that resources may be skipped or duplicated across pages of a paginated collection unless the operation has made special provisions to prevent this.
+
+Clients MUST be resilient to collection data being either paged or nonpaged for any given request.
```
johanste: This comment is somewhat unclear to me - it could be interpreted such that a client should expect either an array OR an object with a value + nextLink URL for any given operation...

What we are trying to say is that a client should always be prepared to follow the `nextLink` to retrieve additional results....

azure/Guidelines.md
johanste: You are missing `top`, which is needed for client driven pagination. Also, we should clarify the relationship between `top` and `maxpagesize`.

azure/Guidelines.md
mikekistler: One of the points brought up in the first review was to consider simplifying our pagination guidance.  That's what I've done here.  You can think of this as "unified pagination".  The client controls the number of items in a single response with _maxpagesize_ rather than _top_.  Since the client has to handle server pagination anyway, there is no need for _top_.

azure/Guidelines.md
markweitzel: Let's be consistent within the guidelines. I don't think we use "items" anywhere else. 
I would be +1 for resources.

azure/Guidelines.md
markweitzel: Interestingly, the MS guidelines are inconsistent (maxpagesize orderBy, nextLink). We can check what is implemented today, but if we can change it, I'd vote for camel case -- maxPageSize, orderBy, nextLink.

azure/Guidelines.md
```diff
@@ -477,6 +477,210 @@ Overall, this is a very brittle design that leads to a poor developer experience
...
+
+A number of query parameters that may be supported on a list operation to control which resources are returned and in what order. A summary of these query parameters is shown in the following table, with more detail given below.  All these query parameters are optional.
+
+| Parameter name | type | description |
+| -------------- | ---- | ----------- |
+| _filter_       | string | an expression on the resource type that selects the resources to be returned |
+| _orderby_      | array of string | a list of expressions that specify the order of the returned resources |
+| _skip_         | integer | an offset into the collection of the first resource to be returned |
+| _maxpagesize_  | integer | the maximum number of resources to include in a single response |
+
```
mkistler, line 527: Another possible query option is _count_.  This isn't in Jeffrey's slides but is one of the standard OData options.  I can add that if folks think it is useful.

azure/Guidelines.md
```diff
@@ -477,6 +477,210 @@ Overall, this is a very brittle design that leads to a poor developer experience
...
+:white_check_mark: **DO** treat these query parameter names as case-sensitve.
+
+:no_entry: **DO NOT** prefix any of these query parameter names with "$" (the convention in the [OData standard](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_QueryingCollections)).
```
mkistler, line 530: Interesting data point: a quick search on azure-rest-api-specs turns up 583 instances of `"name": "$top"` but just a handful of `"name": "top"`.

azure/Guidelines.md
mkistler: If we are going to choose between _top_ and _maxpagesize_, here's something to consider.  Currently there are only two instances of a _maxpagesize_ parameter in azure-rest-api-specs, but many (> 500) instances of _top_ (actually most of these are _$top_, but morally equivalent).

tg-msft: I'd prefer `maxPageSize` if that's the semantic.  Every time we see `top` or `$top` on the client we have to have a discussion about exactly what that means.

We might also want to talk about not exposing server pagination parameters if you only need to expose `nextLink`.  Everyone gets that wrong.

azure/Guidelines.md
tg-msft: Should we talk about requirements for the resources?  At a minimum it needs to have enough information to fetch the resource (`id`/`name`/etc.).  It should ideally contain information to manage optimistic concurrency (`etag`).

azure/Guidelines.md
tg-msft: Do we want to include that sort of guidance here or https://azure.github.io/azure-sdk/general_introduction.html?

azure/Guidelines.md
tg-msft: `filter` is used for selecting which resources, but we should also suggest something for specifying which parts of a resource you want returned since that can make list operations more or less expensive.  Some services call this `include=foo|bar` where the values refer to high level concepts that span many properties, OData uses `$select` for literal projections, etc.

azure/Guidelines.md
tg-msft: I think it's fine - `nextLink` is part of the JSON body where the others are query parameters.

azure/Guidelines.md
tg-msft: There was a point when our REST Guidelines were "use OData for everything" and it's had a lasting impact.  We've been pushing newer services away from it because it's kind of frustrating for customers when you call something `$whatever` but don't implement any OData semantics.

azure/Guidelines.md
```diff
@@ -477,6 +477,210 @@ Overall, this is a very brittle design that leads to a poor developer experience
...
+
+:no_entry: **DO NOT** define a _top_ query parameter to limit the number of results returned.  Use the _maxpagesize_ parameter for this purpose.
```
tg-msft: Should we clarify  that `top` is only to be used for total results across pages if that feature is needed?

azure/Guidelines.md
```diff
@@ -477,6 +477,210 @@ Overall, this is a very brittle design that leads to a poor developer experience
...
+
+:white_check_mark: **DO** apply the query options to the collection in the order shown in the table above.
+
+:white_check_mark: **DO** apply _select_ or _expand_ options after applying all the query options in the table above.
+
+#### filter
+
+:heavy_check_mark: **YOU MAY** support filtering of the results of a list operation with the _filter_ query parameter.
+
+The value of the _filter_ option is a Boolean expression which is evaluated for each resource in the collection, and only items where the expression evaluates to true are included in the response.
```
tg-msft: `Boolean expression` is a little vague.  Should we call it `OData filter expression`?

azure/Guidelines.md
```diff
@@ -477,6 +477,210 @@ Overall, this is a very brittle design that leads to a poor developer experience
...
+
+:white_check_mark: **DO** omit all resources from the collection for which the _filter_ expression evaluates to false or to null, or references properties that are unavailable due to permissions.
+
+Example: return all Products whose Price is less than $10.00
+
+```http
+GET https://api.contoso.com/v1.0/products?filter=price lt 10.00
```
tg-msft: Can we remove the version in the path?

azure/Guidelines.md
```diff
@@ -477,6 +477,210 @@ Overall, this is a very brittle design that leads to a poor developer experience
...
+Logical Operators    |                       |
+and                  | Logical and           | price le 200 and price gt 3.5
+or                   | Logical or            | price le 3.5 or price gt 200
+not                  | Logical negation      | not price le 3.5
+Grouping Operators   |                       |
+( )                  | Precedence grouping   | (priority eq 1 or city eq 'Redmond') and price gt 100
+
+:white_check_mark: **DO** respond with an error message as defined in the [Unsupported Requests](??) section if a client includes an operator in a _filter_ expression that is not supported by the operation.
+
+:white_check_mark: **DO** use the following operator precedence for supported operators when evaluating _filter_ expressions. Operators are listed by category in order of precedence from highest to lowest. Operators in the same category have equal precedence:
```
tg-msft, line 572: Is there any easy way to slice the OData spec so we can say what we support without providing these sorts of details?  I worry that this is a slippery slope and would prefer to avoid BNF grammars for OData subsets in our guidelines.

azure/Guidelines.md
```diff
@@ -477,6 +477,210 @@ Overall, this is a very brittle design that leads to a poor developer experience
...
+#### orderby
+
+:heavy_check_mark: **YOU MAY** support sorting of the results of a list operation with the _orderby_ query parameter.
+
+The value of the _orderby_ parameter is a comma-separated list of expressions used to sort the items.
+A special case of such an expression is a property path terminating on a primitive property.
+
+Each expression in the _orderby_ parameter value may include the suffix "asc" for ascending or "desc" for descending, separated from the expression by one or more spaces.
+
+:white_check_mark: **DO** sort the collection in ascending order on an expression if "asc" or "desc" is not specified.
```
tg-msft, line 629: Should we prescribe lexicographic ordering?

azure/Guidelines.md
markweitzel: Decision: 
use top (preferred) only, esp. if you are using cosmos as a backend.
use may but try avoid using maxPageSize and only when you have to customer scenario that requires limited result sets.

azure/Guidelines.md
tg-msft: https://github.com/microsoft/api-guidelines/issues/248

azure/Guidelines.md
tg-msft: https://github.com/microsoft/api-guidelines/issues/247

azure/Guidelines.md
mkistler: How about "an expression involving the fields of the resource that produces a boolean value"?

I think we want to avoid referencing OData if possible.

azure/Guidelines.md
mkistler: A little bit below we say "use the inherent sort order for the type of the property".  We _could_ make that more specific.  Should we?

azure/Guidelines.md
mkistler: I borrowed this from the MS API Guidelines but I agree that it is out of place here.  I think we need someplace to collect this kind of guidance for clients.  But I will remove it from here.

mkistler:
I've made the changes we discussed in the review on Friday, rebased onto the latest `azureRestUpdates`, and pushed up to my fork.

Thanks @tg-msft for opening the issues to track the follow-on work.

markweitzel (approved):
Looks good to me. I'm going to merge this now. 
Thanks!

PR #249, opened 2021-06-21
mdw-cleanup-pass

markweitzel:
A quick pass cleaning up a few things.

### DIFF
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -738,7 +738,7 @@ While removing a value from an enum is a breaking change, adding an enum can be 
 }
 ```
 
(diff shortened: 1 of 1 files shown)


azure/Guidelines.md
```diff
@@ -740,7 +738,7 @@ While removing a value from an enum is a breaking change, adding an enum can be
 }
 ```
 
-Always model an enum as a string unless you are positive that the symbol set will **NEVER** change over time.
+:ballot_box_with_check: **DO** slways model an ```enum``` as a string unless you are positive that the symbol set will **NEVER** change over time.
```
mkistler:
```suggestion
:ballot_box_with_check: **DO** model an ```enum``` as a string unless you are positive that the symbol set will **NEVER** change over time.
```

mkistler (approved):
Looks good! 👍 

Fixed one small typo.

PR #250, opened 2021-06-21
az-actions

markweitzel:
First pass at adding the "actions" API pattern.

### DIFF
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -713,8 +713,10 @@ Timestamp shouldn't be returned with more than subsecond precision if you'll als
 
 
 MAY consider Weak ETags if you have a valid scenario for distinguishing between meaningful and cosmetic changes.  You can also use weak etags if it's expensive to compute an ETag (i.e., weak etag is size in bytes which might not be super accurate compared to an MD5 hash of a sizeable resource).
+
 If you choose to use a Weak ETag, then...
 (add text from mike)
+
 (Review w/team, e.g. when to use in Azure)
 
 #### Multiple conditions: If-Match && If-Unmodified-Since && (If-None-Match || If-Modified-Since)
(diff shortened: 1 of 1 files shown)


azure/Guidelines.md
```diff
@@ -524,7 +537,7 @@ While removing a value from an enum is a breaking change, adding an enum can be
 }
 ```
 
-Always model an enum as a string unless you are positive that the symbol set will **NEVER** change over time.
+:ballot_box_with_check: **DO** slways model an ```enum``` as a string unless you are positive that the symbol set will **NEVER** change over time.
```
mkistler:
```suggestion
:ballot_box_with_check: **DO** model an ```enum``` as a string unless you are positive that the symbol set will **NEVER** change over time.
```

azure/Guidelines.md
```diff
@@ -474,8 +472,23 @@ Overall, this is a very brittle design that leads to a poor developer experience
 ## Common API Patterns
 
 ### Performing an Action
+The REST specificaiton is used to model the state of a resource, and is primarily intended to handle CRUD (Create, Read, Update, Delete) operations. However, many services have the requirement to perform an action on a resource, e.g. getting the thumbnail of an image, sending an SMS message. To perform an action on a resource, we introduce the concept of a "controller resource," or simply "controller."
```
mkistler, line 475: I think the notion of a "controller resource" is confusing.  These are _not_ "resources" in the way we generally use this term in REST APIs -- you cannot "create" or "delete" or "get" them.  Why not simply call these what they are: "actions"?
> To perform an action on a resource, issue a request using the resource URL with a single additional path segment with the name of the action.

mkistler (changes requested):
I think "controller resources" are overly complicated and confusing.  Let's simply our guidance by just describing "actions".

azure/Guidelines.md
markweitzel: +1. I like your simplified version *much* better and will include it. In this case, I think I pulled too much from the existing literature.

PR #251, opened 2021-06-21
az-guidelines - First pass at summary section

markweitzel:
First pass at the summary section.


azure/Guidelines.md
```diff
@@ -951,12 +951,15 @@ o	Consider forcing conditional headers on resource mutation, then use 428 Precon
...
-
-
-## Final Thoughts / Summary
-* Careful consideration up front
-* Long term decisions that are often codified in SDKs, CODE, etc.
-* Reach out and engage the stewardship team!
+## Final thoughts
+These guidelines describe the ..., and the common patterns that teams encounter when building an Azure service. There is a great deal of information in them that can be difficult to follow. Fortunately, at Microsoft, there is a team committed to ensuring your success. 
+
+The Azure REST API Stewardship board is a collection of dedicated architects that are passionate about helping service teams build interfaces that are intuitive, maintainable, consistent, and most importantly, delight our customers. Because APIs affect nearly all downstream decisions, you are encouraged to reach out to the Stewardship board early in the development process. These architects will work with you to apply these guidelines and identify any hidden pitfalls in your design.
```
mkistler: Is this an offer we are making to the whole world?  If so, we will need a lot more resources ;-).  If not, we should probably leave this out of the public document.

mkistler (commented):
While I think this content is fine, I think we should find a different forum to communicate it since I think this is really only directed at teams within Microsoft.

azure/Guidelines.md
markweitzel: I think this text could be a bit refined to focus on the service team since they are, ultimately, responsible for the API. It would be great if our community would provide input/comments/revisions via PRs on the specs in the repos. If the service team wants to accept that input and/or needs guidance on that PR, then I think we should follow the process we've outlined.

PR #278, opened 2021-08-08
Simplify BYOS section

JeffreyRichter:


azure/Guidelines.md
```diff
@@ -760,79 +760,64 @@ while the Microsoft guidelines use the name "Operation-Location".
 
 :white_check_mark: **DO** look for **both** HEADERS in client code, preferring the `Operation-Location` version.
 
-### Bring your own Storage
-When implementing your service, it is very common to store and retrieve data and files. When you encounter this scenario, avoid implementing your own storage strategy and instead use Azure Bring Your Own Storage (BYOS). BYOS provides significant benefits to service implementors, e.g. security, an aggressively optimized frontend, uptime, etc.
-While Azure Managed Storage may be easier to get started with, as your service evolves and matures, BYOS will provide the most flexibility and implementation choices. Further, when designing your APIs, be cognizant of expressing storage concepts and how clients will access your data. For example, if you are working with blobs, then you should not expose the concept of folders, nor do they have extensions.
+### Bring your own Storage (BYOS)
+Many services need to store and retrieve data files. For this scenario, teh service should not implement its own
```
```suggestion
Many services need to store and retrieve data files. For this scenario, the service should not implement its own 
```

johanste: My strong opinion is that BYOS should be an option not a requirement. Forcing the user to create and provide a storage account with the required permissions etc. drastically raises the bar for starting to use a simple service. This is also what service teams find when they talk to customers. Both service managed storage and BYOS are valid scenarios.

azure/Guidelines.md
```diff
@@ -760,79 +760,64 @@ while the Microsoft guidelines use the name "Operation-Location".
...
+storage mechanism and should instead leverage the existing Azure Storage service. When doing this, the customer 
+"owns" the storage account and just tells your service to use it. Colloquially, we call this <i>Bring Your Own Storage</i> as the customer is bringing their storage account to another service. BYOS provides significant benefits to service implementors: security, performance, uptime, etc. And, of course, most Azure customers are already with the Azure Storage service. 
 
-:white_check_mark: **DO** use Azure Bring Your Own Storage.
+While Azure Managed Storage may be easier to get started with, as your service evolves and matures, BYOS provides the most flexibility and implementation choices. Further, when designing your APIs, be cognizant of expressing storage concepts and how clients will access your data. For example, if you are working with blobs, then you should not expose the concept of folders.
 
-:white_check_mark: **DO** use a blob prefix
+:white_check_mark: **DO** use the Bring Your Own Storage pattern
+
+:white_check_mark: **DO** use a blob prefix for logical folders
```
johanste: What does this mean?

azure/Guidelines.md
tg-msft: I read this as warning against adding your own API façade on top of Blobs like we see so many teams do.  If they want to go through the work for Managed Storage, that's great and will improve the getting started experience for their customers.  But if they try to add their own `UploadFile` API for files larger than around 256MB, they're probably going to do a much worse job than the Storage team's end to end developer experience and should just accept a SAS URI to a blob.

azure/Guidelines.md
JeffreyRichter: I'm afraid that teams don't really understand how storage works. Specifically, that there are containers but no subcontainers and that there aren't sub-directories either. There are just blob names which have delimiters in them (like / or \) and the customer defines the delimiter. So, I agree that I have not said this well.  How about:
**DO** use a URL indicating a path to a root directory.

This doesn't explain anything but I did add a link to the List Blobs API docs in one bullet so maybe engineers will ready that to understand how it really works.

azure/Guidelines.md
```diff
@@ -760,79 +760,64 @@ while the Microsoft guidelines use the name "Operation-Location".
...
-While Azure Managed Storage may be easier to get started with, as your service evolves and matures, BYOS will provide the most flexibility and implementation choices. Further, when designing your APIs, be cognizant of expressing storage concepts and how clients will access your data. For example, if you are working with blobs, then you should not expose the concept of folders, nor do they have extensions.
+### Bring your own Storage (BYOS)
+Many services need to store and retrieve data files. For this scenario, the service should not implement its own 
+storage mechanism and should instead leverage the existing Azure Storage service. When doing this, the customer 
+"owns" the storage account and just tells your service to use it. Colloquially, we call this <i>Bring Your Own Storage</i> as the customer is bringing their storage account to another service. BYOS provides significant benefits to service implementors: security, performance, uptime, etc. And, of course, most Azure customers are already with the Azure Storage service.
```
```suggestion
"owns" the storage account and just tells your service to use it. Colloquially, we call this <i>Bring Your Own Storage</i> as the customer is bringing their storage account to another service. BYOS provides significant benefits to service implementors: security, performance, uptime, etc. And, of course, most Azure customers are already familiar with the Azure Storage service. 
```

azure/Guidelines.md
tg-msft: Is it worth adding "instead of APIs called `directory`, `folder`, or `path`" to make that clearer?

azure/Guidelines.md
JeffreyRichter: Yes, I was trying to say what Ted is saying (don't write your own API on top of Azure storage).
I'm OK if teams want to use managed storage too. We can add a sentence to this effect in the intro but then the team is writing a storage API. We can discuss but I'm OK with all of this.

azure/Guidelines.md
```diff
@@ -760,79 +760,64 @@ while the Microsoft guidelines use the name "Operation-Location".
...
+:white_check_mark: **DO** include a JSON object that has string values for "location" and "delimiter". For "location", the customer must pass a URL to a blob prefix which represents a directory. For "delimiter", the customer must specify the delimiter character they desire to use in the location URL; typically "/" or "\". 
 
-:white_check_mark: **DO** use a URL to a blob prefix with a container scoped SAS on the end with a minimum of `listing` and `read` permissions.
+:heavy_check_mark: **YOU MAY** support the "lastModified" field for input directories (see guideline above).
 
-For each of the "output" sections the following apply:
+:white_check_mark: **DO** support a "location" URL with a container-scoped SAS that has a minimum of `listing` and `read` permissions for input directories.
 
-:white_check_mark: **DO** use a URL to a blob prefix with a container scoped SAS on the end with a minimum of `write` permissions
+:white_check_mark: **DO** support a "location" URL with a container-scoped SAS that has a minimum of `write` permissions for output directories.
```
tg-msft, line 956: What are your thoughts on [`write` vs. `create`](https://docs.microsoft.com/en-us/rest/api/storageservices/create-service-sas#permissions-for-a-directory-container-or-blob) here?  Is `create` a little safer if you don't want to trash anything already in `output.location`?  Or should we recommend `write` because it'll always get the job done?

tg-msft (approved):
Looks great - I think this adds more clarity.

azure/Guidelines.md
```diff
@@ -760,79 +760,64 @@ while the Microsoft guidelines use the name "Operation-Location".
...
+:heavy_check_mark: **YOU MAY** use Shared Access Signatures [SAS](https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview) to provide access to Storage objects.
 
 :ballot_box_with_check: **YOU SHOULD** if using HTTP (not HTTPS) document to users that all information is sent over the wire in clear text.
 
-:ballot_box_with_check: **YOU SHOULD** support managed identity using Azure Storage by default (if using Azure services).
+:white_check_mark: **DO** return an HTTP status code representing the result of your service operation's behavior. If the operation uses Storage internally, this an implementation detail from the client's perspective of the operation but may happen due to a missing Storage object or insufficient permissions. If the internal Storage operation fails, include the Storage error information in the 'inner-error' section of your operation's response body to help the client know that they may have configured their Storage improperly.
```
assafi: I feel like we might want to expend on this a bit in all manners regarding permissions, which is probably the No. 1 issue customers will encounter. Specifically, errors regarding setting service principles, CMK setup and key rotation, etc. Having a consistent experience here across service will be great.

azure/Guidelines.md
```diff
@@ -760,79 +760,64 @@ while the Microsoft guidelines use the name "Operation-Location".
...
+:heavy_check_mark: **YOU MAY** allow the customer to provide a [last-modified](https://datatracker.ietf.org/doc/html/rfc7232#section-2.2) timestamp (in RFC1123 format) for read-only files. This allows the client to specify exactly which version of the files your service should use. When reading a file, your service passes this timestamp to Azure Storage using the [if-unmodified-since](https://datatracker.ietf.org/doc/html/rfc7232#section-3.4) request header. If the Storage operation fails with 412, the Storage object was modified and your service operation should return an appropriate 4xx status code and return the Storage error in your operation's 'inner-error' (see guideline above).
 
-:white_check_mark: **DO** When using a Shared Access Signature (SAS), ensure this is assigned to the container and that the permissions apply to the content as well.
+:white_check_mark: **DO** allow the customer to specify a URL path to a logical directory (via prefix and delimiter) if your service requires access to multiple files (within this directory). For more information, see [List Blobs API](https://docs.microsoft.com/en-us/rest/api/storageservices/list-blobs)
```
assafi: Should we expand about file/path selectors? like file prefixes/suffixes (e.g. `*.pdf`).

azure/Guidelines.md
markweitzel: +1 on Ted's suggestions.

azure/Guidelines.md
markweitzel: @assafi , would you expand on your scenario a bit? I'm not sure I'm fully understanding the issue/question you are raising.

markweitzel (commented):
@JeffreyRichter -- This is a good clarification/cleanup of the BYOS section. There are several open questions that I'd like to work with you/Assaf/Ted to close before I approve this.

azure/Guidelines.md
assafi: In some cases APIs will point to a folder but will only want/allow to process certain files. For example, [Document Translator API](https://docs.microsoft.com/en-us/azure/cognitive-services/translator/document-translation/reference/start-translation#inputs) let's users filter which files they want to translate.

azure/Guidelines.md
```diff
@@ -760,79 +760,64 @@ while the Microsoft guidelines use the name "Operation-Location".
 
 :white_check_mark: **DO** look for **both** HEADERS in client code, preferring the `Operation-Location` version.
 
-### Bring your own Storage
-When implementing your service, it is very common to store and retrieve data and files. When you encounter this scenario, avoid implementing your own storage strategy and instead use Azure Bring Your Own Storage (BYOS). BYOS provides significant benefits to service implementors, e.g. security, an aggressively optimized frontend, uptime, etc.
-While Azure Managed Storage may be easier to get started with, as your service evolves and matures, BYOS will provide the most flexibility and implementation choices. Further, when designing your APIs, be cognizant of expressing storage concepts and how clients will access your data. For example, if you are working with blobs, then you should not expose the concept of folders, nor do they have extensions.
+### Bring your own Storage (BYOS)
+Many services need to store and retrieve data files. For this scenario, teh service should not implement its own 
+storage mechanism and should instead leverage the existing Azure Storage service. When doing this, the customer
```
```suggestion
storage APIs and should instead leverage the existing Azure Storage service. When doing this, the customer 
```

azure/Guidelines.md
JeffreyRichter:
```suggestion
:white_check_mark: **DO** use a blob prefix for a logical folder (avoid terms such as ```directory```, ```folder```, or ```path```).
```

azure/Guidelines.md
JeffreyRichter: We think what you're suggesting is already covered by the text. If you think more should be here, can you please provide the actual text you'd like to see? Thanks.

azure/Guidelines.md
JeffreyRichter:
```suggestion
:white_check_mark: **DO** allow the customer to specify a URL path to a logical folder (via prefix and delimiter) if your service requires access to multiple files (within this folder). For more information, see [List Blobs API](https://docs.microsoft.com/en-us/rest/api/storageservices/list-blobs)

:heavy_check_mark: **YOU MAY** offer an ```extensions``` field representing an array of strings indicating file extensions of desired blobs within the logical folder. 
```

azure/Guidelines.md
```diff
@@ -760,79 +760,64 @@ while the Microsoft guidelines use the name "Operation-Location".
...
-
-For example, in a request payload may look similar to the following:
+For example, a service's request body to configure BYOS may look like this:
 
 ```json
 {
-"input":{
+  "input":{
     "location": "https://mycompany.blob.core.windows.net/documents/english/?<sas token>",
+    "delimiter": "/",
```
JeffreyRichter, line 937:
```suggestion
    "delimiter": "/",
    "extensions" : [ ".bmp", ".jpg", ".tif", ".png" ],
```

azure/Guidelines.md
JeffreyRichter: We agreed to leave this as is

azure/Guidelines.md
```diff
@@ -760,79 +760,67 @@ while the Microsoft guidelines use the name "Operation-Location".
...
+### Bring your own Storage (BYOS)
+Many services need to store and retrieve data files. For this scenario, the service should not implement its own 
+storage APIs and should instead leverage the existing Azure Storage service. When doing this, the customer 
+"owns" the storage account and just tells your service to use it. Colloquially, we call this <i>Bring Your Own Storage</i> as the customer is bringing their storage account to another service. BYOS provides significant benefits to service implementors: security, performance, uptime, etc. And, of course, most Azure customers are already familiar with the Azure Storage service. 
 
-:white_check_mark: **DO** use Azure Bring Your Own Storage.
+While Azure Managed Storage may be easier to get started with, as your service evolves and matures, BYOS provides the most flexibility and implementation choices. Further, when designing your APIs, be cognizant of expressing storage concepts and how clients will access your data. For example, if you are working with blobs, then you should not expose the concept of folders.
```
mikekistler, line 887: Is “Azure Managed Storage” different from the “Azure Storage service”? If so, will that be obvious to readers?  Do we need to clarify that?

azure/Guidelines.md
```diff
@@ -760,79 +760,67 @@ while the Microsoft guidelines use the name "Operation-Location".
 
 :white_check_mark: **DO** look for **both** HEADERS in client code, preferring the `Operation-Location` version.
 
-### Bring your own Storage
-When implementing your service, it is very common to store and retrieve data and files. When you encounter this scenario, avoid implementing your own storage strategy and instead use Azure Bring Your Own Storage (BYOS). BYOS provides significant benefits to service implementors, e.g. security, an aggressively optimized frontend, uptime, etc.
-While Azure Managed Storage may be easier to get started with, as your service evolves and matures, BYOS will provide the most flexibility and implementation choices. Further, when designing your APIs, be cognizant of expressing storage concepts and how clients will access your data. For example, if you are working with blobs, then you should not expose the concept of folders, nor do they have extensions.
+### Bring your own Storage (BYOS)
+Many services need to store and retrieve data files. For this scenario, the service should not implement its own 
+storage APIs and should instead leverage the existing Azure Storage service. When doing this, the customer
```
mikekistler, line 884: Should we make “Azure Storage service” a link to the service documentation?

azure/Guidelines.md
```diff
@@ -760,79 +760,67 @@ while the Microsoft guidelines use the name "Operation-Location".
...
+:heavy_check_mark: **YOU MAY** use Shared Access Signatures [SAS](https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview) to provide access to Storage objects.
 
 :ballot_box_with_check: **YOU SHOULD** if using HTTP (not HTTPS) document to users that all information is sent over the wire in clear text.
 
-:ballot_box_with_check: **YOU SHOULD** support managed identity using Azure Storage by default (if using Azure services).
+:white_check_mark: **DO** return an HTTP status code representing the result of your service operation's behavior. If the operation uses Storage internally, this an implementation detail from the client's perspective of the operation but may happen due to a missing Storage object or insufficient permissions. If the internal Storage operation fails, include the Storage error information in the 'inner-error' section of your operation's response body to help the client know that they may have configured their Storage improperly.
```
mikekistler: I think we can make this a little more concise:
```suggestion
:white_check_mark: **DO** include the Storage error information in the `inner-error` section of an error response if the error was the result of an an internal Storage operation failure.  This may help the client determine the underlying cause of the error, e.g a missing Storage object or insufficient permissions.
```

mikekistler (approved):
I left a few questions and a suggestion that I hope are useful.

Overall the content is an improvement so I’ll approve.

azure/Guidelines.md
JeffreyRichter:
```suggestion
storage APIs and should instead leverage the existing [Azure Storage service](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blobs-introduction). When doing this, the customer 
```

azure/Guidelines.md
JeffreyRichter:
```suggestion
:white_check_mark: **DO** return an HTTP status code representing the result of your service operation's behavior. 

:white_check_mark: **DO** include the Storage error information in the 'inner-error' section of an error response if the error was the result of an internal Storage operation failure. This helps the client determine the underlying cause of the error, e.g.: a missing storage object or insufficient permissions. 
```

PR #282, opened 2021-08-23
Integrate LRO Guidance to make Guidelines self-contained

### LINKED ISSUE #266: Refine Guidance on Long Running Operations (LRO)
What is in the current rev of the specification is an improvement and a good start. However, we'd like to add more guidance in the area of LROs.

mkistler:
This PR adds Azure-specific guidance on long-running operations rather than reference the more generic content on LRO in the Microsoft Guidelines.  This makes the Azure guidelines self-contained which will be easier for service teams to consume.

Fixes #266


azure/Guidelines.md
```diff
@@ -747,18 +747,66 @@ The ability to retry failed requests for which a client never received a respons
...
+
+Some common situations where the RELO pattern should be used:
+1. A "create" operation for a resource where the basic structure of the resource is created immediately and includes a status field that indicates when the create has completed, e.g. "provisioning" -> "active".
+2. An action operation for a resource where both the initiation of the action and the completion of the action cause a change to the "status" property of the resource.
+
+:white_check_mark: **DO** return a `200-OK` response, `201-Created` for create operations, from the request that initiates the operation.  The response body should contain a representation of the resource that clearly indicates that the operation has been accepted or started.
+
+:white_check_mark: **DO** support a get method on the resource that returns a representation of the resource including the status field that indicates when the operation has completed.
+
+#### Stepwise long-running operations
```
mkistler: We should try to come up with a better name than "Stepwise" LRO.

azure/Guidelines.md
```diff
@@ -747,18 +747,66 @@ The ability to retry failed requests for which a client never received a respons
...
-while the Microsoft guidelines use the name "Operation-Location".
+:white_check_mark: **DO** implement an operation as an LRO if the 99th percentile response time is greater than 1s.
 
-:white_check_mark: **DO** support both `Azure-AsyncOperation` and `Operation-Location` HEADERS, even though they are redundant so that existing SDKs and clients will continue to operate.
+There are two basic patterns that can be used for long-running operations:
+1. Resource-based long-running operations (RELO)
+2. Stepwise long-running operations
+
+:white_check_mark: **DO** use the RELO pattern when the operation is on a resource that contains a "status" property that can be used to determine the outcome of the operation.
+
```
mkistler, line 796: I think we need to add guidance on the name of the status field and terminal values for RELO too.

azure/LongRunningOperations.md
```diff
@@ -0,0 +1,69 @@
+# Long-Running Operations
```
mkistler: I made this a separate document for now -- I figured we could work out what the content should be and then decide where it should live.

azure/Guidelines.md
```diff
@@ -747,18 +747,66 @@ The ability to retry failed requests for which a client never received a respons
 <!-- Open API Spec -->
 [OpenAPI Specification]: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md
 
-### Long Running Operations & Jobs
+### Long-Running Operations & Jobs
 
-Azure generally follows the [Microsoft REST API guidelines for Long running operations](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#13-long-running-operations).  Follow the Microsoft guidelines with the modifications / extensions given here when designing / implementing long running operations.
+When the processing for an operation may take a significant amount of time to complete, it should be
+implemented as a _long-running operation (LRO)_. This allows clients to continue running while the
+operation is being processed. The client determines the outcome of the operation at some later time
```
markweitzel: @mikekistler -- I'm reacting to the word "determines" in this sentence. When I first read it, it sounded like the client can affect the outcome of the operation through an API call at a later time. What about something like...

```suggestion
operation is being processed. The client discovers the outcome of the operation at some later time
```

azure/Guidelines.md
markweitzel: I did a quick google search and didn't come up with anything better.

azure/Guidelines.md
```diff
@@ -747,18 +747,66 @@ The ability to retry failed requests for which a client never received a respons
...
+
+:heavy_check_mark: **YOU MAY** support a `get` method on the operation-results collection endpoint that returns a list of all recently initiated operations.
+
+:no_entry: **DO NOT** return any other `2xx` status code from the initial request of a stepwise LRO -- return `202-Accepted` and an operations-result URL even if processing was completed before the initiating request returns.
+
+:no_entry: **DO NOT** return any data in the response body of a `202-Accepted` response.
+
+Previous Azure guidelines specified "Azure-AsyncOperation" as the name of the response header containing the operation-results URL.
+
+:white_check_mark: **DO** return both `Azure-AsyncOperation` and `Operation-Location` headers if your service previously returned `Azure-AsyncOperation`, even though they are redundant, so that existing clients will continue to operate.
```
```suggestion
:white_check_mark: **DO** return **both** `Azure-AsyncOperation` and `Operation-Location` headers if your service previously returned `Azure-AsyncOperation`, even though they are redundant, so that existing clients will continue to operate.
```

azure/Guidelines.md
```diff
@@ -747,18 +747,66 @@ The ability to retry failed requests for which a client never received a respons
...
+:white_check_mark: **DO** implement an operation as an LRO if the 99th percentile response time is greater than 1s.
 
-:white_check_mark: **DO** support both `Azure-AsyncOperation` and `Operation-Location` HEADERS, even though they are redundant so that existing SDKs and clients will continue to operate.
+There are two basic patterns that can be used for long-running operations:
+1. Resource-based long-running operations (RELO)
+2. Stepwise long-running operations
+
+:white_check_mark: **DO** use the RELO pattern when the operation is on a resource that contains a "status" property that can be used to determine the outcome of the operation.
+
+:ballot_box_with_check: **YOU SHOULD** use the stepwise LRO pattern only when the RELO pattern is not applicable.
```
```suggestion
:ballot_box_with_check: **YOU SHOULD** only use the stepwise LRO pattern when the RELO pattern is not applicable.
```

azure/LongRunningOperations.md
markweitzel: This doc & content look good. One thing we may want to call out is that this is a REST binding to LROs and that implementations for other protocols should follow the same pattern. For example, in resource based pattern, you get a "shallow" representation of the resource. In stepwise, you get an alternate location to go ask for the state of the resource. 

Do we want to add anything about the identity of the resource in the two different models? For example, in resource oriented LROs, you can get back the shallow representation of the resource, which should have its ID. This would not be the case in stepwise.

What about idempotency? Should we explicitly call out that these are idempotent operations?

markweitzel (approved):
Nice addition to the updated guidelines--Thanks!
LGTM

azure/Guidelines.md
```diff
@@ -131,7 +131,7 @@ GET    | Read (i.e. list) a resource collection | `200-OK`
 GET    | Read the resource | `200-OK`
 DELETE | Remove the resource | `204-No Content`\; avoid `404-Not Found`
 
-:white_check_mark: **DO** return status code `202-Accepted` and follow the guidance in [Long Running Operations & Jobs](#long-running-operations--jobs) when a PUT, POST, or DELETE method completes asynchronously
+:white_check_mark: **DO** return status code `202-Accepted` and follow the guidance in [Long-Running Operations & Jobs](#long-running-operations--jobs) when a PUT, POST, or DELETE method completes asynchronously
```
JeffreyRichter: PATCH too, right?

azure/Guidelines.md
```diff
@@ -747,18 +747,70 @@ The ability to retry failed requests for which a client never received a respons
...
 [OpenAPI Specification]: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md
 
-### Long Running Operations & Jobs
+### Long-Running Operations & Jobs
 
-Azure generally follows the [Microsoft REST API guidelines for Long running operations](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#13-long-running-operations).  Follow the Microsoft guidelines with the modifications / extensions given here when designing / implementing long running operations.
+When the processing for an operation may take a significant amount of time to complete, it should be
+implemented as a _long-running operation (LRO)_. This allows clients to continue running while the
+operation is being processed. The client obtains the outcome of the operation at some later time
+through another API call.
```
JeffreyRichter, line 782: We should be clearer here with "significant amount of time to complete". I just worked with a team where the processing could take 2 weeks and we don't want to model this as an LRO.  I think the processing should be on the order of a few minutes at most; even 5 or 10 minutes seems too long to me to have an app poll.

azure/Guidelines.md
```diff
@@ -747,18 +747,70 @@ The ability to retry failed requests for which a client never received a respons
...
 
-Previous Azure guidelines specified "Azure-AsyncOperation" as the name of the response header containing the operation URL,
-while the Microsoft guidelines use the name "Operation-Location".
+:white_check_mark: **DO** implement an operation as an LRO if the 99th percentile response time is greater than 1s.
```
JeffreyRichter, line 786: This was from the old guidelines so I know where this came from. But, see my comment above.

azure/Guidelines.md
JeffreyRichter: Also, how does this work if 2 LROs are issued to the same resource at the same time? I don't think this guideline is sufficient.

azure/Guidelines.md
```diff
@@ -747,18 +747,70 @@ The ability to retry failed requests for which a client never received a respons
...
 
-:white_check_mark: **DO** support both `Azure-AsyncOperation` and `Operation-Location` HEADERS, even though they are redundant so that existing SDKs and clients will continue to operate.
+There are two basic patterns that can be used for long-running operations:
+1. Resource-based long-running operations (RELO)
+2. Stepwise long-running operations
+
+:white_check_mark: **DO** use the RELO pattern when the operation is on a resource that contains a "status" property that can be used to obtain the outcome of the operation.
+
+:ballot_box_with_check: **YOU SHOULD** only use the stepwise LRO pattern when the RELO pattern is not applicable.
```
JeffreyRichter: I think it would be good to give an example of when it is not applicable.

azure/Guidelines.md
```diff
@@ -747,18 +747,70 @@ The ability to retry failed requests for which a client never received a respons
...
+
+#### Resource-based long-running operations
+
+Some common situations where the RELO pattern should be used:
+1. A "create" operation for a resource where the basic structure of the resource is created immediately and includes a status field that indicates when the create has completed, e.g. "provisioning" -> "active".
```
JeffreyRichter: Create means POST, PUT or PATCH - I think these should be mentioned.
If PUT or PATCH (not POST), then how does this work for retries/idempotency?

azure/Guidelines.md
```diff
@@ -747,18 +747,70 @@ The ability to retry failed requests for which a client never received a respons
...
+2. An action operation for a resource where both the initiation of the action and the completion of the action cause a change to the "status" property of the resource.
```
JeffreyRichter, line 803: Again, what if multiple actions are applied to the resource at the same time?
How would etags fit into this story?

azure/Guidelines.md
```diff
@@ -747,18 +747,70 @@ The ability to retry failed requests for which a client never received a respons
...
+
+:white_check_mark: **DO** return a `200-OK` response, `201-Created` for create operations, from the request that initiates the operation.  The response body should contain a representation of the resource that clearly indicates that the operation has been accepted or started.
```
JeffreyRichter, line 805: I'm confused here. Does RELO mean the operation is acting on a "normal resource" or that the act of creating an operation is creating an "operation resource" which is being created purely to manage an ongoing operation?
Perhaps we need to break RELO into 2 different patterns: a pattern that performs an operation on a "normal resource" and a pattern for monitoring an operation using an operation-specific resrouce?

azure/Guidelines.md
```diff
@@ -747,18 +747,70 @@ The ability to retry failed requests for which a client never received a respons
...
+
+:white_check_mark: **DO** support a get method on the resource that returns a representation of the resource including the status field that indicates when the operation has completed.
```
JeffreyRichter, line 807: This feels like it's for the operation-specific resource.
It could be for the normal resource but then there is no way to execute multiple operations on the 1 resource at the same time.

azure/Guidelines.md
JeffreyRichter: In my LRO slides (which I never published), I thought of LROs as these 2 catagories:
- 1:1 - Performing an action that mutates a resource in-place (but this has the 2 at once issue)
- 1:n - Performing an action that does not mutate a resource

azure/Guidelines.md
```diff
@@ -747,18 +747,70 @@ The ability to retry failed requests for which a client never received a respons
...
+
+:white_check_mark: **DO** define the "status" field of the resource as an enum with all the values it may contain including the "terminal" values "Succeeded", "Failed", and "Canceled".
+
+:ballot_box_with_check: **YOU SHOULD** use the name `status` for the "status" field of the resource.
+
+#### Stepwise long-running operations
+
+In a stepwise long-running operation, the client makes a request to initiate the operation processing and receives a URL in the response where it can obtain the operation results. The [HTTP specification](https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.3) calls the target of this URL a "status monitor" but we will refer to it as the operation-results.
```
JeffreyRichter: We should say how this happens; for example, is this a POST? If so, how is idempotency achieved? If the client gets the URL in the response (and we should say which header this should be in) and the client crashes, how can a client instance catch of with this operation?

azure/Guidelines.md
```diff
@@ -747,18 +747,70 @@ The ability to retry failed requests for which a client never received a respons
...
+
+:white_check_mark: **DO** return a `202-Accepted` status code from the request that initiates a stepwise LRO if the processing of the operation was successfully initiated.
+
+:white_check_mark: **DO** perform as much validation of the initial request as practical and return an error response when appropriate.
```
```suggestion
:white_check_mark: **DO** perform as much validation of the initial request as practical and return an error response immediately when appropriate (without starting the operation).
```

azure/Guidelines.md
```diff
@@ -747,18 +747,70 @@ The ability to retry failed requests for which a client never received a respons
...
+
+:white_check_mark: **DO** return the operation-results URL in the `Operation-Location` response header.
```
JeffreyRichter: I'd merge this bullet with the above bullet

azure/Guidelines.md
```diff
@@ -747,18 +747,70 @@ The ability to retry failed requests for which a client never received a respons
...
+
+:white_check_mark: **DO** support the `get` method on the operation-results endpoint that returns a `200-OK` response with a response body that indicates whether the operation has completed and completion status (success, failed) with sufficient error information to diagnose failures.
```
```suggestion
:white_check_mark: **DO** support the `get` method on the operation-results endpoint that returns a `200-OK` response with a response body that indicates whether the operation has completed and completion status (success, failed, canceled) with sufficient information to diagnose any potential failures.
```

azure/Guidelines.md
```diff
@@ -747,18 +747,70 @@ The ability to retry failed requests for which a client never received a respons
...
+
+:white_check_mark: **DO** include a field in the operation-results named `status` to contain the operation status. This field should be a string with well defined values that must include "Succeeded", "Failed", and "Canceled".
```
```suggestion
:white_check_mark: **DO** include a field in the operation-results resource named `status` indicating the operation's status. This field should be a string with well-defined values. Indicate the terminal state using "Succeeded", "Failed", or "Canceled".
```

azure/Guidelines.md
```diff
@@ -747,18 +747,70 @@ The ability to retry failed requests for which a client never received a respons
...
+
+:white_check_mark: **DO** include a field in the operation-results named `error` to contain error information -- minimally `code` and `message` fields -- when an operation fails.
+
+:white_check_mark: **DO** support return of the operation-results for some documented period of time, a minimum of 24 hours, after the operation has completed.
```
```suggestion
:white_check_mark: **DO** retain the operation-results resource for some documented period of time (at least 24 hours) after the operation completes. 
```

azure/Guidelines.md
```diff
@@ -747,18 +747,70 @@ The ability to retry failed requests for which a client never received a respons
...
+
+:white_check_mark: **DO** include a `Retry-After` header in the response to the initiating request and requests to the operation-results endpoint. The value of this header should be an integer number of seconds to wait before making the next request to the operations-results URL.
```
```suggestion
:white_check_mark: **DO** include a `Retry-After` header in the response to the initiating request and requests to the operation-location URL. The value of this header should be an integer number of seconds to wait before making the next request to the operation-location URL.
```

azure/Guidelines.md
```diff
@@ -747,18 +747,70 @@ The ability to retry failed requests for which a client never received a respons
...
+
+:heavy_check_mark: **YOU MAY** support a `get` method on the operation-results collection endpoint that returns a list of all recently initiated operations.
```
```suggestion
:heavy_check_mark: **YOU MAY** support a `get` method on the operation-results collection URL that returns a list of all recently initiated operations.
```

JeffreyRichter: "Recently initiated" is ambiguous. Does them in-flight?
I think we want to say a GET on the collection by default returns in-flight operations (which is probably a small set so a good default). And then, maybe we have some query parameters to further include completed, error, canceled, and maybe a timestamp range?

azure/Guidelines.md
```diff
@@ -747,18 +747,70 @@ The ability to retry failed requests for which a client never received a respons
...
+
+:no_entry: **DO NOT** return any other `2xx` status code from the initial request of a stepwise LRO -- return `202-Accepted` and an operations-result URL even if processing was completed before the initiating request returns.
```
JeffreyRichter: The terms operation-Location & operation-result are not used consistently. 
I think this whole thing is creaming for an example -- I do have some in my slide deck (even though I never recorded videos).

azure/Guidelines.md
```diff
@@ -747,18 +747,70 @@ The ability to retry failed requests for which a client never received a respons
...
+:no_entry: **DO NOT** return any data in the response body of a `202-Accepted` response.
+
+Previous Azure guidelines specified "Azure-AsyncOperation" as the name of the response header containing the operation-results URL.
+
+:white_check_mark: **DO** return **both** `Azure-AsyncOperation` and `Operation-Location` headers if your service previously returned `Azure-AsyncOperation`, even though they are redundant, so that existing clients will continue to operate.
 
 :white_check_mark: **DO** return the same value for **both** headers.
 
-:white_check_mark: **DO** look for **both** HEADERS in client code, preferring the `Operation-Location` version.
+:white_check_mark: **DO** look for **both** headers in client code, preferring the `Operation-Location` header.
```
JeffreyRichter, line 845: This is client guideline; not a service guideline. If you want to say something about this, then I'd merge it with the previous bullet and I'd give better guidance. Specifically, clients should look for 'operation-location' first and if not present, then look for 'azure-asyncoperation'.

azure/LongRunningOperations.md
```diff
@@ -0,0 +1,69 @@
...
+to a value indicating that the operation processing has been started.
+
+3. The client then issues a GET request to the resource to determine if the operation processing
+has completed.
+
+4. The resource responds with a representation of the resource. While the operation is still being
+processed, the status field will contain a "non-terminal" value, like "processing".
+
+5. After the operation processing has completed, a GET request from the client will receive a response
+where the status field contains a "terminal" value -- `succeeded`, `failed`, or `cancelled` --
```
JeffreyRichter: Inconsistent casing has been used "Failed" vs "failed".
Also, i think it's "Success"; not "succeeded".
Also, I think cancelled with 1 L and 2 Ls has also been used.
All this needs to be cleaned up.

azure/LongRunningOperations.md
```diff
@@ -0,0 +1,69 @@
...
+
+![The stepwise LRO flow](./stepwise.jpg)
+
+1. The client sends the initial request to initiate the long-running operation.
+As in the RELO pattern, the initial request could be a PUT, PATCH, POST, or DELETE method.
+
+2. The resource validates the request and initiates the operation processing.
+It sends a response to client with a `202-Accepted` HTTP status code.
+Included in this response is an `Operation-location` response header with the absolute URL of an
+operation-results endpoint for this specific operation.
```
JeffreyRichter: To me, this question comes up: Why does the client need the operation-location URL if the operation-result URL is well-known? I think we're over complicating.
Now, you might say because a POST to initiate the operation has to return the operation ID so it can be used to look up the operation in the operation-results collection but then why don't we say that operation should just be initated via a PUT to the operation-results collection and then the ID is passed by the client and so there is no need for the operation-location header at all. And, this makes starting the operation idempotent too.

azure/LongRunningOperations.md
```diff
@@ -0,0 +1,69 @@
...
+The response also includes a `Retry-after` header telling the client a minimum time to wait (in seconds)
+before sending a request to the operation-results URI.
```
JeffreyRichter: URI or URL - please be consistent

azure/LongRunningOperations.md
```diff
@@ -0,0 +1,69 @@
...
+
+3. After waiting at least the amount of time specified by the previous response's `Retry-after` header,
+the client issues a GET request to the operation-results URI.
+
+4. The operation-results URI responds with an operation-results object, containing information about the
```
JeffreyRichter:
```suggestion
4. The operation-results URL responds with an operation-results resource containing information about the
```

azure/LongRunningOperations.md
```diff
@@ -0,0 +1,69 @@
...
+operation including its current status. If the operation is still being processed,
+the status field will contain a "non-terminal" value, like "processing".
+
+5. After the operation processing has completed, a GET request to operation-results URI will return an
```
JeffreyRichter:
```suggestion
5. After the operation processing completes, a GET request to operation-results URL returns an
```

azure/LongRunningOperations.md
```diff
@@ -0,0 +1,69 @@
...
+operation-results object with a status field containing a "terminal" value -- `succeeded`, `failed`, or `cancelled` --
```
JeffreyRichter:
```suggestion
operation-results resource with its status field containing a terminal value: `succeeded`, `failed`, or `cancelled` --
```

JeffreyRichter: Fix values for consistency

azure/LongRunningOperations.md
```diff
@@ -0,0 +1,69 @@
...
+that indicates the result of the operation.
+If the status is `failed`, the operation-results object will contain an `error` field with a `code` and `message`
```
JeffreyRichter:
```suggestion
If the status is `failed`, the operation-results resource must contain an `error` field with a `code` and `message`
```

JeffreyRichter: We really need an example here

azure/Guidelines.md
johanste: Agreed, but we do need to give additional guidance on how to correctly model a process that takes 15+min to complete. 

In general, a long long running operation should be modeled as a discoverable (listable) resource in its own right - e.g. a "build" resource in a CI system can be "rediscovered" by listing pending builds rather than just being modeled as a generic operations resource.

azure/Guidelines.md
johanste: Generally, you only want to wait until the resource is in a stable state when you use RELO. You don't care which operation kicked it out of the stable state. So multiple LROs to the same resource tend to - in practice - work just fine. At least I have never gotten actual customer feedback on that pattern.

azure/Guidelines.md
johanste: If the requested target state of the created resource is the same (e.g. what will be the case for a retry), then the service can respond with a 200 and the current state of the resource. 

It would be up to the service to determine if it allows for partially created resources to be overwritten with conflicting settings or not. Is that the guidance you are looking for?

azure/Guidelines.md
mkistler: I agree that this pattern is not appropriate if users need to know the completion status of _a specific operation_, as opposed to knowing if the desired effect of their operation has been achieved.  E.g. if a VM is "failed" and a user issues a "reboot" operation on that VM, what then want to know is when the machine is available -- not necessarily that it was _their particular operation_ that caused it to become available.

So I propose adding something like the following:
> The RELO pattern should not be used in cases where the completion status of individual operations may be important to users, as opposed to simply learning that _an_ operation of the type they requested (e.g. reboot of a VM) has successfully completed.

azure/Guidelines.md
mkistler: I'll take a stab at language to cover the "long-long-running operation" case.

azure/Guidelines.md
mkistler: If the multiple actions of the same type were requested -- e.g. reboot a VM -- then the service could simply "piggyback" duplicate actions on the first one.  Callers would see the first action complete and take that to mean that "their" action completed.

In other cases, either the service would have to reject actions that were inconsistent with the current outstanding action or the service should use the "stepwise" LRO pattern so that each user could see the outcome of their specific operation.

azure/Guidelines.md
mkistler: I believe that RELO means there is no operation resource exposed to the user.  The status field of the "normal resource" is used to convey the completion status to the user.

I really do not want multiple RELO patterns if we can avoid it.  Thing are already complicated enough with RELO and stepwise LROs. Let's avoid adding a third type unless there is a real compelling need.

azure/Guidelines.md
mkistler: I think RELO has this restriction -- you can't execute multiple "incompatible" operations on the resource at the same time.

azure/Guidelines.md
mkistler: I think stepwise LRO could be initiated by any of PUT, PATCH, POST, or DELETE.  The guidance below does say that the operation-results URL comes back in the `Operation-Location` header.

azure/Guidelines.md
mkistler: I'm open to that if others think it is better.  I made it separate because it seems more consistent with the "checklist" style of guidelines we are using throughout.

azure/Guidelines.md
mkistler: I agree we probably should be more specific.  But I would go the other way -- GET returns all operations by default -- in-flight, completed (success or error), and cancelled, and then we can offer query parameters to select specific types or timestamp ranges.

azure/Guidelines.md
mkistler: I put examples in the separate document, and we can reference these once we decide where to put them.

azure/LongRunningOperations.md
mkistler: I did some research on "Success" vs "Succeeded":
- The [ARM guidelines](https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/async-api-reference.md#provisioningstate-property) say "Succeeded"
- Looking just at resource-manager API definitions in azure-rest-api-specs:
  - "Succeeded" appears as an enum value (well -- on a line by itself) 1339 times in 694 files
  - "Success" appears as an enum value 78 times in 68 files

I think it is also desirable for enum values to be consistent in their part of speech -- e.g. all verbs or all nouns.  "Success" is a noun.  I believe that "Succeeded", "Failed", and "Canceled" are all intransitive verbs.

So I think we want to use "Succeeded".

I will make the other corrections.

azure/LongRunningOperations.md
mkistler: I think this confusion comes from my choice of naming.  I was calling the URL returned in the `Operation-location` header the "operation-results" URL.  Perhaps calling it the "operation-location" URL is better -- or maybe just "operation URL"?

azure/Guidelines.md
JeffreyRichter: How should a service handle multiple concurrent long-running PATCH operations to a single resource? My guess is that the PATCH operations get queued, so they execute serially. This is good until you throw etags into the picture. If a client does a PATCH with an if-match etag the service should do what if another PATCH is already queued up? The queued PATCH may succeed and then the if-match will fail or the previous PATCH may fail and then the if-match is successful. But, the service returned a 202 accepting the PATCH before knowing the outcome. Or, if the client needs to know if a specific PATCH succeeded or failed, then they should not PATCH the actual resource itself and should issue the PATCH to an operations collection.

Maybe we say that this situation is so rare, we don't care but it makes me feel a bit uncomfortable. If we're ignoring dealing with this, then let's be sure we discussed and decided to ignore it.

azure/Guidelines.md
JeffreyRichter: No, I misspoke in my comment. I meant we should give guidance for POST idempotency; not PUT/PATCH. We should also have guidance about resource visibility: should the resource be visible (but not in fully-constructed state) as soon as the operation is initiated, or should the resource not be visible until complete? I assume the former because this is the only way for a client to detect failure.

azure/Guidelines.md
JeffreyRichter: I agree that this is built into the RELO pattern but what if clients do this anyway as there is no way to prevent it?
An example is a long-running operation that subtracts money from a bank account resource and fails if the balance hits $0.00

azure/Guidelines.md
mikekistler: We should use "operation-status" rather than "operation-results".

azure/Guidelines.md
mikekistler: Other options:
- "job-based" LRO
- "status-monitor" LRO
- "monitored LRO"

mikekistler: "status-monitor" LRO

azure/Guidelines.md
```diff
@@ -747,18 +747,77 @@ The ability to retry failed requests for which a client never received a respons
...
+
+:white_check_mark: **DO** include a field in the operation-results named `error` to contain error information -- minimally `code` and `message` fields -- when an operation fails.
+
+:white_check_mark: **DO** retain the operation-results resource for some documented period of time (at least 24 hours) after the operation completes. 
+
+:white_check_mark: **DO** include a `Retry-After` header in the response to the initiating request and requests to the operation-location URL. The value of this header should be an integer number of seconds to wait before making the next request to the operation-location URL.
+
+:heavy_check_mark: **YOU MAY** support a `get` method on the operation-results collection URL that returns a list of all recently initiated operations.
+
+:no_entry: **DO NOT** return any other `2xx` status code from the initial request of a stepwise LRO -- return `202-Accepted` and an operations-result URL even if processing was completed before the initiating request returns.
```
mikekistler: We need to consider how "short LROs" should work here. Maybe just soften this one notch to discourage teams from doing this.

azure/Guidelines.md
mikekistler: Remove "white_check_mark".

rysweet (approved):

azure/LongRunningOperations.md
mkistler: Fixed.  There is now just one occurrence of "URI" and that is within the name of `414` status code, which is the official name so I kept it.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -110,13 +112,107 @@ Another important design pattern for avoiding surprises is idempotency. An opera
...
 
 :ballot_box_with_check: **YOU SHOULD** use extensible enumerations. Extensible enumerations are modeled as strings - expanding an extensible enumeration is not a breaking change.
 
 :ballot_box_with_check: **YOU SHOULD** implement [conditional requests](https://tools.ietf.org/html/rfc7232) early. This allows you to support concurrency, which tends to be a concern later on.
 
+## Long-Running Operations
+
+Long-running operations are an API design pattern that should be used when the processing of
+an operation may take a significant amount of time -- longer than a client will want to block
+waiting for the result.
```
tg-msft, line 126: Is it worth adding a blurb about "worst case performance" in particular?  Even if your API completes quickly 99% of the time, it still might need to be an LRO and that's something teams should consider up front.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -110,13 +112,107 @@ Another important design pattern for avoiding surprises is idempotency. An opera
...
+operation request.  Here's what the status monitor LRO pattern looks like:
+
+<!-- markdownlint-disable MD033 -->
+<p align="center">
+  <img src="./statmon.jpg" alt="The status monitor LRO flow"/>
+</p>
+<!-- markdownlint-enable MD033 -->
+
+1. The client sends the request to initiate the long-running operation.
+As in the RELO pattern, the initial request could be a PUT, PATCH, POST, or DELETE method.
```
tg-msft, line 191: Do we want to mention idempotency here and suggest the client send the operation's ID?

azure/Guidelines.md
```diff
@@ -773,18 +774,75 @@ The ability to retry failed requests for which a client never received a respons
...
+
+Some common situations where the RELO pattern should be used:
+1. A "create" operation (PUT, PATCH, or POST) for a resource where the basic structure of the resource is created immediately and includes a status field that indicates when the create has completed, e.g. "provisioning" -> "active".
+2. An action operation for a resource where both the initiation of the action and the completion of the action cause a change to the "status" property of the resource.
+
+:white_check_mark: **DO** return a `200-OK` response, `201-Created` for create operations, from the request that initiates the operation.  The response body should contain a representation of the resource that clearly indicates that the operation has been accepted or started.
+
+:white_check_mark: **DO** support a get method on the resource that returns a representation of the resource including the status field that indicates when the operation has completed.
+
+:white_check_mark: **DO** define the "status" field of the resource as an enum with all the values it may contain including the "terminal" values "Succeeded", "Failed", and "Canceled".
```
tg-msft, line 809: Should we suggest an extensible enum for this so it versions better?  The status monitor section below calls out a `string` in particular for this.

tg-msft (approved):

PR #296, opened 2021-10-03
Reference RFC 1123 for header date format

mikekistler:
This PR reverts a change I had made to use RFC 7231 rather than RFC 1123 as the reference for the date format to used for date values passed in headers.

I had switched the reference to RFC 7231 because I couldn't find the part of RFC 1123 that described date formats of the kind we use in headers. That's because RFC 1123 describes the date format in an obscure fashion -- as a modification to the date formation of RFC 822.  So not that this is clear I'm reverting the change (and adding an explanation of the RFC 1123 date format).


markweitzel (approved):
LGTM - Thank you for reverting this back.

JeffreyRichter (approved):

PR #301, opened 2021-10-31
A passel of minor fixes

### LINKED ISSUE #290: DELETE entries in Optimistic Concurrency table should probably be If-Match
In the table in the Optimistic Concurrency section (https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#optimistic-concurrency), should the header for the DELETE entries be If-Match (not If-None-Match)?

If-Match is commonly used on state changing methods, like DELETE.

For DELETE, If-Match: "xyz" should return 412 if the value on the server does not match the value provided.
For DELETE, If-Match: "xyz" should return 204 if the value on the server matches the value provided.
For DELETE, If-Match: * should return 204 regardless of the value on the server.

### LINKED ISSUE #298: Broken links in Azure API Guidelines
There are a few broken links in the [table of headers](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#http-query-parameters-and-header-values) in the Azure API Guidelines:

- the "see Customer Support" link is just a "TODO"

- the "Distributed Tracing & Telemetry" links do not work.

### LINKED ISSUE #299: Clarify format of date-base version
The documentation (see [here](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#api-versioning)) isn't clear on the format. From the example, I can infer the first 4 digits are the year, but it's not clear whether it is YYYY-MM-DD or YYYY-DD-MM.

mikekistler:
This PR fixes number of small problems such as typos and broken links.

Fixes #290, fixes #298, fixes #299.

### DIFF
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -64,7 +64,7 @@ Where:
  | - | - |
  | service | Name of the service (ex: blobstore, servicebus, directory, or management)
  | cloud | Cloud domain name, e.g. `azure.net` (see Azure CLI's "az cloud list")
- | tenant | Globally-unique ID of container representing tenant isolation, billing, enforced quotas, lifetime of containees (ex: subscription UUID)
+ | tenant | Globally-unique ID of container representing tenant isolation, billing, enforced quotas, lifetime of containers (ex: subscription UUID)
  | service&#x2011;root | Service-specific path (ex: blobcontainer, myqueue)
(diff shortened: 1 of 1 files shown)


JeffreyRichter (approved):

markweitzel (approved):
LGTM - Thanks for the clean up.

PR #302, opened 2021-11-01
change notifications and change tracking patterns

baywet:
- adds pattern description for change notifications

I've only added the pattern for change tracking so far. Especially considering we already have webhook and delta in the same directory.
Before I move on to change notifications, I wanted to know whether the format, length and tone was what's expected here?
Thanks for the feedback!


graph/change-tracking.md
```diff
@@ -0,0 +1,46 @@
...
+
+Microsoft Graph API Design Pattern
+
+Change tracking enables users to keep a third party system in sync with changes in Microsoft Graph without having to continuously poll the API.
+
+Context
+-------
+
+In the world of distributed systems and richness of devices, keeping data in sync across systems is paramount to delivering a great user experience.
+
```
OlgaPodo: In general i like it!  But i think we need to accommodate developers request to make it short and actionable thus  may be provide a concrete example like users or messages to describe the context then demonstrate a problem for API provider like how to enable frequent client querying without significant performance cost. Then solution would be like the workload must support a watermarks, API needs a function, annotations, etc. I created one pattern  as i see it, still need to do peer review though

graph/change-tracking.md
```diff
@@ -0,0 +1,46 @@
...
+--------
+
+Change Tracking (aka delta) provides developers with a new endpoint they can use to sync their third party system, get a delta link with a watermark, go offline, and comeback with that delta link to get new changes since their last request. Delta guarantees integrity of data through the watermark, regardless of service partitions and other obscure aspects for clients.
+
+* *
+
+Issues and Considerations
+-------------------------
+
+Implementer MUST implement a watermark storage system in case of active watermarks.
```
OlgaPodo: like the placement of the prerequisite

graph/change-tracking.md
```diff
@@ -0,0 +1,46 @@
...
+- Partitions management and requests affinity -> two external requests can fall on different replicas that don't have the same sync state.
+- Entities updates tracking -> the way entities are sync internally and tracking mechanisms should not "leak" externally. (update id, update date, conflicts resolution...)
+- Multiple sync destinations -> having multiple systems syncing to the same dataset should not impact behavior/performance of the system.
+
+* *
+
+Solution
+--------
+
+Change Tracking (aka delta) provides developers with a new endpoint they can use to sync their third party system, get a delta link with a watermark, go offline, and comeback with that delta link to get new changes since their last request. Delta guarantees integrity of data through the watermark, regardless of service partitions and other obscure aspects for clients.
```
OlgaPodo: need all the main components of the solution like annotations, function, etc

baywet: I didn't include any of that because it's already included in delta.md. What is the plan for the existing pages?

graph/change-tracking.md
baywet: so nothing required on my end for this comment?

graph/change-tracking.md
baywet: again most of the technical implementation aspects are already in delta.md.
As for the "architecture" pattern, the goal is to remove the need for clients to query often by providing them with an efficient to track changes.

graph/change-tracking.md
OlgaPodo: I think that we will replace existing mds with new ones since developers asked to remove philosophy and add examples.

graph/change-tracking.md
OlgaPodo: it's good

graph/change-tracking.md
baywet: why are we starting from scratch and not just updating the existing then?

OlgaPodo: The idea, coming from API Council, was to make pattern articles short and sweet without loosing the main context

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,94 @@
...
+---------
+
+Third party systems require to keep data in sync with Microsoft Graph, and the API design should not allow for continuous polling as it'd be costly and wouldn't guarantee data integrity.
+
+## Solution
+--------
+
+API designers leverage the change tracking (delta) capability on the entity set and declare a delta function for API consumers to use when tracking changes happening in the system.
+
+This new endpoint can be uses to sync third party systems. This is achieved through returning a delta link with a watermark. Once the third party system comes back online, it uses the last provided delta link to catch up on new changes since their last request. Delta guarantees integrity of data through the watermark, regardless of service partitions and other obscure aspects for clients.
```
PetRich-MSFT:
```suggestion
This new endpoint can be used to sync API consumers This is achieved through returning a delta link with a watermark. When the API consumer needs to refresh the data, it uses the last provided delta link to catch up on new changes since their last request. Delta guarantees integrity of data through the watermark, regardless of service partitions and other obscure aspects for clients.
```

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Issues and Considerations
+-------------------------
+
+Implementer MUST implement a watermark storage system in case of active watermarks (cursor in data store, partition affinity, sync state in data store...).
```
PetRich-MSFT: Do we need to define what an ***active*** watermark is?

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Problem
+---------
+
+Third party systems require to keep data in sync with Microsoft Graph, and the API design should not allow for continuous polling as it'd be costly and wouldn't guarantee data integrity.
+
+## Solution
+--------
+
+API designers leverage the change tracking (delta) capability on the entity set and declare a delta function for API consumers to use when tracking changes happening in the system.
```
PetRich-MSFT:
```suggestion
API designers leverage the change tracking (delta) capability on the entity set and declares a delta function for API consumers to use when tracking changes happening in the system.
```

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,94 @@
+# Change Tracking
+
+Microsoft Graph API Design Pattern
+
+*The change tracking pattern provides the ability to keep a third party system in sync with changes in Microsoft Graph without having to continuously poll the API.*
+
+## Problem
+---------
+
+Third party systems require to keep data in sync with Microsoft Graph, and the API design should not allow for continuous polling as it'd be costly and wouldn't guarantee data integrity.
```
PetRich-MSFT:
```suggestion
API consumers require an efficient way to keep data in sync with Microsoft Graph, and the API design should not allow for continuous polling as it'd be costly and wouldn't guarantee data integrity.
```

graph/patterns/change-tracking.md
darrelmiller:
```suggestion
API designers can enable the change tracking (delta) capability on entity collections by declaring a delta function for API consumers to use to track changes in that collection.
```

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,94 @@
+# Change Tracking
+
+Microsoft Graph API Design Pattern
+
+*The change tracking pattern provides the ability to keep a third party system in sync with changes in Microsoft Graph without having to continuously poll the API.*
```
PetRich-MSFT:
```suggestion
*The change tracking pattern provides the ability to keep API consumers in sync with changes in Microsoft Graph without having to continuously poll the API.*
```

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,94 @@
...
+            "userPrincipalName": "GradyA@contoso.onmicrosoft.com",
+            "id": "0baaae0f-b0b3-4645-867d-742d8fb669a2",
+            "manager@delta": [
+                {
+                    "@odata.type": "#microsoft.graph.user",
+                    "id": "99789584-a1e1-4232-90e5-866170e3d4e7"
+                }
+            ]
+        },
+    ]
```
PetRich-MSFT, line 93: We should have an example that covers something being added, something being updated, and something being removed

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,94 @@
...
+Third party systems require to keep data in sync with Microsoft Graph, and the API design should not allow for continuous polling as it'd be costly and wouldn't guarantee data integrity.
+
+## Solution
+--------
+
+API designers leverage the change tracking (delta) capability on the entity set and declare a delta function for API consumers to use when tracking changes happening in the system.
+
+This new endpoint can be used to sync third party systems. This is achieved through returning a delta link with a watermark. Once the third party system comes back online, it uses the last provided delta link to catch up on new changes since their last request. Delta guarantees integrity of data through the watermark, regardless of service partitions and other obscure aspects for clients.
+
+## Issues and Considerations
```
PetRich-MSFT, line 21: Should we mention that delta can be a composable function, therefore you can pass $filter etc.

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,94 @@
...
+---------
+
+API consumers require an efficient way to keep data in sync with Microsoft Graph, and the API design should not allow for continuous polling as it'd be costly and wouldn't guarantee data integrity.
+
+## Solution
+--------
+
+API designers can enable the change tracking (delta) capability on entity collections by declaring a delta function for API consumers to use to track changes in that collection.
+
+This new endpoint can be used to sync third party systems. This is achieved through returning a delta link with a watermark. Once the third party system comes back online, it uses the last provided delta link to catch up on new changes since their last request. Delta guarantees integrity of data through the watermark, regardless of service partitions and other obscure aspects for clients.
```
PetRich-MSFT:
```suggestion
This new endpoint can be used to sync API consumers. This is achieved through returning a delta link with a watermark. Once the API consumer needs to refresh the data it uses the last provided delta link to catch up on new changes since their last request. Delta guarantees integrity of data through the watermark, regardless of service partitions and other obscure aspects for clients.
```

graph/patterns/change-tracking.md
baywet:
```suggestion
Implementer MUST implement a watermark storage system in case of active watermarks. Passive watermarks are watermarks that can be retrieved from the context (e.g. timestamp), active watermarks represent information required to track the sync state which cannot be retrieved from the context (e.g. cursor from data store, partition affinity marker, partition id, generated unique sync identifier...)
```

What do you think of that?

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,94 @@
...
+
+Implementer MUST implement a watermark storage system in case of active watermarks (cursor in data store, partition affinity, sync state in data store...).
+
+## When to Use this Pattern
+------------------------
+
+Before using the change tracking pattern in your API definition, make sure your scenario fits the following criteria:
+
+- API consumers want to sync the data.
+- API consumers don't want to get notified of changes. (see change notifications pattern for this scenario).
```
PetRich-MSFT:
```suggestion
- API consumers don't want to be immediately notified of changes. (see change notifications pattern for this scenario).
```

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,94 @@
...
+                    "@odata.type": "#microsoft.graph.user",
+                    "id": "99789584-a1e1-4232-90e5-866170e3d4e7"
+                }
+            ]
+        },
+    ]
+}
+```
+
+> Note: the response contains an `@odata.nextLink` instance annotation with the watermark only when all the changes are enumerated. If more changes need to be enumerated, the response instead contains an `@odata.skipLink` instance annotation the application can request right away to get the next page.
```
PetRich-MSFT: Is ```@odata.nextLink``` right here, shouldn't it be ```@odata.deltaLink```

graph/patterns/change-tracking.md
baywet: delta is not composable in the sense that it doesn't accept any path segment after (navigation properties, other functions, ...). But it does accept filter and select statements (I think that technically top and expand could be supported as well).

What do you think of that.

> Note: the delta function also provides support for $filter and $select to allow the API consumer to narrow down the number of entities and properties retrieved as well as the number of changes that are tracked. Additionally the delta function can also support $top to allow the API consumer to sync smaller sets of changes as well as $expand to allow the API consumer to sync related data. Expand across workloads is not supported today however.

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,94 @@
...
+
+API designers can enable the change tracking (delta) capability on entity collections by declaring a delta function for API consumers to use to track changes in that collection.
+
+This new endpoint can be used to sync third party systems. This is achieved through returning a delta link with a watermark. Once the third party system comes back online, it uses the last provided delta link to catch up on new changes since their last request. Delta guarantees integrity of data through the watermark, regardless of service partitions and other obscure aspects for clients.
+
+## Issues and Considerations
+-------------------------
+
+Implementer MUST implement a watermark storage system in case of active watermarks (cursor in data store, partition affinity, sync state in data store...).
+
```
PetRich-MSFT, line 38: Should also call out that the implementor should have a way to track when entities are deleted e.g. soft-delete patterns

graph/patterns/change-tracking.md
baywet:
```suggestion
> Note: the response contains an `@odata.deltaLink` instance annotation with the watermark only when all the changes are enumerated. If more changes need to be enumerated, the response instead contains an `@odata.nextLink` instance annotation the application can request right away to get the next page.
```

graph/patterns/change-tracking.md
baywet: do you mean that so they are able to enumerate deletions when clients issue a request to the delta endpoint? usually the deleted entries only return their ids

graph/patterns/change-tracking.md
baywet: added a removed example.
Added and updated are the same with the instructions to the client app to uppsert.

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,94 @@
+# Change Tracking
```
gareth-a-jones, line 1: I'd prefer we drop the term change tracking and call this delta from the get-go.  change tracking seems like an unnecessary odata-ism to use for the broad feature area.

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,94 @@
...
+
+Microsoft Graph API Design Pattern
+
+*The change tracking pattern provides the ability to keep API consumers in sync with changes in Microsoft Graph without having to continuously poll the API.*
+
+## Problem
+---------
+
+API consumers require an efficient way to keep data in sync with Microsoft Graph, and the API design should not allow for continuous polling as it'd be costly and wouldn't guarantee data integrity.
```
gareth-a-jones: I think this needs a rewrite, because many customers do poll against the delta api pretty frequently.
It's not idea, but we don't have that many webhooks to make it unnecessary.

SO some language more like 'api design should be optimized to avoid polling as it is costly for consumers and producers alike.'

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Solution
+--------
+
+API designers can enable the change tracking (delta) capability on entity collections by declaring a delta function for API consumers to use to track changes in that collection.
+
+This new endpoint can be used to sync API consumers. This is achieved through returning a delta link with a watermark. Once the API consumer needs to refresh the data it uses the last provided delta link to catch up on new changes since their last request. Delta guarantees integrity of data through the watermark, regardless of service partitions and other obscure aspects for clients.
```
gareth-a-jones, line 17: Some explicit explanation needed around us choosing to standardize on use of a delta function rather than the odata standard to access the delta.  Doesn't need to justify why, rather explain that it is the case.

Also shoudl decide whether it is OK to support the odata standard in addition.

graph/patterns/change-tracking.md
baywet:
```suggestion
API consumers require an efficient way to keep data in sync with Microsoft Graph, and the API design should be optimized to avoid polling as it is costly for consumers and producers alike as well as wouldn't guarantee data integrity.
```

what do you think of this?

graph/patterns/change-tracking.md
baywet: we call it change tracking in our public docs and training, I thought aligning on that was good in case they wanted to do more research and get more information.

graph/patterns/change-tracking.md
baywet:
```suggestion
This new endpoint can be used to sync API consumers. This is achieved through returning a delta link with a watermark. Once the API consumer needs to refresh the data it uses the last provided delta link to catch up on new changes since their last request. Delta guarantees integrity of data through the watermark, regardless of service partitions and other obscure aspects for clients.

> Note: although this capability is similar to the [OData $delta feed](https://docs.oasis-open.org/odata/odata-json-format/v4.0/errata02/os/odata-json-format-v4.0-errata02-os-complete.html#_Toc403940644) capability, it is a different construct. Microsoft Graph APIs MUST provide change tracking through the delta function and MUST NOT implement the OData $delta feed when providing change tracking capabilities to ensure the uniformity of the API experience.
```

What do you think of that?

graph/patterns/change-tracking.md
OlgaPodo: I also think that Change Tracking is more generic and should stay  for consistency.

baywet: great, resolving

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Issues and Considerations
+-------------------------
+
+Implementer MUST implement a watermark storage system in case of active watermarks (cursor in data store, partition affinity, sync state in data store...).
+
+## When to Use this Pattern
```
baywet, line 39:
```suggestion
> Note: the delta function also provides support for $filter and $select to allow the API consumer to narrow down the number of entities and properties retrieved as well as the number of changes that are tracked. Additionally the delta function can also support $top to allow the API consumer to sync smaller sets of changes as well as $expand to allow the API consumer to sync related data. Expand across workloads is not supported today however.

## When to Use this Pattern
```

PetRich-MSFT (approved):
Thanks for the updates, looks good to me

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,104 @@
...
+
+> Note: although this capability is similar to the [OData $delta feed](https://docs.oasis-open.org/odata/odata-json-format/v4.0/errata02/os/odata-json-format-v4.0-errata02-os-complete.html#_Toc403940644) capability, it is a different construct. Microsoft Graph APIs MUST provide change tracking through the delta function and MUST NOT implement the OData $delta feed when providing change tracking capabilities to ensure the uniformity of the API experience.
+
+## Issues and Considerations
+-------------------------
+
+Implementer MUST implement a watermark storage system in case of active watermarks. Passive watermarks are watermarks that can be retrieved from the context (e.g. timestamp), active watermarks represent information required to track the sync state which cannot be retrieved from the context (e.g. cursor from data store, partition affinity marker, partition id, generated unique sync identifier...)
```
gareth-a-jones, line 24: You know I'm not sure we shoudl; use MUST language when we're talking about implementation choices.  These guidelines should confine themselves on the whole to what behavior is observable from the outside caller.

If the cursor is fully encoded and internalized in the delta next token then no internal storage is required.

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,104 @@
...
+
+> Note: the delta function also provides support for $filter and $select to allow the API consumer to narrow down the number of entities and properties retrieved as well as the number of changes that are tracked. Additionally the delta function can also support $top to allow the API consumer to sync smaller sets of changes as well as $expand to allow the API consumer to sync related data. Expand across workloads is not supported today however.
+
+## When to Use this Pattern
+------------------------
+
+Before using the change tracking pattern in your API definition, make sure your scenario fits the following criteria:
+
+- API consumers want to sync the data.
+- API consumers don't want to be immediately notified of changes. (see change notifications pattern for this scenario).
```
gareth-a-jones, line 45: This isn't quite correct.   Cos when people are doing change notifications they should ALWAYS implement delta as the premium and reliable way to get the actual change data.

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,104 @@
...
+- API consumers are not looking for a "one-time" export or back-up mechanism.
+
+### Alternatives
+
+- Change notifications pattern (TODO add link when described)
+- Export pattern (TODO)
```
gareth-a-jones: I'm not sure yet that backup/export  will end up as patterns.  Let's add them in if they end up existing.

graph/patterns/change-tracking.md
gareth-a-jones: I think you need to explain the structure of the @removed/reason part of the pattern here.

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Issues and Considerations
+-------------------------
+
+Implementer MUST implement a watermark storage system in case of active watermarks. Passive watermarks are watermarks that can be retrieved from the context (e.g. timestamp), active watermarks represent information required to track the sync state which cannot be retrieved from the context (e.g. cursor from data store, partition affinity marker, partition id, generated unique sync identifier...)
+
+> Note: the delta function also provides support for $filter and $select to allow the API consumer to narrow down the number of entities and properties retrieved as well as the number of changes that are tracked. Additionally the delta function can also support $top to allow the API consumer to sync smaller sets of changes as well as $expand to allow the API consumer to sync related data. Expand across workloads is not supported today however.
+
```
gareth-a-jones, line 38: Should document the  "manager@delta": [] for nav properties part of the pattern here.

graph/patterns/change-tracking.md
baywet: not really, they could either rely on rich notifications when available or just read the full entity and overwrite their copy. The later might be more costly but it's simpler to implement (but I agree it doesn't guarantee data integrity, especially when you factor in partitions affinity and potential sync delays).

graph/patterns/change-tracking.md
baywet:
```suggestion
```

graph/patterns/change-tracking.md
baywet: yeah that's the point, because the cursor is already stored, no need to store it twice and the workload would satisfy the requirement.

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,104 @@
...
+> Note: although this capability is similar to the [OData $delta feed](https://docs.oasis-open.org/odata/odata-json-format/v4.0/errata02/os/odata-json-format-v4.0-errata02-os-complete.html#_Toc403940644) capability, it is a different construct. Microsoft Graph APIs MUST provide change tracking through the delta function and MUST NOT implement the OData $delta feed when providing change tracking capabilities to ensure the uniformity of the API experience.
+
+## Issues and Considerations
+-------------------------
+
+Implementer MUST implement a watermark storage system in case of active watermarks. Passive watermarks are watermarks that can be retrieved from the context (e.g. timestamp), active watermarks represent information required to track the sync state which cannot be retrieved from the context (e.g. cursor from data store, partition affinity marker, partition id, generated unique sync identifier...)
+
```
baywet, line 25:
```suggestion

Implementer MUST implement soft deletion for entities in the backend storage system. The soft deletion will provide useful information to the client to appropriately reflect deletions.

When an entity is soft deleted, the delta function MUST return the id of the deleted entity as well as a `@removed` annotation with the `reason` field.
- The reason MUST be set to `changed` if the entity can be restored. `"@removed": {"reason": "changed"}`. 
- The reason MUST be set to `deleted` if the entity cannot be restored. `"@removed": {"reason": "deleted"}`.

When a link to an entity is deleted, or when the linked entity is deleted, or when a link to an entity is added, implementer MUST return a `property@delta` annotation. e.g. considering the entity Group has a navigation property named members of type Collection(user):

- When a user is added to the group `"members@delta": ["id of the added user"]`
- When a user is removed from the group, or the target user is deleted `"members@delta": ["id of the added user"]`

```

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,114 @@
...
+Implementer MUST implement soft deletion for entities in the backend storage system. The soft deletion will provide useful information to the client to appropriately reflect deletions.
+
+When an entity is soft deleted, the delta function MUST return the id of the deleted entity as well as a `@removed` annotation with the `reason` field.
+- The reason MUST be set to `changed` if the entity can be restored. `"@removed": {"reason": "changed"}`. 
+- The reason MUST be set to `deleted` if the entity cannot be restored. `"@removed": {"reason": "deleted"}`.
+
+When a link to an entity is deleted, or when the linked entity is deleted, or when a link to an entity is added, implementer MUST return a `property@delta` annotation. e.g. considering the entity Group has a navigation property named members of type Collection(user):
+
+- When a user is added to the group `"members@delta": ["id of the added user"]`
```
baywet:
```suggestion
- When a user is added to the group `"members@delta": [{ "@odata.type": "#microsoft.graph.user", "id of the added user"}]`
```

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,114 @@
...
+- When a user is removed from the group, or the target user is deleted `"members@delta": ["id of the added user"]`
```
baywet:
```suggestion
- When a user is removed from the group, or the target user is deleted `"members@delta": ["id of the added user"]`
```
```suggestion
- When a user is removed from the group, or the target user is deleted `"members@delta": [{"@removed": {"reason": "deleted"}, "id of the deleted or removed user"}]`
```

graph/patterns/change-tracking.md
baywet: added above

graph/patterns/change-tracking.md
baywet: added above

OlgaPodo (approved):

PR #305, opened 2021-11-17
Adding design guidelines for modeling collections with state associations

duchau:


mikepizzo (commented):
Thanks for the quick turn around on this @duchau!

See my comments in the email thread. We made some revisions to the guidance in our API Council meeting, largely around adding a memberType enum to the base type, which both makes the payloads easier to read and simplifies querying on all versus an enumerated membership.

duchau:
> Thanks for the quick turn around on this @duchau!
> 
> See my comments in the email thread. We made some revisions to the guidance in our API Council meeting, largely around adding a memberType enum to the base type, which both makes the payloads easier to read and simplifies querying on all versus an enumerated membership.

Addressed in latest iteration.

graph/State-Associated-Subsets.md
```diff
@@ -0,0 +1,113 @@
...
+## Context
+
+A common pattern is to apply a policy or state to a collection of objects. With this, there also comes the question of how to model cases where we want to apply to 'all' or 'none' as well without having to special case these values within the collection set or introducing cross property dependencies. Likewise we'd like to model it in a way where it is easy to understand and interpret usage from just looking at the schema.
+
+A simple example is where you have some policy you need to be able to apply to users in an organization. You may want to support default **None**, enablement for **All** or enablement for **Select** users where you only grant it to a handful of users.
+
+## Problem
+--------
+
+Existing patterns for this either have special-cased 'strings' or have tightly coupled dependencies between two independent properties. Neither are intuitive, require reading documentation and cannot be inferred from the schema or within client libraries.
```
darrelmiller, line 18: The problem statement should be expressing the problem that is being solved by this pattern, not discussing the disadvantages of other patterns that try to solve this problem.

graph/State-Associated-Subsets.md
```diff
@@ -0,0 +1,113 @@
...
+
+The abstract base class should also hold an enum for all possible variants. The purpose of including this is to allow for easier ways to do query and filter operations on variants like 'all' and 'none' without relying on isof functions.
+
+Base Type
+```xml
+    <ComplexType Name="membership" IsAbstract="true">
+      <Property Name="membershipKind" Type="graph.membershipKind"/>
+    </ComplexType>
+
+    <Enum Name="membershipKind">
```
darrelmiller, line 33: As this is duplicating the @OData.Type as a discriminator, we should be explicit as to why we think it is necessary in this case.

graph/State-Associated-Subsets.md
```diff
@@ -0,0 +1,113 @@
+# Pattern Name
+
+Microsoft Graph API Design Pattern
+
+
+
+Modeling state associated to a collection that may include all instances, an included subset, an excluded subset, no instances, or any combinations of the above.
```
darrelmiller, line 7: This pattern doesn't seem to be associating any state to a collection.  It feels more like defining a set of well known predicates for the purpose of determining subsets of a collection.   The state is the predicate attached to the policy (or whatever) that wants to operate on the collection.

graph/State-Associated-Subsets.md
```diff
@@ -0,0 +1,113 @@
...
+## Issues and Considerations
+-------------------------
+
+Given we are using overarching subtype model, subtyping model limitations apply here as well - see [subtyping documentation](https://github.com/microsoft/api-guidelines/blob/op-graphPatterns/graph/Modelling%20with%20Subtypes%20Pattern.md) for more details.
+ 
+
+## When to Use this Pattern
+------------------------
+
+This pattern is to be used when supporting two or more collection states of the following where at least one of the states is a subset variant.
```
gareth-a-jones, line 65: As Darrel points out above, this really isn't a 'state'

graph/State-Associated-Subsets.md
```diff
@@ -0,0 +1,113 @@
...
+{
+    "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#conditionalAccessPolicy",
+    "values": [
+        {
+            "id": "66d36273-fe4c-d478-dc22-e0179d856ce7",
+            "conditions": {
+                "users": {
+                    "includeGuestsOrExternalUsers": {
+                        "externalTenants": {
+                            "@odata.type":"microsoft.graph.conditionalAccessAllExternalTenants",
```
gareth-a-jones, line 90: I am struggling to remember from the discussion Mike led her, but wht is the value of having BOTH the odata.type annotation AND the membershipKind enum?   Isn't the odata.typ eon its own adequate?  Perhaps I am just forgetting the reason.

gareth-a-jones (changes requested):

PR #309, opened 2021-12-06
Graph Dictionary pattern

darrelmiller:
Here's a first effort at converting the existing Dictionary documentation to the pattern format.  I also created a patterns folder for all pattern related docs to live.


graph/patterns/dictionary/index.md
```diff
@@ -1,12 +1,40 @@
...
+--------
+
+The API design requires a resource to include an unknown quantity of data elements of the same type that must be named using values provided by the API consumer.
+
+## Solution
+--------
+
+API designers use a JSON object to represent a dictionary in a `application/json`response payload. When describing the model in CSDL, a new complex type can be created that derives from `Org.OData.Core.V1.Dictionary` and then use the `Org.OData.Validation.V1.OpenPropertyTypeConstraint`to constrain the type that can be used for the values in the dictionary.
+
+Dictionary entries can be added via `POST`, updated via `PATCH`, and they can be removed by setting the entry value to `null`. Multiple dictionaries can be updated at once by using `PATCH` on the dictionary property.
```
gareth-a-jones: Multiple dictionaries?   Multiple entries?

gareth-a-jones: Surprised they support POST.  I would expect only PATCH.

graph/patterns/dictionary/index.md
```diff
@@ -1,12 +1,40 @@
...
 More information:
 
 - [OData reference](https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Core.V1.md#dictionary)
 
-## When to use dictionary types
+ 
+## When to Use this Pattern
+------------------------
 
 Before using a dictionary type in your API definition make sure your scenario fits the following criteria:
```
gareth-a-jones: "The client MUST define the keys of this type. As opposed to the service defining it in advance."
"The client MUST define the keys of this type. As opposed to the service defining them in advance.

graph/patterns/dictionary/index.md
```diff
@@ -147,13 +178,13 @@ PATCH https://graph.microsoft.com/v1.0/users/10/roles/author
...
 
-### Deleting an entry from the dictionary
+#### Deleting an entry from the dictionary
 
 ```HTTP
 DELETE https://graph.microsoft.com/v1.0/users/10/roles/author
 ```
 
-## CDSL example
+### CDSL example
```
gareth-a-jones: ""

No idea what this is about but it needs more explanation.

graph/patterns/dictionary/index.md
darrelmiller: Should be multiple entries.
```suggestion
Dictionary entries can be added via `POST`, updated via `PATCH`, and they can be removed by setting the entry value to `null`. Multiple entries can be updated at once by using `PATCH` on the dictionary property.
```

darrelmiller: Yeah me too.  That's why I have an open question later on in the doc.

graph/patterns/dictionary/index.md
```diff
@@ -1,12 +1,40 @@
...
+
+## Issues and Considerations
+-------------------------
 
 Dictionaries, sometimes called maps, are a collection of name-value pairs. They allow dynamic data sets to be accessed in a systematic manner and are a good compromise between a strictly defined ahead of time structure with all its named properties and between a loosely defined dynamic object (i.e. OData OpenTypes).
 
+As dictionary entries are removed via setting the value to null, this means that dictionaries can only support values that are non-nullable.
+
+  OpenQuestions:
+  -  Can/should PUT be supported on the dictionary property and/or the entry value
```
baywet:
```suggestion
  - Can/should PUT be supported on the dictionary property and/or the entry value 
```

baywet (approved):

graph/patterns/dictionary/index.md
OlgaPodo: Just verifying if a dictionary can be created via POST?

Issue #286, opened 2021-09-08
Determine guidance for top level error code mapping to HTTP status codes


markweitzel:
Determine if top level error codes should map to HTTP status codes or be more granular. Provide a consistent set of guidance for our service team partners.

johanste:
The [OData v4](http://docs.oasis-open.org/odata/odata-json-format/v4.0/os/odata-json-format-v4.0-os.html#_Toc372793091)  specification says:
> The value for the code name/value pair is a language-independent string. Its value is a service-defined error code. This code serves as a sub-status for the HTTP error code specified in the response.

There is no strict requirement that we follow OData to a tee here (we are clearly taking liberties in other areas), but the rationale as to what the value represents should carry over.

PR #311, opened 2021-12-10
Update error guidance for Azure data plane services

### LINKED ISSUE #261: Azure Guidelines: Include Error Codes as part of OpenAPI spec
During one of the API reviews, a question came up regarding the inclusion of service error codes in the OpenAPI specification for a service. In the guidelines, we indicate that the "code" part of the error body IS part of the API contract. Given this, should we recommend that all "codes" be enumerated in the OpenAPI document? Does this impact the breaking change policy? 

```json
{
  "error": {
    "code": "InvalidPasswordFormat",
    "message": "Human-readable description",
    "target": "target of error",
    "innererror": {
      "code": "PasswordTooShort",
      "minLength": 6,
    }
  }
} 
```

tg-msft:
Fixes https://github.com/microsoft/api-guidelines/issues/261 (but not every issue mentioned in the discussion).  I'm specifically not making any accommodations for https://github.com/microsoft/api-guidelines/issues/286 because I prefer the approach clarified here of "HTTP status codes for generic error handling and error codes for specific error handling".


azure/ConsiderationsForServiceDesign.md
```diff
@@ -229,6 +229,33 @@ An important distinction between RELO and status monitor LROs is that there is a
...
+- Accessing resources from multiple microservices can quickly lead to complex race conditions so conditional requests provide optimistic concurrency for safe usage.
+- Reframing the purpose of an API can obviate some errors.  This is most often specific to your operations, but an example from the Guidelines is thinking about `DELETE`s as _"ensure no resource at this location exists"_ so they can return an easier to use `204` instead of _"delete this exact resource instance"_ which would fail with a non-idempotent `404`.
+
+There are two types of errors returned from your service and customers handle them differently.
+- Usage errors where the customer is calling your API incorrectly.  The customer can easily make these errors go away by fixing their code.  We expect most usage errors to be found during testing.
+- Runtime errors that can't be prevented by the customer and need to be recovered from.  Some runtime errors like `429` throttling will be handled automatically by client libraries, but most will be situations like a `409` conflict that requires knowledge about the customer's application to remedy.
```
rysweet: minor: s/conflict that requires/conflict requiring/

azure/ConsiderationsForServiceDesign.md
```diff
@@ -229,6 +229,33 @@ An important distinction between RELO and status monitor LROs is that there is a
 status monitor LRO, whereas the status of all RELO operations is combined into the status of the resource.
 So status monitor LROs are "one-to-one" with their operation status, whereas RELO-style LROs are "many-to-one".
 
+## Errors
+One of the most important parts of service design is also one of the most overlooked.  The errors returned by your service are a critical part of your developer experience.  Your service and your customer's application together form a distributed system.  Errors are inevitable, but well designed errors enabling customers to self-diagnose and recover from faults won't take down their applications or incur overwhelming customer support costs for you.
```
rysweet: minor concision suggestion: "Well-designed errors can help you avoid costly customer support incidents by empowering customers to self-diagnose problems".

azure/ConsiderationsForServiceDesign.md
```diff
@@ -229,6 +229,33 @@ An important distinction between RELO and status monitor LROs is that there is a
...
+
+We also return the top-level error code as the `x-ms-error-code` header so client libraries have the ability to automatically retry requests when possible without having to parse a JSON payload.  We recommend unique error codes like `ContainerBeingDeleted` for every distinct recoverable error that can occur, but suggest reusing common error codes like `InvalidHeaderValue` for usage errors where a descriptive error message is more important for resolving the problem.  The Storage [Common](https://docs.microsoft.com/en-us/rest/api/storageservices/common-rest-api-error-codes) and [Blob](https://docs.microsoft.com/en-us/rest/api/storageservices/blob-service-error-codes) error codes are a good starting point if you're looking for examples.
```
tg-msft: We should include a detailed list of good common error codes in this document rather than linking to Storage.  Tracking at https://github.com/microsoft/api-guidelines/issues/312.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -229,6 +229,33 @@ An important distinction between RELO and status monitor LROs is that there is a
...
+
+First, you should always try to design errors out of existence if possible.  You'll get a lot of this for free by following the Guidelines.  Some examples include:
+- Idempotent APIs solve a whole class of network issues where customers have no idea how to proceed if they send a request to the service but never get a response.
+- Accessing resources from multiple microservices can quickly lead to complex race conditions so conditional requests provide optimistic concurrency for safe usage.
+- Reframing the purpose of an API can obviate some errors.  This is most often specific to your operations, but an example from the Guidelines is thinking about `DELETE`s as _"ensure no resource at this location exists"_ so they can return an easier to use `204` instead of _"delete this exact resource instance"_ which would fail with a non-idempotent `404`.
```
mikekistler: I don't think idempotency requires that the _response_ to a repeated operation in the same -- only that the state on the server is the same.  So calling `404` "non-idempotent" is a bit of a mischaracterization.  I suggest just removing "non-idempotent" here.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -229,6 +229,33 @@ An important distinction between RELO and status monitor LROs is that there is a
...
+
+You should not use your OpenAPI/Swagger spec to document every error that can occur.  The `"default"` response is the only thing AutoRest considers an error response unless you provide other annotations.  Every unique status code turns into a separate code path in your client libraries so we do not encourage this practice.  The only reason to document specific error codes is if they return a different error response than the default, but that is also heavily discouraged.  You can [define an enum in your spec](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/storage/data-plane/Microsoft.BlobStorage/preview/2021-04-10/blob.json#L10419) with `"modelAsString": true` that lists all of the top-level error codes to make it [easier for your customers to handle specific error codes](https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/storage/Azure.Storage.Blobs#troubleshooting).
```
mikekistler: Suggesting minor changes to clarify "status codes" vs "errors"/"error codes":
```suggestion
You generally should not document specific error status codes in your OpenAPI/Swagger spec.  The `"default"` response is the only thing AutoRest considers an error response unless you provide other annotations.  Every unique status code turns into a separate code path in your client libraries so we do not encourage this practice.  The only reason to document specific error  status codes is if they return a different error response than the default, but that is also heavily discouraged. 
```

azure/ConsiderationsForServiceDesign.md
mikekistler: Changing "we" to "you should" here -- to make clear this is their responsibility.
Also moved the last sentence from the following paragraph here, to keep the discussion of "error codes" and "status codes" separate.
```suggestion
You should also return the top-level error code as the `x-ms-error-code` header so client libraries have the ability to automatically retry requests when possible without having to parse a JSON payload.  We recommend unique error codes like `ContainerBeingDeleted` for every distinct recoverable error that can occur, but suggest reusing common error codes like `InvalidHeaderValue` for usage errors where a descriptive error message is more important for resolving the problem.  The Storage [Common](https://docs.microsoft.com/en-us/rest/api/storageservices/common-rest-api-error-codes) and [Blob](https://docs.microsoft.com/en-us/rest/api/storageservices/blob-service-error-codes) error codes are a good starting point if you're looking for examples.
You can [define an enum in your spec](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/storage/data-plane/Microsoft.BlobStorage/preview/2021-04-10/blob.json#L10419) with `"modelAsString": true` that lists all of the top-level error codes to make it [easier for your customers to handle specific error codes](https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/storage/Azure.Storage.Blobs#troubleshooting).
```

azure/Guidelines.md
```diff
@@ -344,6 +349,11 @@ Example:
 
 :heavy_check_mark: **YOU MAY** treat the other fields as you wish as they are _not_ considered part of your service's API contract and customers should not take a dependency on them or their value. They exist to help customers self-diagnose issues.
 
+:heavy_check_mark: **YOU MAY** add additional properties for any data values in your error message so customers don't resort to parsing your error message.  For example, an error with `"message": "A maximum of 16 keys are allowed per account."` might also add a `"maximumKeys": 16` property.  This is not part of your API contract and should only be used for diagnosing problems.
+
+*Note: Do not use this mechanism to provide information developers need to rely on in code (ex: the error message can give details about why you've been throttled, but the `Retry-After` should be what developers rely on to back off).*
+
+:warning: **YOU SHOULD NOT** use your OpenAPI/Swagger specification to document every failing status code or error code for each operation.
```
mikekistler: I think we want this to be a bit stronger.
```suggestion
:warning: **YOU SHOULD NOT** document specific error status codes in your OpenAPI/Swagger spec unless the "default" response cannot properly describe the specific error response (e.g. body schema is different).
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -229,6 +229,33 @@ An important distinction between RELO and status monitor LROs is that there is a
...
+
+There are two types of errors returned from your service and customers handle them differently.
+- Usage errors where the customer is calling your API incorrectly.  The customer can easily make these errors go away by fixing their code.  We expect most usage errors to be found during testing.
+- Runtime errors that can't be prevented by the customer and need to be recovered from.  Some runtime errors like `429` throttling will be handled automatically by client libraries, but most will be situations like a `409` conflict requiring knowledge about the customer's application to remedy.
+
+We provide appropriate [HTTP status codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses) for customers to handle errors generically and error code strings in our common error schema and the `x-ms-error-code` header for customers to handle errors specifically.  As an example, consider what a customer would do when trying to get the properties of a Storage blob:
```
mikekistler:
```suggestion
You should use appropriate [HTTP status codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses) for customers to handle errors generically and error code strings in our common error schema and the `x-ms-error-code` header for customers to handle errors specifically.  As an example, consider what a customer would do when trying to get the properties of a Storage blob:
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -229,6 +229,33 @@ An important distinction between RELO and status monitor LROs is that there is a
...
+
+Be as precise as possible when writing error messages.  A message with just `Invalid Argument` is almost useless to a customer who sent 100KB of JSON to your endpoint.  ``Query parameter `top` must be less than or equal to 1000`` tells a customer exactly what went wrong so they can quickly fix the problem.  Don't go overboard while writing great, understandable error messages and include any sensitive customer information or secrets though.  Many developers will blindly write any error to logs that don't have the same level of access control as Azure resources.
+
```
mikekistler, line 255: I think it is also worth mentioning `x-ms-request-id` here -- something like:
```suggestion

All responses should include the `x-ms-request-id` header with a unique id for the request, but this is particularly important for error responses.  Service logs for the request should contain the `x-ms-request-id` so that support staff can use this value to diagnose specific customer reported errors.
```

azure/ConsiderationsForServiceDesign.md
mikekistler: Would it be unreasonable to simply use the status code name as the code for usage type errors, e.g. "BadRequest"?  Said differently, if an error is not expected to be handled programmatically, then the developer will have the benefit of the error message -- which should be as specific as possible -- to determine the cause of the error.  Is "InvalidHeader" any better that "BadRequest" given that this situation will not be resolved programmatically? Is it even desirable to have more than one error code for "BadRequest" if none can be handled programmatically?

This is the part that still confuses me about the two error types.  If both types get named error codes (other than the status code name), then how can I tell which are expected to be "recoverable"?

mikekistler (commented):
This looks really good.  I left a few comments / suggestions / questions.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -229,6 +229,33 @@ An important distinction between RELO and status monitor LROs is that there is a
 status monitor LRO, whereas the status of all RELO operations is combined into the status of the resource.
 So status monitor LROs are "one-to-one" with their operation status, whereas RELO-style LROs are "many-to-one".
 
+## Errors
+One of the most important parts of service design is also one of the most overlooked.  The errors returned by your service are a critical part of your developer experience.  Your service and your customer's application together form a distributed system.  Errors are inevitable, but well-designed errors can help you avoid costly customer support incidents by empowering customers to self-diagnose problems.
```
markweitzel:
```suggestion
One of the most important parts of service design is also one of the most overlooked.  The errors returned by your service are a critical part of your developer experience and are part of your API contract.  Your service and your customer's application together form a distributed system.  Errors are inevitable, but well-designed errors can help you avoid costly customer support incidents by empowering customers to self-diagnose problems.
```

Introduced the concept that errors are part of the API contract. This will reinforce the breaking change guidance.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -229,6 +229,33 @@ An important distinction between RELO and status monitor LROs is that there is a
...
+
+First, you should always try to design errors out of existence if possible.  You'll get a lot of this for free by following the Guidelines.  Some examples include:
```
markweitzel:
```suggestion
First, you should always try to design errors out of existence if possible.  You'll get a lot of this for free by following the [API Guidelines](https://aka.ms/azapi/guidelines).  Some examples include:
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -229,6 +229,33 @@ An important distinction between RELO and status monitor LROs is that there is a
...
+- Idempotent APIs solve a whole class of network issues where customers have no idea how to proceed if they send a request to the service but never get a response.
+- Accessing resources from multiple microservices can quickly lead to complex race conditions so conditional requests provide optimistic concurrency for safe usage.
```
markweitzel:
```suggestion
- Accessing resources from multiple microservices can quickly lead to complex race conditions. These can be avoided by implementing supporting conditional request through an [optimistic concurrency](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#optimistic-concurrency) strategy, e.g. by leveraging an `ETag` request headers.
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -229,6 +229,33 @@ An important distinction between RELO and status monitor LROs is that there is a
...
+- A `BlobNotFound` or `ContainerNotFound` error code will tell them why the blob doesn't exist so they can take steps to recreate it
+
+The common error schema in the Guidelines allows nested details and inner errors that have their own error codes, but the top-level error code is the most important.  The HTTP status code and the top-level error code are the only part of your error that we consider part of your API contract that follows the same compatibility requirements as the rest of your API.  Importantly, this means you **can not change the HTTP status code or top-level error code for an API in GA'ed service version**.  You can only return new status codes and error codes in future API versions if customers make use of new features that trigger new classes of errors.  Battle tested error handling is some of the hardest code to get right and we can't break that for customers when they upgrade to the latest version.  The rest of the properties in your error like `message`, `details`, etc., are not considered part of your API contract and can change to improve the diagnosability of your service.
+
```
markweitzel, line 249: Should we add some more guidance here around being "conservative" with the errors they introduce and leverage the `inner error` which is not part of the contract? e.g.

Something like... 

```suggestion
Therefore, it is recommended to take a thoughtful, but conservative approach to introducing top level error codes and leverage `inner errors` for managing change over time. 
```

azure/ConsiderationsForServiceDesign.md
markweitzel: @mikekistler & @tg-msft -- Would it make sense if we recommend that "recoverable" errors be documented? This would result in a distinct code path being generated in the SDK? This might be an opportunity for us to add "helpers" to make recovering easier.

azure/Guidelines.md
```diff
@@ -292,11 +292,15 @@ There are 2 kinds of errors:
 
 :white_check_mark: **DO** return an `x-ms-error-code` response header with a string error code indicating what went wrong.
 
-*NOTE: Error code values are part of your API contract (because customer code is likely to do comparisons against them) and cannot change in the future.*
+*NOTE: `x-ms-error-code` values are part of your API contract (because customer code is likely to do comparisons against them) and cannot change in the future.*
```
markweitzel, line 295:
```suggestion
*NOTE: `x-ms-error-code` values are part of your API contract (because customer code is likely to do comparisons against them) and cannot change in the future without a new version of your API*
```

azure/Guidelines.md
```diff
@@ -292,11 +292,15 @@ There are 2 kinds of errors:
...
 
-:white_check_mark: **DO** carefully craft `x-ms-error-code` string values for errors that are recoverable at runtime.
+:heavy_check_mark: **YOU MAY** implement the `x-ms-error-code` values as an enum with `"modelAsString": true` because it's possible add new values over time.  In particular, it's only a breaking change if the same conditions result in a different top-level error code.
```
```suggestion
:heavy_check_mark: **YOU MAY** implement the `x-ms-error-code` values as an [Extensible Enum](#use-extensible-enums)--an enum with `"modelAsString": true`--because it's possible add new values over time.  In particular, it's only a breaking change if the same conditions result in a *different* top-level error code. 
```

markweitzel (changes requested):
@tg-msft -- This is GREAT!!! A huge improvement over what we have there now. I like @mikekistler suggestions as well and also made a few additional comments.

azure/ConsiderationsForServiceDesign.md
robch: should `request` be plural? also ... are both words necessary in `implementing supporting` ?

azure/ConsiderationsForServiceDesign.md
johanste: Yes, spec lawyers will point out that DELETE in HTTP is defined as having to be idempotent. DELETEing the resource more than once doesn't change the state on the server, which is what matters for idempotency purposes. 

We can point out that you can combine conditional requests/If-Match: * and DELETE for clients that really really want to know that the DELETE actually happened. If we want an annotated version of the guidelines.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -229,6 +229,33 @@ An important distinction between RELO and status monitor LROs is that there is a
...
+- A `BlobNotFound` or `ContainerNotFound` error code will tell them why the blob doesn't exist so they can take steps to recreate it
+
+The common error schema in the Guidelines allows nested details and inner errors that have their own error codes, but the top-level error code is the most important.  The HTTP status code and the top-level error code are the only part of your error that we consider part of your API contract that follows the same compatibility requirements as the rest of your API.  Importantly, this means you **can not change the HTTP status code or top-level error code for an API in GA'ed service version**.  You can only return new status codes and error codes in future API versions if customers make use of new features that trigger new classes of errors.  Battle tested error handling is some of the hardest code to get right and we can't break that for customers when they upgrade to the latest version.  The rest of the properties in your error like `message`, `details`, etc., are not considered part of your API contract and can change to improve the diagnosability of your service.
```
drwill-ms: And the top-level exception type, right?

azure/ConsiderationsForServiceDesign.md
MushMal: Q: Wonder if this should be phrased in terms of faults and failures which might or might not result in an API error?

azure/ConsiderationsForServiceDesign.md
```diff
@@ -229,6 +229,33 @@ An important distinction between RELO and status monitor LROs is that there is a
...
+- Accessing resources from multiple microservices can quickly lead to complex race conditions so conditional requests provide optimistic concurrency for safe usage.
+- Reframing the purpose of an API can obviate some errors.  This is most often specific to your operations, but an example from the Guidelines is thinking about `DELETE`s as _"ensure no resource at this location exists"_ so they can return an easier to use `204` instead of _"delete this exact resource instance"_ which would fail with a non-idempotent `404`.
+
+There are two types of errors returned from your service and customers handle them differently.
+- Usage errors where the customer is calling your API incorrectly.  The customer can easily make these errors go away by fixing their code.  We expect most usage errors to be found during testing.
+- Runtime errors that can't be prevented by the customer and need to be recovered from.  Some runtime errors like `429` throttling will be handled automatically by client libraries, but most will be situations like a `409` conflict requiring knowledge about the customer's application to remedy.
```
MushMal, line 242: Is there a ready document that can be linked here describing how the generated clients handle the errors and expose to the higher level app?

MushMal (commented):
Minor questions

azure/ConsiderationsForServiceDesign.md
bojunehsu: What is the recommendation when we have both long-running and "sync" versions of the same operations?  The sync operations may return various 4xx.  But the long-running version would need to return 200 and embed the status code in the inner error.  Thus, should the top level error code more or less reflect HTTP status codes?  Or should we have more specific errors?

Specifically, for [Form Recognizer](https://docs.microsoft.com/en-us/azure/applied-ai-services/form-recognizer/preview-error-guide), should the top-level error codes be InvalidRequest/InvalidArgumen/NotFound/... or ContentSourceNotAccessible/InvalidContentSourceFormat/ModelNotFound/...?

Examples of best practcies will be useful.

azure/ConsiderationsForServiceDesign.md
bojunehsu: Should we consider adding a link to more documentation so customers can self-diagnose?  If we include a link, should we use Markdown syntax?  Here is an example from Azure ML.

image

azure/ConsiderationsForServiceDesign.md
bojunehsu: When we say the message is not part of the API contract and can change, does it mean we can change the messages to a GAed service to improve diagnosability?

JeffreyRichter: Yes, for any given operation, the only parts to the contract are the HTTP status code & the error code string - customers can write code that accesses these values. ALL other values are not part of the contract and may be changed by the service as desired. In our Go SDK, our error type only has fields for the HTTP status code & error code string and then raw response - we do no expose the other fields because we don't want customers to write code that access them. They can get the values from the raw response which will give them a JSON string which they have to parse to do whatever they want to do with it.

azure/ConsiderationsForServiceDesign.md
JeffreyRichter: Mike: I don't think that returning BadRequest for 400 is useful.  The [HTTP spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html) mandates that all HTTP responses have a status code and a "reason phrase". For 400, The reason phrase will already have "Bad Request" in it. Adding it again in the error message isn't useful and bloats what's sent over the wire. However, having the reason for the bad request is useful in the error code/message.

azure/Guidelines.md
JeffreyRichter: You cannot change them even if there is a new api-version. Doing so is breaking.

azure/Guidelines.md
JeffreyRichter: I think it would be nice to separate the error concepts from the swagger suggestions. The service implementors need to know what's in the contract and how to craft an error. This is not related to swagger.
The swagger is how to describe the implementation and explaining these rules is useful but I think should be separated out because different people are likely to care about these 2 things differently.

JeffreyRichter (approved):

azure/ConsiderationsForServiceDesign.md
tg-msft: This is an area where our default client experience for a 204 + 404 DELETE feels very non-idempotent.  Getting a 204 response on your first request and an exception that hijacks your control flow on the second feels like you're seeing different results for the same request.  Removed the `non-idempotent` though to avoid confusing folks and/or inspiring this debate again in the future.  :smile:

azure/ConsiderationsForServiceDesign.md
tg-msft: Applied, but I left out `generally` so the wording was stronger.

azure/ConsiderationsForServiceDesign.md
tg-msft: I'm going to punt this to https://github.com/microsoft/api-guidelines/issues/312 where we can discuss whether we only want `400`/`BadRequest` or to go further with `400`/`InvalidValue`, `400`/`MissingRequiredValue`, etc.  I personally like the "more specific generic error codes" and see it as a happy medium between the too generic `BadRequest` and too specific `ContainerNameMustHaveBetween3And32Characters`.  I don't think we can tell people to Goldilocks their APIs until they're just right though, so https://github.com/microsoft/api-guidelines/issues/312 as an example is probably our best bet.

As for documenting recoverable errors, I'm going to punt that to [my side project](https://microsoft-my.sharepoint.com/:p:/p/teglaza/EeTYvK4JCJNIjtA1BLufg34B4MROlR-kaUACj3HqXidpNQ?e=Resebh) (see slide 34 in particular).

azure/ConsiderationsForServiceDesign.md
tg-msft: I've got a lot of thoughts about troubleshooting documentation that are outside the scope of these changes.  If we have additional pointers today, I personally don't like the idea of using markdown because I want error messages to be as human readable as possible.  Seeing `You can find troubleshooting steps [here](https://www.azurehelp.net/storage/blobs/blob_delete/404/BlobNotFound)` doesn't do more than `You can find troubleshooting steps at https://www.azurehelp.net/storage/blobs/blob_delete/404/BlobNotFound`.  I know at least VS and VS Code make hyperlinks in error messages easily clickable as well.

azure/ConsiderationsForServiceDesign.md
tg-msft: I also wondered how much to frame this from a distributed systems perspective.  I ultimately don't think most of the people reading this document or coming for reviews are distributed systems experts and tried to aim it more toward an average C# developer.  We could probably include links to resources like Chapter 3 of Jim Gray's book if people want to dig deeper into fault tolerance, etc., but the rest of this document is pretty high level at the moment.

azure/ConsiderationsForServiceDesign.md
tg-msft: I actually fall on the other side of this issue.  Since they can never change an error code, I'd rather they introduce more top-level error codes today for anything recoverable.  Customers have HTTP status codes for generic error handling but we won't ever get more specific than what we release as our first top-level error code.

azure/Guidelines.md
tg-msft: That's a fair suggestion, but I'd still say if the service designers aren't thinking about how they'll eventually do this in the swagger then we'll be told it's an impossible amount of work to pull the list of errors out of the service's front end.  Would you separate this out into a separate guideline?  Or remove from the document entirely?

azure/ConsiderationsForServiceDesign.md
tg-msft: The top-level exception type is going to be a language by language thing and that'll version according to the language guidelines.  C#, for example, has lots of specific rules in section D.12 at https://aka.ms/fxdg3/breaking (visit https://aka.ms/oreilly once for SSO).

Yes, it's possible to change error messages in GA'ed services to make them more diagnosable.  I'd probably recommend waiting until the next version bump if possible though because _someone_ is still probably parsing it even though they shouldn't.

azure/ConsiderationsForServiceDesign.md
tg-msft: We have things like https://azure.github.io/azure-sdk/general_implementation.html#error-handling, https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-errors, and https://docs.microsoft.com/en-us/dotnet/api/azure.requestfailedexception?view=azure-dotnet, but those don't really talk about the underlying mechanism we use to decide if an error is retriable.  We don't actually recommend most services change from https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/ResponseClassifier.cs#L19 though.

azure/ConsiderationsForServiceDesign.md
tg-msft: I don't see the sync/async aspect as any different.  I'd return a `404` with an error code `ModelNotFound` from the sync API and embed both of those values in your LRO's `GetStatus`.

azure/ConsiderationsForServiceDesign.md
bojunehsu: To confirm, we can/should include hyperlinks to external content in error messages.  But we should use plain URL and let the UI/tool render them as clickable links.  One argument for using plain URL is that UI/tool would need to parse Markdown otherwise.  And only parsing URLs would be awkward.

azure/ConsiderationsForServiceDesign.md
bojunehsu: I understand that we can change the error message in the next API version.  But to improve customer satisfaction, can we make hotfixes to improve the error message for previously released GA versions?  Specifically, can we add a hyperlink to the error message to provide additional information?

azure/ConsiderationsForServiceDesign.md
bojunehsu: When the LRO fails, it is my understanding that GetStatus should return 200 since the GetStatus operation itself succeeded.  Instead, we return the error via the _error_ property.  Thus, there is no place to return 404.  Is my understanding incorrect?

azure/ConsiderationsForServiceDesign.md
mikekistler: @bojunehsu OpenAPI explicitly allows markdown in many of the `description` elements of the doc -- I don't think we should advise against using it. Rather, if our tooling is lacking in support of markdown then lets identify those gaps and see if we can get them fixed.

bojunehsu: We are currently referring to the custom-defined _message_ property, which doesn't need to follow OpenAPI conventions.

On the separate topic of markdown in Swagger descriptions, below are places where our tooling currently does not support markdown:
* API Management - We can import APIs definitions from Markdown.  If I recall correctly, the imported documentation does not properly render markdown.
* Autorest - The description properties are used to populate the documentation rendered in tooltips.  VS tooltips does not support Markdown (it would be very awkward).  I don't recall how markdown documentation is rendered in the source code comments.
* JSON Editor - Viewing/editing markdown in JSON editors is not a pleasant experience.  The entire markdown needs to be in a single line, using escaped newlines.   It is nearly usable.  YAML has better Markdown support.

tg-msft: https://github.com/Azure/autorest.csharp/issues/938 is tracking translation of markdown to C# doc comments in AutoRest - though we need similar support across languages.

azure/ConsiderationsForServiceDesign.md
tg-msft: @JeffreyRichter should answer definitively, but I think it's a reasonable to update an error message without a version bump if you think there's significant customer value.  I'd still try to append to rather than edit the existing error message if possible to minimize the potential impact when we're changing without a version bump.

azure/ConsiderationsForServiceDesign.md
tg-msft: I think we should just include the status code as an additional property inside the 200's error property.  We do that in many of our batching APIs for a similar example that's worked well.  If folks agree, I think we'd want that recommendation in the LRO section.

azure/ConsiderationsForServiceDesign.md
JeffreyRichter: Yes, as error messages are not part of the contract, they can be updated without a version bump. Just like improving the performance of an operation can be done without a version bump since perf is not part of the contract either.

azure/Guidelines.md
```diff
@@ -306,7 +310,7 @@ There are 2 kinds of errors:
 
 Property | Type | Required | Description
 -------- | ---- | :------: | -----------
-`error` | ErrorDetail | ✔ | The error object.
+`error` | ErrorDetail | ✔ | The top-level error object whose `code` matches `x-ms-error-code`
```
BlackRider97: `x-ms-error-code` response header

azure/ConsiderationsForServiceDesign.md
```diff
@@ -229,6 +229,34 @@ An important distinction between RELO and status monitor LROs is that there is a
...
+- A `BlobNotFound` or `ContainerNotFound` error code will tell them why the blob doesn't exist so they can take steps to recreate it
+
+The common error schema in the Guidelines allows nested details and inner errors that have their own error codes, but the top-level error code is the most important.  The HTTP status code and the top-level error code are the only part of your error that we consider part of your API contract that follows the same compatibility requirements as the rest of your API.  Importantly, this means you **can not change the HTTP status code or top-level error code for an API in GA'ed service version**.  You can only return new status codes and error codes in future API versions if customers make use of new features that trigger new classes of errors.  Battle tested error handling is some of the hardest code to get right and we can't break that for customers when they upgrade to the latest version.  The rest of the properties in your error like `message`, `details`, etc., are not considered part of your API contract and can change to improve the diagnosability of your service.
```
BlackRider97: [common error schema in the Guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#handling-errors)

BlackRider97: `code`

BlackRider97: instead of this can we rephrase sentence changing to 
changing the HTTP status code or top-level error `code` or `x-ms-error-code` response header  will be considered as a API breaking change

Point is that this guidelines also applies for Public Preview or Preview preview services

BlackRider97: Is it worth mentioning common-type error object schema here ? 
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/common-types/data-plane/v1/types.json

azure/ConsiderationsForServiceDesign.md
```diff
@@ -229,6 +229,34 @@ An important distinction between RELO and status monitor LROs is that there is a
...
+
+You should also return the top-level error code as the `x-ms-error-code` header so client libraries have the ability to automatically retry requests when possible without having to parse a JSON payload.  We recommend unique error codes like `ContainerBeingDeleted` for every distinct recoverable error that can occur, but suggest reusing common error codes like `InvalidHeaderValue` for usage errors where a descriptive error message is more important for resolving the problem.  The Storage [Common](https://docs.microsoft.com/rest/api/storageservices/common-rest-api-error-codes) and [Blob](https://docs.microsoft.com/rest/api/storageservices/blob-service-error-codes) error codes are a good starting point if you're looking for examples.  You can [define an enum in your spec](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/storage/data-plane/Microsoft.BlobStorage/preview/2021-04-10/blob.json#L10419) with `"modelAsString": true` that lists all of the top-level error codes to make it [easier for your customers to handle specific error codes](https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/storage/Azure.Storage.Blobs#troubleshooting).
```
BlackRider97: `x-ms-error-code` response header

BlackRider97 (approved):

azure/ConsiderationsForServiceDesign.md
markweitzel: LGTM.

markweitzel: Agree. Let's provide this updated guidance and move this to #312 .

azure/ConsiderationsForServiceDesign.md
markweitzel: The updated guidance is good. I don't think we should return formatting, MD, HTML, etc. in the returned text.

azure/Guidelines.md
markweitzel: @JeffreyRichter -- We already talk about `Extensible Enums` in the spec, see: [Extensible Enums](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#use-extensible-enums). These are referenced in a number of places. We did this, in part, because at the time, we didn't have the [Open API Style Guidelines](https://aka.ms/azapi/style). What I'd like to do is be consistent within the Guidelines that we have now, and point to the extensible enum section. We will open a new issue to determine if we should refactor the guidelines to remove the section to the style guide--which we can do on a separate pass. Captured this thought in a new issue: #316.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -229,6 +229,34 @@ An important distinction between RELO and status monitor LROs is that there is a
...
+- Reframing the purpose of an API can obviate some errors.  This is most often specific to your operations, but an [example from the API Guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#http-return-codes) is thinking about `DELETE`s as _"ensure no resource at this location exists"_ so they can return an easier to use `204` instead of _"delete this exact resource instance"_ which would fail with a `404`.
+
+There are two types of errors returned from your service and customers handle them differently.
+- Usage errors where the customer is calling your API incorrectly.  The customer can easily make these errors go away by fixing their code.  We expect most usage errors to be found during testing.
+- Runtime errors that can't be prevented by the customer and need to be recovered from.  Some runtime errors like `429` throttling will be handled automatically by client libraries, but most will be situations like a `409` conflict requiring knowledge about the customer's application to remedy.
+
```
mikekistler, line 243: I like the distinction between usage errors and runtime errors -- I like it so much that I think our error design should clearly indicate what category each error falls into.  I opened #315 to track this.

azure/ConsiderationsForServiceDesign.md
markweitzel: Captured this thread as issue #317

azure/ConsiderationsForServiceDesign.md
markweitzel: If we reference a common error type, it should be what is in the [Handling Errors](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#handling-errors) section. We've had a number of discussions about supporting common types and the consensus thus far is to declare them in the spec vs. providing a referenceable schema.

tg-msft: I'm not clear on what this comment refers to.  _(Consider using [suggestions](http://haacked.com/archive/2019/06/03/suggested-changes/) in the future for fine grained edits to large walls of text to make it easier to pinpoint.)_

markweitzel (approved):
@tg-msft -- Thank you for pulling this together and working through the reviews. This is a great addition to our content and is another step at clarifying our guidance.

PR #314, opened 2021-12-27
Adding descriptions for x-ms-*-request-id headers

johanste:
...intended usage and descriptions.


azure/Guidelines.md
```diff
@@ -993,6 +995,14 @@ Client libraries are required to send telemetry and distributed tracing informat
 - [Azure SDK Distributed tracing policy](https://azure.github.io/azure-sdk/general_azurecore.html#distributed-tracing-policy)
 - [Open Telemetry](https://opentelemetry.io/)
 
+In addition to distributed tracing, Azure also uses a set of common correlation headers:
+
+|Name                         |Applies to|Description|
+|-----------------------------|----------|-----------|
+|x-ms-client-request-id.      |Both      |Optional. Caller-specified value identifying the request, in the form of a GUID with no decoration such as curly braces (e.g. `x-ms-client-request-id: 9C4D50EE-2D56-4CD3-8152-34347DC9F2B0`). If the caller provides this header the service **must** log this with their traces to facilitate tracing a single request. Because this header can be client-generated, it should not be assumed to be unique by the service implementation.
+|x-ms-return-client-request-id|Request.  |Optional. If specified, the service **must** include a `x-ms-client-request-id` header with the corresponding value in the response. Allows clients to correlate responses with requests.
```
BlackRider97: If `true` or `false` value specified, the service ...

azure/Guidelines.md
```diff
@@ -993,6 +995,14 @@ Client libraries are required to send telemetry and distributed tracing informat
 - [Azure SDK Distributed tracing policy](https://azure.github.io/azure-sdk/general_azurecore.html#distributed-tracing-policy)
 - [Open Telemetry](https://opentelemetry.io/)
 
+In addition to distributed tracing, Azure also uses a set of common correlation headers:
+
+|Name                         |Applies to|Description|
+|-----------------------------|----------|-----------|
+|x-ms-client-request-id.      |Both      |Optional. Caller-specified value identifying the request, in the form of a GUID with no decoration such as curly braces (e.g. `x-ms-client-request-id: 9C4D50EE-2D56-4CD3-8152-34347DC9F2B0`). If the caller provides this header the service **must** log this with their traces to facilitate tracing a single request. Because this header can be client-generated, it should not be assumed to be unique by the service implementation.
```
BlackRider97: This header should be returned in response only if `x-ms-return-client-request-id` request header set to `true` ?

BlackRider97: x-ms-client-request-id

Please remove trailing dot `.`

azure/Guidelines.md
```diff
@@ -993,6 +995,14 @@ Client libraries are required to send telemetry and distributed tracing informat
...
+|x-ms-return-client-request-id|Request.  |Optional. If specified, the service **must** include a `x-ms-client-request-id` header with the corresponding value in the response. Allows clients to correlate responses with requests.
+|x-ms-request-id              |Response  |Required. Service generated correlation id identifying the request, in the form of a GUID with no decoratoin such as curly braces. In contrast to the the `x-ms-client-request-id`, the service **must** ensure that this valu is globally unique. Services should log this value with their traces to facilitate tracing of a single request.
```
BlackRider97: this value is globally

BlackRider97: facilitate distributed tracing of a single request

azure/Guidelines.md
JeffreyRichter: I don't think this request header is useful. Why not just always return the x-ms-client-request-id and simplify what a service has to implement? If the client doesn't send the ID, then the service doesnt' have to return it (since it doesn[t even have it).

johanste: This is an existing header recommendation (see aka.ms/rpapiv2). We can chose to not recommend its use and I would be fine with that.

azure/Guidelines.md
johanste: Correct.

azure/Guidelines.md
JeffreyRichter: I see. My preference would be to not mention this header at all

mikekistler: I also prefer to drop this header. There is no harm in returning `x-ms-client-request-id` if the caller passed it, so for the sake of simplicity we should recommend that it be returned whenever it was specified on the request.

azure/Guidelines.md
mikekistler: Incorporating @BlackRider97 's comments with one more suggestion of mine.
```suggestion
|x-ms-request-id              |Response  |Required. Service generated correlation id identifying the request, in the form of a GUID with no decoration such as curly braces. In contrast to the the `x-ms-client-request-id`, the service **must** ensure that this value is globally unique. Services should log this value with their traces to facilitate distributed tracing of a single request.
```

mikekistler (approved):
Looks good. I left a few more comments on top of previous reviewers.

Approving now to avoid re-review after comments are addressed.

azure/Guidelines.md
johanste: Removed

azure/Guidelines.md
johanste: Fixed.

azure/Guidelines.md
johanste: I'd like to distinguish x-ms-client-request-id from "real" distributed tracing (e.g. OpenTelemetry). 

Changed to: Services should log this value with their traces to facilitate correlation of log entries for a single request.

rakelkar:
What is the server behavior if the user specifies `x-ms-request-id` as part of the request?

JeffreyRichter:
If the client sends any invalid header, then the service should return 400 with an error code/message

rakelkar:
Is this considered an invalid header?

JeffreyRichter:
x-ms-request-id is a response-only header; it is invalid in a request

markweitzel (approved):

azure/Guidelines.md
```diff
@@ -993,6 +994,13 @@ Client libraries are required to send telemetry and distributed tracing informat
 - [Azure SDK Distributed tracing policy](https://azure.github.io/azure-sdk/general_azurecore.html#distributed-tracing-policy)
 - [Open Telemetry](https://opentelemetry.io/)
 
+In addition to distributed tracing, Azure also uses a set of common correlation headers:
+
+|Name                         |Applies to|Description|
+|-----------------------------|----------|-----------|
+|x-ms-client-request-id       |Both      |Optional. Caller-specified value identifying the request, in the form of a GUID with no decoration such as curly braces (e.g. `x-ms-client-request-id: 9C4D50EE-2D56-4CD3-8152-34347DC9F2B0`). If the caller provides this header the service **must** include this in their log entries to facilitate correlation of log entries for a single request. Because this header can be client-generated, it should not be assumed to be unique by the service implementation.
```
tg-msft, line 1060: For most services this is a free form string rather than a GUID.  Customers can put whatever they want in here and many use it for crazy things (like tagging so they can grep their logs for analytics rather than use the dozen or ways we make that data so much easier to find).

azure/Guidelines.md
```diff
@@ -993,6 +994,13 @@ Client libraries are required to send telemetry and distributed tracing informat
...
+|x-ms-request-id              |Response  |Required. Service generated correlation id identifying the request, in the form of a GUID with no decoration such as curly braces. In contrast to the the `x-ms-client-request-id`, the service **must** ensure that this value is globally unique. Services should log this value with their traces to facilitate correlation of log entries for a single request.
```
tg-msft, line 1061: Do we want to recommend how long this data should be kept around for service supportability?  I think Jeff says a minimum of 24 hours in other places?

azure/Guidelines.md
```diff
@@ -167,25 +167,26 @@ Array      | One of a) a comma-separated list of values (preferred), or b) separ
...
+accept                  | Request    | application/json
+If-Match                | Request    | "67ab43" or * (no quotes) (see [Conditional Requests](#Conditional-Requests))
+If-None-Match           | Request    | "67ab43" or * (no quotes) (see [Conditional Requests](#Conditional-Requests))
+If-Modified-Since       | Request    | Sun, 06 Nov 1994 08:49:37 GMT (see [Conditional Requests](#Conditional-Requests))
+If-Unmodified-Since     | Request    | Sun, 06 Nov 1994 08:49:37 GMT (see [Conditional Requests](#Conditional-Requests))
+date                    | Both       | Sun, 06 Nov 1994 08:49:37 GMT (see [RFC7231, Section 7.1.1.2](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.1.2))
+_content-type_          | Both       | application/merge-patch+json
+_content-length_        | Both       | 1024
+_x-ms-request-id_       | Response   | 4227cdc5-9f48-4e84-921a-10967cb785a0 (see [Distributed Tracing & Telemetry](#Distributed-Tracing--Telemetry))
+_x-ms-client-request-id_| Both       | 227cdc5-9f48-4e84-921a-10967cb785a1 (see [Distributed Tracing & Telemetry](#Distributed-Tracing--Telemetry))
```
tg-msft: I ran your GUID through an online plagiarism detector and it turns out you just copy/pasted the one above with a character tweaked on the end...  but you also accidentally dropped a character from the first segment so it's no longer a valid GUID?  I can't tell if it was a hasty accident or you're playing mind games with readers, but either way, here's an heirloom GUID my family has trusted for generations in case you need one: `2a7d2b1a-cca4-49e4-a958-57fe0f786a2a`.

PR #325, opened 2022-02-01
Fix breaking changes link

mikekistler:
Use the correct shortlink for the "current" version of the Azure Breaking Changes policy doc.

The previous link went to version 1.2, but the current version is 1.2.1.

### DIFF
--- a/azure/README.md
+++ b/azure/README.md
@@ -5,7 +5,7 @@ Designing powerful APIs with strong defaults, consistent behavior across related
 * [Considerations for Service Design](ConsiderationsForServiceDesign.md)
 * [REST API Guidelines](Guidelines.md)
 * [OpenAPI Style Guidelines](https://github.com/Azure/azure-api-style-guide/blob/main/README.md)
-* [Breaking Changes](https://aka.ms/azapi/breakingchanges)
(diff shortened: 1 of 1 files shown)


heaths (approved):

markweitzel (approved):

azure/README.md
```diff
@@ -5,7 +5,7 @@ Designing powerful APIs with strong defaults, consistent behavior across related
 * [Considerations for Service Design](ConsiderationsForServiceDesign.md)
 * [REST API Guidelines](Guidelines.md)
 * [OpenAPI Style Guidelines](https://github.com/Azure/azure-api-style-guide/blob/main/README.md)
-* [Breaking Changes](https://aka.ms/azapi/breakingchanges)
+* [Breaking Changes](http://aka.ms/AzBreakingChangesPolicy/)
```
heaths, line 8: @markweitzel I still think in the aka link manager you should direct https://aka.ms/azapi/breakingchanges and any other permutations to http://aka.ms/AzBreakingChangesPolicy/. There's no telling where else those variants are used, and this way they all point to a single aka link (http://aka.ms/AzBreakingChangesPolicy/) you can manage instead.

PR #327, opened 2022-02-03
publish Microsoft Graph specific guidelines and patterns

OlgaPodo:


graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
...design of your API is arguably the most important investment you will make. API design is what creates the first impression for developers when they discover and learn how to use your APIs. We promote API-first design approach where you begin your product design by focusing on how information will be exchanged and represented and creating an interface contract for your API which is followed by design and implementation of the backing service. This approach ensures decoupling of the interface from your implementation and is essential for agility, predictability, and reuse of your APIs. Established interface contract allows developers to use your API while internal teams are still working on implementation, API specifications enable designing of user experience and test cases in parallel. Starting with user-facing contracts also promotes a good understanding of system interactions, your modeling domain, and understanding of how the service will evolve. Microsoft Graph supports resource and query-based API styles which follow HTTP, REST, and JSON standards, where API contract is described using ODATA conventions and schema definition (see Documentation · OData - the Best Way to REST).
```
```suggestion
The design of your API is arguably the most important investment you will make. API design is what creates the first impression for developers when they discover and learn how to use your APIs. We promote API-first design approach where you begin your product design by focusing on how information will be exchanged and represented and creating an interface contract for your API which is followed by design and implementation of the backing service. This approach ensures decoupling of the interface from your implementation and is essential for agility, predictability, and reuse of your APIs. Established interface contract allows developers to use your API while internal teams are still working on implementation; API specifications enable designing of user experience and test cases in parallel. Starting with user-facing contracts also promotes a good understanding of system interactions, your modeling domain, and understanding of how the service will evolve. Microsoft Graph supports resource and query-based API styles which follow HTTP, REST, and JSON standards, where API contract is described using ODATA conventions and schema definition (see Documentation · OData - the Best Way to REST).
```

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+[Documentation · OData - the Best Way to REST](https://www.odata.org/documentation/)).
+
+In general API design includes the following steps:
```
dkershaw10, line 76: For me, one key design element is mapping end-user scenarios/use cases to request/response patterns to understand what the API consumer experience would be.  That doesn't seem to factor into this list though, although it's prevalent in next paragraph.  For those scanning the doc, it would be good to have a bullet here on "user scenarios being the foundation of API design".

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+-   Specify errors
+
+When creating your API contract you will define resources based on the domain model supporting your service and identify interactions based on user scenarios. Good API design goes beyond modeling the current state of resources and it is important to plan ahead how API evolves. For this it is essential to understand and document your user scenarios as the foundation of the API design. There is no one-to-one correspondence between domain model elements and API resources because you should simplify your customer facing APIs for better usability and to obfuscate implementation details. We recommend creating a simple resource diagram, like below, to show resources and their relationships and make it easier to reason about modeling choices and the shape of your API.
+![Resource model example](ModelExample.png)
+
+After resources are defined it’s time to think about the behavior of your API which can be expressed via HTTP methods and operational resources such as functions and actions. As you think about API behavior you identify a happy path and various exceptions and deviations which will be expressed as errors and represented using HTTP codes and error messages.
```
dkershaw10, line 95: Would it be better to describe these as "custom methods" for the lay person?

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+
+-   **pathSegment** is the last navigation segment which can address an entity,
+    collection of entities, property or operation available for an entity
+
+-   **query string** must follow the OData standard for query representations
+    and is covered in [Query](#query) section of OData specifications.
+
+While HTTP defines no constraints on how different resources are related
+together, it does encourage the use of URL path segment hierarchies to convey
+relationships. In Microsoft Graph lifetime relationships between resources are
```
dkershaw10: This "lifetime relationships" statement doesn't really make sense to me.  Not sure that it adds value or would make sense to our readers.  When you talk about lifetime relationships, I'm thinking more about parent-child URL relationships, where if the parent is deleted, so are its children.  However, not sure this is the point to introduce this concept.

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+| :ballot_box_with_check: **SHOULD** prefix Boolean properties with is, unless this leads to awkward or unnatural sounding names for Boolean properties. | - **Right:** isEnabled or isResourceAccount <BR>- **Wrong** enabled or allowResourceAccount <BR> - **Right:** allowNewTimeProposals or allowInvitesFrom — subjectively more natural than the examples below <BR> - **Wrong** isNewTimeProposalsAllowed or isInvitesFromAllowed — subjectively more awkward that the examples above |
+| :no_entry: **MUST NOT** use 'collection', 'response', 'request ' suffixes .                                          |- **Right:** addresses <BR> - **Wrong** addressCollection                                                                                                                                                                                                          |
+
+### Uniform Resource Locators (URLs)
+
+A Uniform Resource Locator (URL) is how developers access the resources of your
+API.
+
+Navigation paths to Microsoft Graph resources are generally broken into multiple
```
dkershaw10, line 133: I can't find anywhere where we talk about collections or identifiers/keys, although pretty sure this is in the top-level Microsoft REST API guidelines

We talk about a pathSegment that can address an entity or collection of entities, but we don't provide any examples.  Should we split pathSegment  into:

- address a collection of entities like `/incidents`
- address an entity in a collection like `/incidents/{id}`
- address an operation like `incidents/{id}/resolve`

Or we could reference the correct section in the Microsoft REST API guidelines?

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+OData specifications and [Microsoft REST API
+Guidelines](https://github.com/microsoft/api-guidelines/blob/master/Guidelines.md#7102-error-condition-responses).
+|Requirements|
+|----------------------------------------------------------------------------------------------------|
+| :heavy_check_mark: **MUST** support \$select on resource to enable properties projection |
+| :ballot_box_with_check: **SHOULD** support \$filter with eq, ne operations on properties of entities for collections| 
+| :heavy_check_mark: **MUST** support server-side pagination for collections |
+| :ballot_box_with_check: **SHOULD** support pagination $top, $skip and $count for collections |
+
+The query options part of an OData URL can be quite long, potentially exceeding
```
dkershaw10, line 189: Interesting that we propose this.  This is likely supported by very few workloads.  If we are trying to get more folks to support this then we need to make this a stronger recommendation. 

What we recommend publicly is the use of  batch in this situation:  https://docs.microsoft.com/en-us/graph/json-batching#bypassing-url-length-limitations-with-batching

Not sure why this isn't an issue when the front-end routes to an internal service, but...

graph/GuidelinesGraph.md
OlgaPodo: Agree.

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+    request body or completely in the request url. Graph doesn't support query
+    options present in both places.
+
+-   The parameters in \$query segment should not span multiple workloads. Support for
+    \$query segment right now is limited to properties belonging to the same workload.
+
+
+### Resource Modeling Patterns
+
+You can model complex resources for your APIs using OData Entity Type or Complex Type. The main difference between these types is that Entity type declares a key property to uniquely identify its objects and  Complex Type does not. In Microsoft Graph this key property has "id" as a prescribed name.
```
dkershaw10: complex resources or structured resources?

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+    <Property Name="address" Type="microsoft.graph.Address" />
+</EntityType>
+<ComplexType Name="Address">
+    <Property Name="city" Type="Edm.String" />
+    <Property Name="street" Type="Edm.String" />
+    <Property Name="stateOrProvince" Type="Edm.String" />
+    <Property Name="country" Type="Edm.String" />
+</ComplexType>
+```
+|  Microsoft Graph rules for modeling complex resources                                  |       |
```
dkershaw10: Nit:  table has an extra empty column

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+have with your service. As an API provider, how you manage the overall request /
+response pattern should be one of the first implementation decisions you make.
+APIs SHOULD use resource-based designs with standard HTTP methods rather than operation resources if possible.
+ Operation resources are either functions or actions. According to [ODATA standards]( http://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part3-csdl/odata-v4.0-errata03-os-part3-csdl-complete.html#_The_edm:Function_Element_2) a function represents an operation which returns a single instance or collection of instances of any type and doesn’t have an observable side effect. An action may have side effects and may return a result represented as a single entity or collection of any type.
+
+|  Microsoft Graph rules for modeling behavior                                                                                          |
+|-----------------------------------------------------------------------------------------------------------------|
+| :heavy_check_mark: **MUST** use POST to create new entities in insertable entity sets or collections |
```
dkershaw10: I think we should mention that in Graph the majority of new entity creation through this mechanism results in the response containing a system generated unique identifier.

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+
+|  Microsoft Graph rules for modeling behavior                                                                                          |
+|-----------------------------------------------------------------------------------------------------------------|
+| :heavy_check_mark: **MUST** use POST to create new entities in insertable entity sets or collections | 
+| :heavy_check_mark: **MUST** use PATCH to edit updatable resources                                                 | 
+| :heavy_check_mark: **MUST** use DELETE to delete deletable resources                                              | 
+| :heavy_check_mark: **MUST** use GET for listing and reading resources. | 
+| :warning: **SHOULD NOT** use PUT for updating resources.                                       | 
+| :ballot_box_with_check: **SHOULD** avoid using multiple round trips to complete a single logical action.       |
```
dkershaw10: action -> operation?
Otherwise this might be confused with an OData service **action**

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+|-----------------------------------------------------------------------------------------------------------------|
+| :heavy_check_mark: **MUST** use POST to create new entities in insertable entity sets or collections | 
+| :heavy_check_mark: **MUST** use PATCH to edit updatable resources                                                 | 
+| :heavy_check_mark: **MUST** use DELETE to delete deletable resources                                              | 
+| :heavy_check_mark: **MUST** use GET for listing and reading resources. | 
+| :warning: **SHOULD NOT** use PUT for updating resources.                                       | 
+| :ballot_box_with_check: **SHOULD** avoid using multiple round trips to complete a single logical action.       | 
+
+
+Bound operations must have a binding parameter matching the type of the bound resource.
```
dkershaw10: Consistent language - what is a bound operation?  Before we were using the term "resource operation",

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+| :heavy_check_mark: **MUST** use PATCH to edit updatable resources                                                 | 
+| :heavy_check_mark: **MUST** use DELETE to delete deletable resources                                              | 
+| :heavy_check_mark: **MUST** use GET for listing and reading resources. | 
+| :warning: **SHOULD NOT** use PUT for updating resources.                                       | 
+| :ballot_box_with_check: **SHOULD** avoid using multiple round trips to complete a single logical action.       | 
+
+
+Bound operations must have a binding parameter matching the type of the bound resource. 
+In addition both actions and functions support overloading, meaning an API definition may contain multiple actions or functions with the same name.
+Microsoft Graph supports the use of optional parameters. You can use the optional parameter annotation instead of creating function or action overloads.
```
dkershaw10: Can we leave it at the first sentence?

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+                           
+                }
+    }
+}
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The top-level error code must be aligned with HTTP response status codes according to [rfc7231 (ietf.org)](https://datatracker.ietf.org/doc/html/rfc7231#section-6). 
+The following examples demonstrate error modeling for common use cases:
+
+-   **Simple error**: A workload wants to report an error with top-level details
```
dkershaw10: Nit:  Not sure we should use the term "workload".  Suggest "service" or "API producer".  Later it says "An API".  Let's be consistent.

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+           
+    }
+  }
+}
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+| Microsoft Graph enforces the following error rules                                                                    | 
+|-----------------------------------------------------------------------------------------------------------------------|
+| :heavy_check_mark: **MUST** return an error property with a child code property in all error responses.                 | 
+| :heavy_check_mark: **MUST** return a 403 Forbidden error when insufficient scopes are present in the auth token.        |
```
dkershaw10: Can we call it "insufficient permissions"?  
There are other valid conditions for 403.  While your statement is true...

1. Signed-in user is not authorized to access the data/API
2. Signed-in user is not licensed to access the data/API

I wonder if a more all-encompassing statement would be "when the application or signed-in user have insufficient permissions".

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+    }
+  }
+}
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+| Microsoft Graph enforces the following error rules                                                                    | 
+|-----------------------------------------------------------------------------------------------------------------------|
+| :heavy_check_mark: **MUST** return an error property with a child code property in all error responses.                 | 
+| :heavy_check_mark: **MUST** return a 403 Forbidden error when insufficient scopes are present in the auth token.        | 
+| :heavy_check_mark: **MUST** return a 429 Too many requests error when the caller has exceeded throttling limits.        |
```
dkershaw10: There are lots of services that incorrectly return 503 (or some other 50x error) instead of 429.  I wonder if we should make that clearer here?
429 - client sending too many requests
503 - service overloaded (client is within throttling limits)

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+Guidelines](https://github.com/microsoft/api-guidelines/blob/graph/Guidelines.md#123-definition-of-a-breaking-change).
+In general, making all but additive changes to the API contract for existing elements is
+considered breaking. Adding new elements is allowed and not considered a
+breaking change.
+
+\*\* Non-breaking changes:\*\*
+
+
+-   Addition of properties that are nullable or have a default value
+-   Addition of a member to an evolvable enumeration
```
```suggestion
-   Addition of a member, after the sentinel member, to an evolvable enumeration 
```

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+
+
+-   Addition of properties that are nullable or have a default value
+-   Addition of a member to an evolvable enumeration 
+-   Removal, rename, or change to the type of an annotation   
+-   Changes to the order of properties
+-   Changes to the length or format of opaque strings, such as resource IDs
+-   Addition or removal of an annotation OpenType="true" 
+
+\*\* Breaking changes:\*\*
```
dkershaw10, line 369: What about functional changes (where schema is not updated)?
Do we class returning new record types in a heterogeneous collection as breaking or non-breaking?

graph/deprecation.md
```diff
@@ -0,0 +1,47 @@
...
+          <PropertyValue Property = "Description" String="The Outlook tasks API is deprecated and will stop returning data on August 20, 2022. Please use the new To Do API."/>
+          <PropertyValue Property = "RemovalDate" Date="2022-08-20"/>
+        </Record>
+      </Collection>
+    </Annotation>
+    ...
+  </EntityType>
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When the request URL contains a reference to a deprecated model element, the gateway will add a [Deprecation
```
dkershaw10, line 36: Nit:  shouldn't the Sunset header date match the Removal Date?
And in the header example, the Sunset date is **before** the deprecation date!!!

graph/patterns/dictionary.md
```diff
@@ -1,26 +1,57 @@
...
 
 - The data values MUST be related to one another semantically as a collection.
 - The value types MUST be a primitive type or is a **ComplexType**. Mixed primitive types are not allowed.
-- The client MUST define the keys of this type. As opposed to the service defining it in advance.
+- The client MUST define the keys of this type. As opposed to the service defining them in advance.
 
 ### Alternatives to consider
 
 - [Open extensions](https://docs.microsoft.com/en-us/graph/extensibility-open-users) when you want to provide clients the ability to extend Microsoft Graph.
 - [Open types](https://docs.microsoft.com/en-us/aspnet/web-api/overview/odata-support-in-aspnet-web-api/odata-v4/use-open-types-in-odata-v4) when your data is not a collection in nature.
```
dkershaw10: Not sure we should be offering this option.  We're pushing to disallow pure open types.  I think we should remove this option.

graph/patterns/facets.md
```diff
@@ -0,0 +1,111 @@
...
+        <Property Name="video" Type="graph.video" />
+        <Property Name="webDavUrl" Type="Edm.String" />
+     ...
+      </EntityType>
+```
+
+API request to get all items from a personal OneDrive will return a heterogenous collection with different facets populated. In the example below there is a folder, a file and an image in the collection. The image entity has two facets populated: file and image.
+
+```
+https://graph.microsoft.com/beta/me/drive/root/children
```
dkershaw10: We shouldn't reference /beta
In this model, for any facet that is not present would the response show those facet properties as null or does the service omit null properties?

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+
+Below is a short summary of the most often used conventions.
+
+| Requirements                                                                                                         | Example                                                                                                                                                                                                                                                                                                                 |
+|---------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| :no_entry: **MUST NOT** use redundant words in names.   |- **Right:** /places/{id}/**type** and /phones/{id}/**number** <BR> -  **Wrong** /places/{id}/*placeType* and /phones/{id}/**phoneNumber** |
```
mikepizzo: "type" itself is not a great name for a property, as it is overloaded.  perhaps pick a different example?

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+
+-   **scheme and host segments** are always
+    [https://graph.microsoft.com](https://graph.microsoft.com/v1.0/users);
+
+-   **version** can be V1.0 or beta;
+
+-   **category** segment is a logical grouping of APIs into top-level
+    categories;
+
+-   **pathSegment** is the last navigation segment which can address an entity,
```
mikepizzo: Is it clear that this is a path that may contain multiple url segments?

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+### Query Support
+
+Microsoft Graph APIs should support basic query options in conformance with
+OData specifications and [Microsoft REST API
+Guidelines](https://github.com/microsoft/api-guidelines/blob/master/Guidelines.md#7102-error-condition-responses).
+|Requirements|
+|----------------------------------------------------------------------------------------------------|
+| :heavy_check_mark: **MUST** support \$select on resource to enable properties projection |
+| :ballot_box_with_check: **SHOULD** support \$filter with eq, ne operations on properties of entities for collections| 
+| :heavy_check_mark: **MUST** support server-side pagination for collections |
```
mikepizzo: Should we say for *large* collections?  i.e., if I have a collection that will only ever have <10 items I may never page.

mikepizzo: Maybe reference [service-side pagination](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_ServerDrivenPaging) using a [nextLink](http://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#sec_ControlInformationnextLinkodatanextL).

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+    request body or completely in the request url. Graph doesn't support query
+    options present in both places.
+
+-   The parameters in \$query segment should not span multiple workloads. Support for
+    \$query segment right now is limited to properties belonging to the same workload.
+
+
+### Resource Modeling Patterns
+
+You can model complex resources for your APIs using OData Entity Type or Complex Type. The main difference between these types is that Entity type declares a key property to uniquely identify its objects and  Complex Type does not. In Microsoft Graph this key property has "id" as a prescribed name.
```
mikepizzo: Actually, MS Graph does not require the property to be named "id".  If there is a natural name for the property then the workload can use that.  However, for server-created key values that have no semantic meaning, "id" is the recommended name.

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+Since objects of complex types on Graph don’t have unique identifiers, they are not directly addressable via URIs and therefore you must not use Complex Type to model addressable resources. Complex types are better suited to represent composite properties of API entities.
```
mikepizzo: Maybe call out that complex types generally shouldn't be used in collections:

"...you must not use Complex Types to model addressable resources, such as individually addressable resources within a collection."

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+-   Type hierarchy is represented by one abstract base type with a few common
+    properties and one sub-type for each variant [Modelling with Subtypes
+    Pattern](./Modelling%20with%20Subtypes%20Pattern.md)
+
+-   Facets are represented by a single entity type with common properties and
+    one facet property (of complex type) per variant. The facet properties only
+    have a value when the object represents that variant [Modelling with Facets
+    Pattern](./Modelling%20with%20Facets%20Pattern.md)
+
+-   Flat bag of properties is represented by one entity type with all the
```
mikepizzo, line 241: don't we discourage this pattern?

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+\*\* Breaking changes:\*\*
+
+-   Changes to the URL or fundamental request/response associated with a
+    resource
+-   Removal, rename, or change to an incompatible type of a declared property
+-   Removal or rename of APIs or API parameters
+-   Addition of a required request header
+-   Addition of a EnumType members for non-evolvable enumerations
+-   Addition of a Nullable="false" properties to existing types
+-   Addition of a Nullable="false" parameters to existing actions and functions
```
mikepizzo: Technically:

-  Addition of a parameter not marked as [Nullable](http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_Nullable) to existing actions
-  Addition of a parameter not marked as [Optional](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.md#OptionalParameter) to an existing function

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+
+3.  A *feature offering* covering a single use case and *shared* across multiple
+    Microsoft products, i.e. /search, /notifications, /subscriptions.
+
+4.  *Administrative configuration* functions for specific products. i.e. /admin/exchange.
+
+5.  Internal Microsoft requirements for publishing Privileged and Hidden APIs,
+    routing, and load testing, i.e./loadTestEntities.
+
+Effectively top-level categories define a perimeter for the API surface thus a
```
dkershaw10, line 173: Should the same be true for adding nav properties to users and groups?

mikepizzo (changes requested):

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+-   Changes to the URL or fundamental request/response associated with a
+    resource
+-   Removal, rename, or change to an incompatible type of a declared property
+-   Removal or rename of APIs or API parameters
+-   Addition of a required request header
+-   Addition of a EnumType members for non-evolvable enumerations
+-   Addition of a Nullable="false" properties to existing types
+-   Addition of a Nullable="false" parameters to existing actions and functions
+-    Changes to top-level error codes
+-    Introduction of server-side pagination to existing collections
```
mikepizzo, line 381: I understand the motivation behind this, but not sure it's practical. I may support server-side paging on a collection if it contains more than, say, 20 items. If users previously only saw a maximum of 18 items they would never see a nextlink, but asthey added more items over time they might eventually see a nextlink on a collection they had never seen a nextlink on before. Would that be considered a breaking change?  The only way to *not* be breaking would be to *always* return a nextLink with a limited (perhaps zero) number of elements in the initial response..

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+-    Changes to the default order of collection elements
```
mikepizzo: do we guarantee this across workloads today?  will /users *always* return users in the same order, now and in the future?

graph/patterns/dictionary.md
```diff
@@ -1,26 +1,57 @@
...
+ 
+## When to Use this Pattern
+------------------------
 
 Before using a dictionary type in your API definition make sure your scenario fits the following criteria:
 
 - The data values MUST be related to one another semantically as a collection.
 - The value types MUST be a primitive type or is a **ComplexType**. Mixed primitive types are not allowed.
-- The client MUST define the keys of this type. As opposed to the service defining it in advance.
+- The client MUST define the keys of this type. As opposed to the service defining them in advance.
```
dkershaw10, line 43: I wonder if this is clear enough
I wonder if we need another statement to be explicit what this is NOT for.  The "As opposed" statement doesn't go far enough for me

"Services MUST NOT use dictionary types as a way to quickly iterate APIs without strong typing or API governance"

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+
+You must create a new version of your element for any breaking change and name it uniquely. 
+In some cases, the API will have evolved such that there is a new, natural unique name.   In other cases, the original name may still be the most descriptive for the evolved element.  In the latter case, the suffix _v2 must be added to the original name to make it unique.
+The original element is then marked as deprecated using annotations.
+
+Microsoft Graph provides two public endpoints to support API lifecycle:
+1.	API sets on the v1.0 endpoint (https://graph.microsoft.com/v1.0) are in general availability (GA) status.
+2.	API sets on the beta endpoint (https://graph.microsoft.com/beta) are in beta or private preview status.
+
+Microsoft Graph APIs in the GA version guarantee API stability and consistency for its clients. If your API requires a breaking change in GA, then you MUST create new element versions and support deprecated elements for a minimum of 36 months.
```
mikepizzo: Technically, it's 36 months, or 24 months with demonstrated non-usage.

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+-    Changes to top-level error codes
+-    Introduction of server-side pagination to existing collections
+-    Changes to the default order of collection elements
+-    Significant changes to the performance of APIs such as increased latency, rate limits or concurrency.
+
+
+
+### Versioning and Deprecation
+As the market and technology evolves your APIs will require modifications in this case you must avoid breaking changes and add new resources and features incrementally. If that is not possible then you must version elements of your APIs.
+Microsoft Graph allows versioning of elements including entities and properties. The versioning process goes along with deprecation and as soon as you introduce a new element update the previous version needs to follow the deprecation process.
```
mikepizzo: Wording reads a bit awkward.  Maybe update 394-398 with something like:

Microsoft Graph allows versioning of elements including entities and properties. Versioning involves adding a new, uniquely named version of the element and marking the old version as deprecated.

In some cases, there will be a natural new name for the element. In other cases, where the original name is still the most descriptive, the suffix _v2 can be added to the original name to make it unique.

graph/GuidelinesGraph.md
OlgaPodo: Outline the key current and future scenarios for API consumers

graph/GuidelinesGraph.md
OlgaPodo: replace type with displayName wrong place display Name

dkershaw10 (commented):
More comments coming

graph/GuidelinesGraph.md
OlgaPodo: is one or many nav segments

graph/GuidelinesGraph.md
OlgaPodo: Our recommendation for workloads is that all collections should support paging

graph/GuidelinesGraph.md
OlgaPodo: will remove based on the working group recommendation

graph/GuidelinesGraph.md
OlgaPodo: agree

graph/GuidelinesGraph.md
darrelmiller: I don't think we should omit mentioning known patterns.  I do think the pattern document needs to contain clear guidance around its usage.

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,441 @@
...
+Microsoft Graph APIs should support basic query options in conformance with
+OData specifications and [Microsoft REST API
+Guidelines](https://github.com/microsoft/api-guidelines/blob/master/Guidelines.md#7102-error-condition-responses).
+|Requirements|
+|----------------------------------------------------------------------------------------------------|
+| :heavy_check_mark: **MUST** support \$select on resource to enable properties projection |
+| :ballot_box_with_check: **SHOULD** support \$filter with eq, ne operations on properties of entities for collections| 
+| :heavy_check_mark: **MUST** support server-side pagination for collections |
+| :ballot_box_with_check: **SHOULD** support pagination $top, $skip and $count for collections |
+
```
PetRich-MSFT, line 188:
```suggestion
| :ballot_box_with_check: **SHOULD** sorting with \$orderby both ascending and descending on properties of the entities
```

graph/GuidelinesGraph.md
OlgaPodo: will provide detailed description in the complex type pattern

graph/GuidelinesGraph.md
darrelmiller: We can add this nuance to a document that describes the Entity Type

graph/GuidelinesGraph.md
OlgaPodo: I think in this context API behavior includes standard HTTP methods, actions, functions and various exceptions.  in my mind "custom methods" have a connotation of non-standard HTTP methods?

graph/GuidelinesGraph.md
OlgaPodo: in the Resource Modeling Patterns section line #207 we cover addressable resources. I can add a reference to the https://github.com/microsoft/api-guidelines/blob/graph/Guidelines.md#93-collection-url-patterns.

graph/patterns/facets.md
OlgaPodo: the service omits null properties

graph/GuidelinesGraph.md
OlgaPodo: structured

graph/GuidelinesGraph.md
OlgaPodo: It would be the second navigation segment after the categories. correct?

graph/patterns/dictionary.md
```diff
@@ -1,26 +1,57 @@
...
 
 ### Alternatives to consider
```
darrelmiller:
```suggestion
### Alternatives
```

graph/patterns/dictionary.md
darrelmiller:
```suggestion
```

darrelmiller: Removed it.

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,443 @@
...
+have with your service. As an API provider, how you manage the overall request /
+response pattern should be one of the first implementation decisions you make.
+APIs SHOULD use resource-based designs with standard HTTP methods rather than operation resources if possible.
+ Operation resources are either functions or actions. According to [ODATA standards]( http://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part3-csdl/odata-v4.0-errata03-os-part3-csdl-complete.html#_The_edm:Function_Element_2) a function represents an operation which returns a single instance or collection of instances of any type and doesn’t have an observable side effect. An action may have side effects and may return a result represented as a single entity or collection of any type.
+
+|  Microsoft Graph rules for modeling behavior                                                                                          |
+|-----------------------------------------------------------------------------------------------------------------|
+| :heavy_check_mark: **MUST** use POST to create new entities in insertable entity sets or collections |
```
darrelmiller:
```suggestion
| :heavy_check_mark: **MUST** use POST to create new entities in insertable entity sets or collections. This approach requires the server to produce system generated  identities. | 
```

graph/GuidelinesGraph.md
```diff
@@ -0,0 +1,443 @@
...
+
+|  Microsoft Graph rules for modeling behavior                                                                                          |
+|-----------------------------------------------------------------------------------------------------------------|
+| :heavy_check_mark: **MUST** use POST to create new entities in insertable entity sets or collections | 
+| :heavy_check_mark: **MUST** use PATCH to edit updatable resources                                                 | 
+| :heavy_check_mark: **MUST** use DELETE to delete deletable resources                                              | 
+| :heavy_check_mark: **MUST** use GET for listing and reading resources. | 
+| :warning: **SHOULD NOT** use PUT for updating resources.                                       | 
+| :ballot_box_with_check: **SHOULD** avoid using multiple round trips to complete a single logical action.       |
```
darrelmiller:
```suggestion
| :ballot_box_with_check: **SHOULD** avoid using multiple round trips to complete a single logical operation.       | 
```

graph/GuidelinesGraph.md
OlgaPodo: Updated

graph/GuidelinesGraph.md
OlgaPodo: Changed to operation resources

graph/GuidelinesGraph.md
OlgaPodo: according to the MS REST API guidelines https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#123-definition-of-a-breaking-change changes in behavior are breaking changes. But we allow teams doing it consistently. It requires SDK versioning.

graph/patterns/dictionary.md
darrelmiller: Let's see if people try to abuse them.  I'd rather not starting looking for all thing things that folks might do wrong and enumerating them.

darrelmiller (approved):

PR #339, opened 2022-03-28
Reorder the guidelines so the error contract is clearer

tg-msft:
There's a gap between the two rules talking about what's part of the error contract and what isn't.  I think this makes it easier for teams to understand.

### DIFF
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -302,8 +302,6 @@ There are 2 kinds of errors:
 
 :white_check_mark: **DO** ensure that the top-level error's `code` value is identical to the `x-ms-error-code` header's value.
 
-:white_check_mark: **DO** document the service's error code strings; they are part of the API contract.
-
 :white_check_mark: **DO** provide a response body with the following structure:
 
 **ErrorResponse** : Object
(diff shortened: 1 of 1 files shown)


azure/Guidelines.md
```diff
@@ -302,8 +302,6 @@ There are 2 kinds of errors:
 
 :white_check_mark: **DO** ensure that the top-level error's `code` value is identical to the `x-ms-error-code` header's value.
 
-:white_check_mark: **DO** document the service's error code strings; they are part of the API contract.
-
```
mikekistler, line 306: I think this rule belongs where it is since the six rules immediately above it all concern various aspects of the error code.

My suggestion is to move the rule at line 346 here to pull all the error code related guidelines together.

tg-msft: I pushed it down because it felt weird to talk about "all the other fields" before we'd defined those fields.  Let's take a vote the next time we have a group.

tg-msft: Discussed as a group and we'll keep it where it is, but move the blurb about error code grouping up near the other error code discussion.

mikekistler (approved):
Looks good! 👍

markweitzel (approved):

PR #340, opened 2022-03-29
Add guidelines for durations

heaths:
Discussed in the REST API leadership meeting.

### DIFF
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -375,7 +375,7 @@ Services, and the clients that access them, may be written in multiple languages
 
 :white_check_mark: **DO** use [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339) for date/time.
 
-:white_check_mark: **DO** use a fixed time interval when using durations and include the time unit in the name e.g., `backupTimeInMinutes` or `ttlSeconds`. The time unit should also be included in the property description to support localization e.g., "How often to back up expressed in minutes," or "The lifetime of the resource expressed in seconds."
+:white_check_mark: **DO** use a fixed time interval to express durations and include the time unit in the property name e.g., `backupTimeInMinutes` or `ttlSeconds`. The time unit should also be included in the property description to support localization e.g., "How often to back up expressed in minutes," or "The lifetime of the resource expressed in seconds."
 
 :heavy_check_mark: **YOU MAY** use [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339) durations only when users must be able to specify a time interval that may change from month to month or year to year e.g., "P1Y" represents 366 days on a leap year.
 
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -375,9 +375,9 @@ Services, and the clients that access them, may be written in multiple languages
 
 :white_check_mark: **DO** use [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339) for date/time.
 
-:white_check_mark: **DO** use a fixed time interval to express durations and include the time unit in the property name e.g., `backupTimeInMinutes` or `ttlSeconds`.
+:white_check_mark: **DO** use a fixed time interval to express durations e.g., milliseconds, seconds, minutes, days, etc., and include the time unit in the property name e.g., `backupTimeInMinutes` or `ttlSeconds`.
 
-:heavy_check_mark: **YOU MAY** use [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339) durations only when users must be able to specify a time interval that may change from month to month or year to year e.g., "P1Y" represents 366 days on a leap year.
+:heavy_check_mark: **YOU MAY** use [RFC3339 durations](https://wikipedia.org/wiki/ISO_8601#Durations) only when users must be able to specify a time interval that may change from month to month or year to year e.g., "P3M" represents 3 months no matter how many days between the start and end, or "P1Y" represents 366 days on a leap year.
 
 :white_check_mark: **DO** use [RFC4122](https://datatracker.ietf.org/doc/html/rfc4122) for UUIDs.
 
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -377,7 +377,7 @@ Services, and the clients that access them, may be written in multiple languages
 
 :white_check_mark: **DO** use a fixed time interval to express durations e.g., milliseconds, seconds, minutes, days, etc., and include the time unit in the property name e.g., `backupTimeInMinutes` or `ttlSeconds`.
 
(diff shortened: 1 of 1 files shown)


azure/Guidelines.md
```diff
@@ -374,6 +375,10 @@ Services, and the clients that access them, may be written in multiple languages
 
 :white_check_mark: **DO** use [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339) for date/time.
 
+:white_check_mark: **DO** use a fixed time interval when using durations e.g., `ttlSeconds`. The unit of time should be part of the name, but must at least be documented in the property description.
```
weidongxu-microsoft: Do we have some preference on the units? E.g. preferable second and millisecond?

In extreme case, what if service use [tick](https://docs.microsoft.com/en-us/dotnet/api/system.timespan.ticks?view=net-6.0)?

markweitzel:
```suggestion
:white_check_mark: **DO** use a fixed time interval when using durations e.g., `ttlSeconds`. The unit of time should be part of the name, e.g. `backupTimeInMinutes` 

✅ **DO** fully document the property and the interval in the description.
```

Any objection to being more direct on the guidance around documentation?
I might also add an example of the operation name.

heaths: I could give a couple examples like I "e.g., `ttlSeconds` or `backupTimeInMinutes`. Should we just be explicit then that they should *always* include the unit?

@weidongxu-microsoft no, I don't feel we should be explicit about the time unit. It will vary from use to use, and if we made everything milliseconds, many models may need to use a 64-bit integer to at least parse the number where a 32-bit integer would do if in seconds, minutes, hours, etc.

weidongxu-microsoft: Sure. Ask because I've just get caught by spring cloud service use int for cpu/memoryGB, then they decided to support 0.5 cpu and 0.5 GB :-(

heaths:
@JeffreyRichter could you also take a look?

azure/Guidelines.md
```diff
@@ -374,6 +375,10 @@ Services, and the clients that access them, may be written in multiple languages
 
 :white_check_mark: **DO** use [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339) for date/time.
 
+:white_check_mark: **DO** use a fixed time interval when using durations and include the time unit in the name e.g., `backupTimeInMinutes` or `ttlSeconds`. The time unit should also be included in the property description to support localization e.g., "How often to back up expressed in minutes," or "The lifetime of the resource expressed in seconds."
```
```suggestion
:white_check_mark: **DO** use a fixed time interval to express durations and include the time unit in the property name e.g., `backupTimeInMinutes` or `ttlSeconds`. The time unit should also be included in the property description to support localization e.g., "How often to back up expressed in minutes," or "The lifetime of the resource expressed in seconds."
```

JeffreyRichter: I don't know what you mean by "support localization"; please explain?

JeffreyRichter (approved):

azure/Guidelines.md
heaths: The schema - models, properties, etc. - won't be localized, but the descriptions appear to be, at least when I look at https://docs.microsoft.com/de-de/rest/api/keyvault/keys/create-key/create-key. So I figure the unit of measure should also be in the description so it's localized, i.e. localizable.

JeffreyRichter: I don't know how this is done (automated or manual) and if done manually, I don't know if this happens for all services. Regardless, we never mention anything about localization in any of our guidelines and so I'd prefer we not mention it here as mentioning it introduced questions and uncertainty.

azure/Guidelines.md
```diff
@@ -374,6 +375,10 @@ Services, and the clients that access them, may be written in multiple languages
 
 :white_check_mark: **DO** use [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339) for date/time.
 
+:white_check_mark: **DO** use a fixed time interval to express durations and include the time unit in the property name e.g., `backupTimeInMinutes` or `ttlSeconds`.
+
+:heavy_check_mark: **YOU MAY** use [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339) durations only when users must be able to specify a time interval that may change from month to month or year to year e.g., "P1Y" represents 366 days on a leap year.
```
tg-msft: Is that the right spec or did we mean ISO 8601?  The `Pwhatever` duration format is only discussed in the ABNF in the appendix and might not be as helpful as section 4.4.3 of ISO 8601?

heaths: RFC 3339 supersedes ISO 8601, and the latter isn't accessible (paywalled) anywhere I can find outside of Wikipedia. So I don't think we can link to that, though I agree RFC 3339's ABNF isn't great. Should I link to Wikipedia as well / in lieu of?

tg-msft: Yeah, I'd suggest linking to Wikipedia as well.  I don't think most internal non-HTTP-experts would be successful with just RFC 3339.

JeffreyRichter:
I think we should not use the word "durations" here as this is what causes the problems./misunderstandings. Languages that have a duration type (like c# TimeSpan) cannot represent this correctly. I propose the term "time interval" instead and the number for years, months, and daysMUST be preserved and round-trippable.

-- Jeffrey Richter
________________________________

heaths:
I use "durations" because that's what they are called: ISO 8601 durations, or RFC 3339 durations. I do mention "fix time interval" initially, though.

azure/Guidelines.md
heaths: Okay, but I'm still going to reference RFC 3339, which we do in other places, like just right about this. I can link to wikipedia for more information. Maybe I just won't make RFC3339 a link.

weidongxu-microsoft (approved):

heaths:
Talking with @tg-msft I will try to reword this. Looking through ISO 8601 / RFC 3339, it does classify "durations" as a type of "time interval" - the latter also represented by a combination of start and/or end ISO 8601 date-times and/or durations separated by a `/`.

heaths:
That said, @JeffreyRichter, I don't see how you'd associate the word "duration" with `TimeSpan` which share no words in common, but "time interval" which does share the word "time" in common would be less confusing. I agree it's important we state it must be round-trippable, but I think it's worth mentioning that built-in types like .NET's `TimeSpan` or Java's `Duration` or `Period` do not round-trip correctly and should not be used. If we don't, it leaves a lot of room for teams to get entrenched into implementing APIs that use those and find out late it doesn't work. I think guidelines should dissuade from using incompatible types.

JeffreyRichter:
A TimeSpan represents a duration of time. It doesn't feel to me that it represents a time interval. Also, let's not forget about other  languages; Go, for example, does has a "Duration" type and it would be problematic to convert an RFC 3339 time interval into a Go Duration.

PR #341, opened 2022-03-31
Update Codeowners

weidongxu-microsoft:
Find out we actually already had `@microsoft/azure-api-stewardship-board` github team.

Please give either `@microsoft/azure-api-stewardship-board` or `@Azure/api-stewardship-board` write permission to the repository, for code owner assignment to team.

> The people you choose as code owners must have read permissions for the repository. When the code owner is a team, that team must be visible and it must have write permissions, even if all the individual members of the team already have write permissions directly, through organization membership, or through another team membership.

Or instead we can add individuals as code owner.


weidongxu-microsoft:
@markweitzel @czubair

PR #347, opened 2022-04-29
Add Alternate Key Pattern

XVincentX:
This PR adds a brief description of the Alternate Key Pattern discussed in the API Council.

It is not particularly long since the pattern per se is already extensively documented in the oData specification. This document could have just been "Use what we have in oData" :trollface:

### DIFF
--- a/graph/patterns/alternate-key.md
+++ b/graph/patterns/alternate-key.md
@@ -19,7 +19,7 @@ While it is still possible to use the oData filter, such as
 ## Solution
 --------
 
-oData offers resource addressing via an alternate key using the same parentheses-style convention as for the canonical key, with one difference: single-part alternate keys MUST specify the key property name to unambiguously determine the alternate key.
+oData offers resource addressing via an alternate key using the same parentheses-style convention as for the canonical key, with one difference: single-part alternate keys MUST specify the key attribute name to unambiguously determine the alternate key.
 
 https://graph.microsoft.com/v1.0/users(0) - Retrieves the employee with ID = 0
 https://graph.microsoft.com/v1.0/users(email='bob@contoso.com') Retrieves the employee with the email matching `bob@contoso.com`
@@ -34,7 +34,7 @@ In such case, we **strongly** advice to throw an exception and encourage the use
 ## Example
 -------
 
-The same user identified via the alternate key SSN, the canonical (primary) key ID using the non-canonical long form with specified key property name, and the canonical short form without key property name
+The same user identified via the alternate key SSN, the canonical (primary) key ID using the non-canonical long form with specified key attribute name, and the canonical short form without key attribute name
 
 https://graph.microsoft.com/v1.0/users/1a89ade6-9f59-4fea-a139-23f84e3aef66
 
--- a/graph/patterns/alternate-key.md
+++ b/graph/patterns/alternate-key.md
@@ -29,12 +29,7 @@ https://graph.microsoft.com/v1.0/users(email='bob@contoso.com') Retrieves the em
 
 ---
 
-This pattern works and makes sense when the alternate key is good enough to identify a single resource and provides an useful alternative to the client; while it does not work if the resultset has more than one element.
-
-In such case, the workload has two choices:
-
-1. If the workload defines a list of alternate keys for a resource and the requested key is not there, the workload SHOULD return `422`
-2. If the workload does not have a list of alternate keys, it can still query the data source, but if the result yields more than a record, it SHOULD return `422` and encourage the user to use `$filter` rather than returning the first result of the query
+This pattern works and makes sense when the alternate key is good enough to identify a single resource and provides an useful alternative to the client; while it does not work if the resultset has more than one element. In such case, the workload SHOULD return `400`
 
 ## Example
 
--- a/graph/patterns/alternate-key.md
+++ b/graph/patterns/alternate-key.md
@@ -31,7 +31,7 @@ https://graph.microsoft.com/v1.0/users(email='bob@contoso.com') Retrieves the em
 
 This pattern works and makes sense when the alternate key is good enough to identify a single resource and provides an useful alternative to the client; while it does not work if the resultset has more than one element.
 
-In such case, we **strongly** advice to throw an exception and encourage the user to use `$filter` rather than returning the first result of the query
+In such case, the system SHOULD throw an exception and encourage the user to use `$filter` rather than returning the first result of the query
 
 ## Example
 
--- a/graph/patterns/alternate-key.md
+++ b/graph/patterns/alternate-key.md
@@ -2,14 +2,14 @@
 
 Microsoft Graph API Design Pattern
 
-*The Alternate Key Pattern provides the ability to query for a single, specific entity identifiable through an alternative attribute (called key) that is not its unique identifier*
+*The Alternate Key Pattern provides the ability to query for a single, specific resource identifiable through an alternative attribute (called key) that is not its unique identifier*
 
 ## Problem
 --------
 
-All entities in our system are identified by an [UUID (Universally Unique Identifier)](https://en.wikipedia.org/wiki/Universally_unique_identifier) - which guarantees uniqueness. Often though, that same entity can also be uniquely identified by an alternative, more convenient attribute.
+All entities in our system are identified by an [UUID (Universally Unique Identifier)](https://en.wikipedia.org/wiki/Universally_unique_identifier) - which guarantees uniqueness. Often though, that same resource can also be uniquely identified by an alternative, more convenient attribute.
 
-Take a look at the `user` entity: while the `id` remains a perfectly valid way to get the entity details, the `mail` address is also an unique attribute that could be used to identify it.
+Take a look at the `user` resource: while the `id` remains a perfectly valid way to get the resource details, the `mail` address is also an unique attribute that could be used to identify it.
 
 While it is still possible to use the oData filter, such as
 
@@ -19,7 +19,7 @@ While it is still possible to use the oData filter, such as
 ## Solution
 --------
 
-oData offers entity addressing via an alternate key using the same parentheses-style convention as for the canonical key, with one difference: single-part alternate keys MUST specify the key property name to unambiguously determine the alternate key.
+oData offers resource addressing via an alternate key using the same parentheses-style convention as for the canonical key, with one difference: single-part alternate keys MUST specify the key property name to unambiguously determine the alternate key.
 
 https://graph.microsoft.com/v1.0/users(0) - Retrieves the employee with ID = 0
 https://graph.microsoft.com/v1.0/users(email='bob@contoso.com') Retrieves the employee with the email matching `bob@contoso.com`
@@ -27,7 +27,7 @@ https://graph.microsoft.com/v1.0/users(email='bob@contoso.com') Retrieves the em
 ## When to Use this Pattern
 ------------------------
 
-This pattern works and makes sense when the alternate key is good enough to identify a single entity and provides an useful alternative to the client; while it does not work if the resultset has more than one element.
+This pattern works and makes sense when the alternate key is good enough to identify a single resource and provides an useful alternative to the client; while it does not work if the resultset has more than one element.
 
 In such case, we **strongly** advice to throw an exception and encourage the user to use `$filter` rather than returning the first result of the query 
 
--- a/graph/patterns/alternate-key.md
+++ b/graph/patterns/alternate-key.md
@@ -21,8 +21,8 @@ While it is still possible to use the oData filter, such as
 
 oData offers entity addressing via an alternate key using the same parentheses-style convention as for the canonical key, with one difference: single-part alternate keys MUST specify the key property name to unambiguously determine the alternate key.
 
-http://host/service/Employees(0) - Retrieves the employee with ID = 0
-http://host/service/Employees(email='hello@microsoft.com') Retrieves the employee with the email matching `hello@microsoft.com`
+https://graph.microsoft.com/v1.0/users(0) - Retrieves the employee with ID = 0
+https://graph.microsoft.com/v1.0/users(email='hello@microsoft.com') Retrieves the employee with the email matching `hello@microsoft.com`
 
 ## When to Use this Pattern
 ------------------------
@@ -36,11 +36,11 @@ In such case, we **strongly** advice to throw an exception and encourage the use
 
 The same user identified via the alternate key SSN, the canonical (primary) key ID using the non-canonical long form with specified key property name, and the canonical short form without key property name
 
-http://host/service/Employees/1a89ade6-9f59-4fea-a139-23f84e3aef66
+https://graph.microsoft.com/v1.0/users/1a89ade6-9f59-4fea-a139-23f84e3aef66
 
-http://host/service/Employees(ssn='123-45-6789')
+https://graph.microsoft.com/v1.0/users(ssn='123-45-6789')
 
-http://host/service/Employees(email='hello@microsoft.com')
+https://graph.microsoft.com/v1.0/users(email='hello@microsoft.com')
 
 
 
--- a/graph/patterns/alternate-key.md
+++ b/graph/patterns/alternate-key.md
@@ -2,15 +2,15 @@
 
 Microsoft Graph API Design Pattern
 
-_The Alternate Key Pattern provides the ability to query for a single, specific resource identifiable through an alternative attribute (called key) that is not its unique identifier_
+_The Alternate Key Pattern provides the ability to query for a single, specific resource identifiable through an alternative property (called key) that is not its unique identifier_
 
 ## Problem
 
 ---
 
-The resources exposed in Graph are identified by an [UUID (Universally Unique Identifier)](https://en.wikipedia.org/wiki/Universally_unique_identifier) - which guarantees uniqueness inside the same resource type. Often though, that same resource can also be uniquely identified by an alternative, more convenient attribute that provides a better developer experience.
+The resources exposed in Graph are identified by an [UUID (Universally Unique Identifier)](https://en.wikipedia.org/wiki/Universally_unique_identifier) - which guarantees uniqueness inside the same resource type. Often though, that same resource can also be uniquely identified by an alternative, more convenient property that provides a better developer experience.
 
-Take a look at the `user` resource: while the `id` remains a perfectly valid way to get the resource details, the `mail` address is also an unique attribute that could be used to identify it.
+Take a look at the `user` resource: while the `id` remains a perfectly valid way to get the resource details, the `mail` address is also an unique property that could be used to identify it.
 
 While it is still possible to use the oData filter, such as
 
@@ -20,7 +20,7 @@ While it is still possible to use the oData filter, such as
 
 ---
 
-oData offers resource addressing via an alternate key using the same parentheses-style convention as for the canonical key, with one difference: single-part alternate keys MUST specify the key attribute name to unambiguously determine the alternate key.
+oData offers resource addressing via an alternate key using the same parentheses-style convention as for the canonical key, with one difference: single-part alternate keys MUST specify the key property name to unambiguously determine the alternate key.
 
 https://graph.microsoft.com/v1.0/users(0) - Retrieves the employee with ID = 0
 https://graph.microsoft.com/v1.0/users(email='bob@contoso.com') Retrieves the employee with the email matching `bob@contoso.com`
@@ -37,7 +37,7 @@ In such case, the system SHOULD throw an exception and encourage the user to use
 
 ---
 
-The same user identified via the alternate key SSN, the canonical (primary) key ID using the non-canonical long form with specified key attribute name, and the canonical short form without key attribute name
+The same user identified via the alternate key SSN, the canonical (primary) key ID using the non-canonical long form with specified key property name, and the canonical short form without key property name
 
 https://graph.microsoft.com/v1.0/users/1a89ade6-9f59-4fea-a139-23f84e3aef66
 
--- a/graph/patterns/alternate-key.md
+++ b/graph/patterns/alternate-key.md
@@ -9,7 +9,7 @@ Microsoft Graph API Design Pattern
 
 All entities in our system are identified by an UUID - which guarantees uniqueness. Often though, that same entity can also be uniquely identified by an alternative, more convenient attribute.
 
-Take a look at the `user` entity: while the UUID remains a perfectly valid way to get the entity details, the `mail` address is also an unique attribute that could be used to identify it.
+Take a look at the `user` entity: while the `id` remains a perfectly valid way to get the entity details, the `mail` address is also an unique attribute that could be used to identify it.
 
 While it is still possible to use the oData filter, such as
 
--- a/graph/patterns/alternate-key.md
+++ b/graph/patterns/alternate-key.md
@@ -9,7 +9,7 @@ Microsoft Graph API Design Pattern
 
 All entities in our system are identified by an UUID - which guarantees uniqueness. Often though, that same entity can also be uniquely identified by an alternative, more convenient attribute.
 
-Take a look at the `user` entity: while the UUID remains a perfectly valid way to get the entity details, the `email` address is also an unique attribute that could be used to identify it.
+Take a look at the `user` entity: while the UUID remains a perfectly valid way to get the entity details, the `mail` address is also an unique attribute that could be used to identify it.
 
 While it is still possible to use the oData filter, such as
 
--- a/graph/patterns/alternate-key.md
+++ b/graph/patterns/alternate-key.md
@@ -14,7 +14,7 @@ Take a look at the `user` resource: while the `id` remains a perfectly valid way
 
 While it is still possible to use the oData filter, such as
 
-`https://graph.microsoft.com/v1.0/users?filter=mail eq 'bob@contoso.com'`, the returned result is wrapped in an array that needs to be unpacked.
+`https://graph.microsoft.com/v1.0/users?$filter=mail eq 'bob@contoso.com'`, the returned result is wrapped in an array that needs to be unpacked.
 
 ## Solution
 
--- a/graph/patterns/alternate-key.md
+++ b/graph/patterns/alternate-key.md
@@ -2,13 +2,13 @@
 
 Microsoft Graph API Design Pattern
 
-_The Alternate Key Pattern provides the ability to query for a single, specific resource identifiable through an alternative property (called key) that is not its unique identifier_
+_The Alternate Key Pattern provides the ability to query for a single, specific resource identifiable through an alternative property (called key) that is not its primary key_
 
 ## Problem
 
 ---
 
-The resources exposed in Graph are identified by an [UUID (Universally Unique Identifier)](https://en.wikipedia.org/wiki/Universally_unique_identifier) - which guarantees uniqueness inside the same resource type. Often though, that same resource can also be uniquely identified by an alternative, more convenient property that provides a better developer experience.
+The resources exposed in Graph are identified through a Primary Key - which guarantees uniqueness inside the same resource type. Often though, that same resource can also be uniquely identified by an alternative, more convenient property (or set of properties) that provides a better developer experience.
 
 Take a look at the `user` resource: while the `id` remains a perfectly valid way to get the resource details, the `mail` address is also an unique property that could be used to identify it.
 
@@ -102,7 +102,7 @@ GET https://graph.microsoft.com/v1.0/users/?$filter=(ssn eq '123-45-6789')
 }
 ```
 
-2. Get a specific resource through the unique identifier, and then through the two alternate keys:
+2. Get a specific resource through its primary key, and then through the two alternate keys:
 
 ```http
 GET https://graph.microsoft.com/v1.0/users/1a89ade6-9f59-4fea-a139-23f84e3aef66
--- a/graph/patterns/alternate-key.md
+++ b/graph/patterns/alternate-key.md
@@ -7,7 +7,7 @@ Microsoft Graph API Design Pattern
 ## Problem
 --------
 
-All entities in our system are identified by an [UUID (Universally Unique Identifier)](https://en.wikipedia.org/wiki/Universally_unique_identifier) - which guarantees uniqueness. Often though, that same resource can also be uniquely identified by an alternative, more convenient attribute.
+All entities in our system are identified by an [UUID (Universally Unique Identifier)](https://en.wikipedia.org/wiki/Universally_unique_identifier) - which guarantees uniqueness. Often though, that same resource can also be uniquely identified by an alternative, more convenient attribute that provides a better developer experience.
 
 Take a look at the `user` resource: while the `id` remains a perfectly valid way to get the resource details, the `mail` address is also an unique attribute that could be used to identify it.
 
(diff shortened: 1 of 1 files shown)


graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,48 @@
+# Alternate Key Pattern
```
dkershaw10, line 1: I have a general editorial comment (maybe more for @OlgaPodo). 
We should use consistent language across these topics.

- resource vs entity?
- property vs attribute?
- Should we be talking about/mentioning OData?

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,48 @@
...
+
+Microsoft Graph API Design Pattern
+
+*The Alternate Key Pattern provides the ability to query for a single, specific entity identifiable through an alternative attribute (called key) that is not its unique identifier*
```
dkershaw10: Rather than "unique identifier", I think you should say "primary key".

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Problem
+--------
+
+All entities in our system are identified by an UUID - which guarantees uniqueness. Often though, that same entity can also be uniquely identified by an alternative, more convenient attribute.
```
dkershaw10: UUID???
If we do use acronyms, we need to define them.

dkershaw10: more convenient attribute
->
more convenient attribute that provides a better developer experience.

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,48 @@
...
+
+Take a look at the `user` entity: while the UUID remains a perfectly valid way to get the entity details, the `email` address is also an unique attribute that could be used to identify it.
```
dkershaw10: If we are using the Microsoft Graph `user` resource type as an example, we should use its property names:
UUID -> `id`
`email` -> `mail`  (this one is unfortunate, as this property really should have been called email :( )

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,48 @@
...
+
+While it is still possible to use the oData filter, such as
+
+`serviceRoot/Users?$filter=email eq 'hello@microsoft.com'`, the returned result is wrapped in an array that needs to be unpacked.
```
dkershaw10: `GET https://graph.microsoft.com/v1.0/users?filter=mail eq 'bob@contoso.com'`
@XVincentX - for future reference - example domains/addresses etc should use the fictional "contoso.com" company.  There are other fictional companies, but start with this one.

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,48 @@
...
+
+
+## Solution
+--------
+
+oData offers entity addressing via an alternate key using the same parentheses-style convention as for the canonical key, with one difference: single-part alternate keys MUST specify the key property name to unambiguously determine the alternate key.
+
+http://host/service/Employees(0) - Retrieves the employee with ID = 0
```
dkershaw10: Rather than using the OData spec examples, let's use real Microsoft Graph examples - and let's stick with `user`.

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,48 @@
...
+http://host/service/Employees(email='hello@microsoft.com') Retrieves the employee with the email matching `hello@microsoft.com`
+
+## When to Use this Pattern
+------------------------
+
+This pattern works when the alternate key is good enough to identify a single entity that is properly namespaced; while it does not work if the resultset has more than one element.
```
dkershaw10: What do you mean by "that is properly namespaced"? I would remove it

I wonder whether we should also say that this is a good pattern to adopt when customers/developers ask for alternative keys to make their lives easier.

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,48 @@
...
+
+In such case, we **strongly** advice to throw an exception and encourage the user to use `$filter` rather than returning the first result of the query 
+
+## Example
+-------
+
+The same user identified via the alternate key SSN, the canonical (primary) key ID using the non-canonical long form with specified key property name, and the canonical short form without key property name
+
+http://host/service/Employees/1a89ade6-9f59-4fea-a139-23f84e3aef66
```
dkershaw10: Again it would be good to use "real" Graph examples
`GET https://graph.microsoft.com/v1.0/users/1a89ade6-9f59-4fea-a139-23f84e3aef66`  <- key-as-segment
`GET https://graph.microsoft.com/v1.0/users('1a89ade6-9f59-4fea-a139-23f84e3aef66')`  <- canonical format
`GET https://graph.microsoft.com/v1.0/users(mail='bob@contoso.com')` <- alternate key

graph/patterns/alternate-key.md
OlgaPodo: IMO we should use resource as it's more generic than an entity and consistent with JSON representation which is an ultimate target of our design. We also should stick with properties as structural elements of a resource for consistency with other public documentation https://docs.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-1.0#properties ( i see that it's not quite consistent with JSON spec) and leave attributes for metadata description.
I don't think that we need to mention OData in this context. In some patterns Type hierarchy, it's unavoidable.

graph/patterns/alternate-key.md
XVincentX: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request

```suggestion
-All entities in our system are identified by an UUID - which guarantees uniqueness. Often though, that same entity can also be uniquely identified by an alternative, more convenient attribute.
+All entities in our system are identified by an UUID - which guarantees uniqueness. Often though, that same entity can also be uniquely identified by an alternative, more convenient attribute that provides a better developer experience.
```

Welcome to GitHub @dkershaw10 :)

XVincentX: I will expand it, I though [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier) was common enough to avoid the expansion

graph/patterns/alternate-key.md
XVincentX: The reason I did not call it that way is that primary key conveys the idea that there is a database backing the system in question - which might not be the case: I might have an unique identifier for a 100% in memory system, or something else.

What I am trying to convey here is more the generic concept of a resource that is uniquely identified. How - up to the developer/specific workload. Am I making sense?

graph/patterns/alternate-key.md
XVincentX: Good point, done in 9f54298

graph/patterns/alternate-key.md
XVincentX: Done in 9f54298

graph/patterns/alternate-key.md
XVincentX: I have probably chosen an unhappy wording for this.

What I was trying to communicate is that you need to be in a context that is narrow enough where the alternate key returns a single entity - I do see how saying "namespaced" makes the things confusing. I have reworded in 8f4966c, let me know if that is better.

graph/patterns/alternate-key.md
XVincentX: * It should be `resource`. Wherever I wrote entity, it was a blunder. a3966c7
* It should be `attribute`. I will change this to `properties` if you want to keep consistency, but the definition is the documentation page is inaccurate
* That is an interesting question. Graph is very heavily reliant (unfortunately) on oData. I would see the case that we have this pattern here and then the workload developers might be like "but that is what oData says anyway, why specify it again here?". I think that saying in this document "hey what we are saying is that we embrace oData" makes it a little bit redundant, but I would say less confusing.

I am open to suggestions though!

OlgaPodo: @XVincentX In the context of Microsoft internal and external documentation when you refer to a resource,  not JSON object, using attribute will be very confusing because Property is used in documentation, DSL/CSDL, and MS REST API guidelines(https://github.com/microsoft/api-guidelines/blob/graph/Guidelines.md). We are creating these patterns for the MS ecosystem and developers so i think it should be consistent.

graph/patterns/alternate-key.md
```diff
@@ -2,47 +2,45 @@
...
-Take a look at the `user` entity: while the UUID remains a perfectly valid way to get the entity details, the `email` address is also an unique attribute that could be used to identify it.
+All entities in our system are identified by an [UUID (Universally Unique Identifier)](https://en.wikipedia.org/wiki/Universally_unique_identifier) - which guarantees uniqueness. Often though, that same resource can also be uniquely identified by an alternative, more convenient attribute that provides a better developer experience.
 
-While it is still possible to use the oData filter, such as
+Take a look at the `user` resource: while the `id` remains a perfectly valid way to get the resource details, the `mail` address is also an unique attribute that could be used to identify it.
 
-`serviceRoot/Users?$filter=email eq 'hello@microsoft.com'`, the returned result is wrapped in an array that needs to be unpacked.
+While it is still possible to use the oData filter, such as
 
+`https://graph.microsoft.com/v1.0/users?filter=mail eq 'bob@contoso.com'`, the returned result is wrapped in an array that needs to be unpacked.
```
OlgaPodo: missing $ sign before filter. I think that it makes sense to elaborate on unpacking in the Issues and Consideration section of the pattern

graph/patterns/alternate-key.md
```diff
@@ -2,47 +2,45 @@
...
 
+---
 
+The same user identified via the alternate key SSN, the canonical (primary) key ID using the non-canonical long form with specified key attribute name, and the canonical short form without key attribute name
 
+https://graph.microsoft.com/v1.0/users/1a89ade6-9f59-4fea-a139-23f84e3aef66
 
+https://graph.microsoft.com/v1.0/users(ssn='123-45-6789')
 
+https://graph.microsoft.com/v1.0/users(email='bob@contoso.com')
```
OlgaPodo: Could you provide examples with responses consistent with other patterns like https://github.com/microsoft/api-guidelines/blob/graph/graph/patterns/dictionary.md?

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,46 @@
...
+
+Microsoft Graph API Design Pattern
+
+_The Alternate Key Pattern provides the ability to query for a single, specific resource identifiable through an alternative attribute (called key) that is not its unique identifier_
+
+## Problem
+
+---
+
+All entities in our system are identified by an [UUID (Universally Unique Identifier)](https://en.wikipedia.org/wiki/Universally_unique_identifier) - which guarantees uniqueness. Often though, that same resource can also be uniquely identified by an alternative, more convenient attribute that provides a better developer experience.
```
OlgaPodo: IMO this statement reads like there is a single UUID for all entities while there is no uniqueness guaranteed outside of an entity type.

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,46 @@
...
+https://graph.microsoft.com/v1.0/users(0) - Retrieves the employee with ID = 0
+https://graph.microsoft.com/v1.0/users(email='bob@contoso.com') Retrieves the employee with the email matching `bob@contoso.com`
+
+## When to Use this Pattern
+
+---
+
+This pattern works and makes sense when the alternate key is good enough to identify a single resource and provides an useful alternative to the client; while it does not work if the resultset has more than one element.
+
+In such case, we **strongly** advice to throw an exception and encourage the user to use `$filter` rather than returning the first result of the query
```
OlgaPodo: We are using rfc2119 to be consistent with MUST, SHOULD and etc. Could you please change strongly accordingly?

XVincentX: Great point. Done in 67610b2

graph/patterns/alternate-key.md
XVincentX: Added the sign.

I am not too sure what is to elaborate; this pattern does not require any unpacking; this describes  the current situation using the `$filter`.

Can you clarify?

graph/patterns/alternate-key.md
XVincentX: I have added an example response in 155a3f9, though I am not too sure I see what the value is

graph/patterns/alternate-key.md
XVincentX: That is what I meant and I hope it was the case, but I do see how it might be not. I have reworded it in 81ca72a, let me know if that works better!

graph/patterns/alternate-key.md
OlgaPodo: We usually show Request and Response (may be shorten) in an example, please check  evolvable-enums.md

graph/patterns/alternate-key.md
XVincentX: I understand. While I still believe we are using the incorrect term, I have changed it everywhere in e50eb11, we can table a discussion in another occasion.

graph/patterns/alternate-key.md
dkershaw10: Yeah - that makes sense.  The problem is that there is still a notion of a "primary" unique identifier and an "alternative" unique identifier.  The addressing patterns for the "primary" and the alternative are different too.

OK if you have an issue with this being too "database" backed (which I totally get), then maybe we need to change the filename and topic to be called "The alternate unique identifier pattern"?

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,99 @@
...
+      "preferredLanguage": "en-US",
+      "surname": "Vance",
+      "userPrincipalName": "bob@contoso.com",
+      "id": "1a89ade6-9f59-4fea-a139-23f84e3aef66"
+    }
+  ]
+}
+```
+
+2. Get a specific resource through the alternate key:
```
dkershaw10: The first example is using the primary key and not the alternate key, using the key-as-segment pattern.
The alternative to the key-as-segment pattern, using the primary key is 
`GET https://graph.microsoft.com/v1.0/users('1a89ade6-9f59-4fea-a139-23f84e3aef66')`

Somewhere it might be worth pointing out that the "key-as-segment" is **only** supported for the primary key.

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,99 @@
...
+---
+
+This pattern works and makes sense when the alternate key is good enough to identify a single resource and provides an useful alternative to the client; while it does not work if the resultset has more than one element.
+
+In such case, the workload has two choices:
+
+1. If the workload defines a list of alternate keys for a resource and the requested key is not there, the workload SHOULD return `422`
+2. If the workload does not have a list of alternate keys, it can still query the data source, but if the result yields more than a record, it SHOULD return `422` and encourage the user to use `$filter` rather than returning the first result of the query
+
+## Example
```
dkershaw10, line 34: Sorry I missed this first time around.
I think you need to show the CSDL that is necessary to specify an alternative key (and maybe the primary key for comparison).  I see this kind of thing in the other pattern topics.

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,99 @@
...
+oData offers resource addressing via an alternate key using the same parentheses-style convention as for the canonical key, with one difference: single-part alternate keys MUST specify the key property name to unambiguously determine the alternate key.
+
+https://graph.microsoft.com/v1.0/users(0) - Retrieves the employee with ID = 0
+https://graph.microsoft.com/v1.0/users(email='bob@contoso.com') Retrieves the employee with the email matching `bob@contoso.com`
+
+## When to Use this Pattern
+
+---
+
+This pattern works and makes sense when the alternate key is good enough to identify a single resource and provides an useful alternative to the client; while it does not work if the resultset has more than one element.
```
dkershaw10: Hmm not sure about this.
I would think that with the annotations in the CSDL, a service is explicit about which alternative keys it supports, and would only support those that return a single record. Hence I think it's more likely for the service to return a 400 error that the alternate key used (in the client request) is not supported.  Anyways @darrelmiller is the expert on HTTP here...

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Problem
+
+---
+
+The resources exposed in Graph are identified by an [UUID (Universally Unique Identifier)](https://en.wikipedia.org/wiki/Universally_unique_identifier) - which guarantees uniqueness inside the same resource type. Often though, that same resource can also be uniquely identified by an alternative, more convenient property that provides a better developer experience.
+
+Take a look at the `user` resource: while the `id` remains a perfectly valid way to get the resource details, the `mail` address is also an unique property that could be used to identify it.
+
+While it is still possible to use the oData filter, such as
```
darrelmiller:
```suggestion
While it is still possible to use the OData filter, such as
```

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,99 @@
...
+
+`https://graph.microsoft.com/v1.0/users?$filter=mail eq 'bob@contoso.com'`, the returned result is wrapped in an array that needs to be unpacked.
+
+## Solution
+
+---
+
+oData offers resource addressing via an alternate key using the same parentheses-style convention as for the canonical key, with one difference: single-part alternate keys MUST specify the key property name to unambiguously determine the alternate key.
```
darrelmiller:
```suggestion
OData offers resource addressing via an alternate key using the same parentheses-style convention as for the canonical key, with one difference: single-part alternate keys MUST specify the key property name to unambiguously determine the alternate key.
```

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,99 @@
...
+      "id": "1a89ade6-9f59-4fea-a139-23f84e3aef66"
+    }
+  ]
+}
+```
+
+2. Get a specific resource through the alternate key:
+
+```http
+GET https://graph.microsoft.com/v1.0/users/1a89ade6-9f59-4fea-a139-23f84e3aef66
```
OlgaPodo, line 103: this is an example of unique primary key

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,99 @@
...
+`https://graph.microsoft.com/v1.0/users?$filter=mail eq 'bob@contoso.com'`, the returned result is wrapped in an array that needs to be unpacked.
+
+## Solution
+
+---
+
+oData offers resource addressing via an alternate key using the same parentheses-style convention as for the canonical key, with one difference: single-part alternate keys MUST specify the key property name to unambiguously determine the alternate key.
+
+https://graph.microsoft.com/v1.0/users(0) - Retrieves the employee with ID = 0
+https://graph.microsoft.com/v1.0/users(email='bob@contoso.com') Retrieves the employee with the email matching `bob@contoso.com`
```
darrelmiller, line 26: This doesn't actually work.  We should probably mention that this is a hypothetical example

XVincentX: I have added a note in 590da3e

I tried with both `users` and `applications` in graph explorer, but they are both not working at this point.

graph/patterns/alternate-key.md
XVincentX: I have added the necessary CSDL markup in 7250de3

graph/patterns/alternate-key.md
XVincentX: > and would only support those that return a single record. 

I fail to understand your point; isn't that what I am saying in the paragraph above?

graph/patterns/alternate-key.md
XVincentX: > The first example is using the primary key and not the alternate key, using the key-as-segment pattern.

That was my intention (and I misarticulated the point) - the idea was to show that the 3 calls are all equivalent, since they all return the same entity.

Let me know if ba51283 works for you

graph/patterns/alternate-key.md
XVincentX: Thinking more about this, I believe you are right, and primary key is generic enough. For some reason my brain immediately linked in to a database situation, which is definitely not always the case. I have changed the wording in 1f3d11e

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,135 @@
...
+## When to Use this Pattern
+
+---
+
+This pattern works and makes sense when the alternate key is good enough to identify a single resource and provides an useful alternative to the client; while it does not work if the resultset has more than one element.
+
+In such case, the workload has two choices:
+
+1. If the workload defines a list of alternate keys for a resource and the requested key is not there, the workload SHOULD return `422`
+2. If the workload does not have a list of alternate keys, it can still query the data source, but if the result yields more than a record, it SHOULD return `422` and encourage the user to use `$filter` rather than returning the first result of the query
```
PetRich-MSFT: This seems problematic here in the scenario where an entity is adding to the collection which makes the alternative key non-unique e.g.
1) Caller makes a request for https://graph.microsoft.com/v1.0/users(email='bob@contoso.com') and gets a 200
2) Another user is added who also has the email of bob@contoso
3) Caller makes a request for https://graph.microsoft.com/v1.0/users(email='bob@contoso.com') and now gets a 422

In my opinion a workload should only support an alternative key if that key is guaranteed to be unique within the dataset, which would imply that we should only support explicitly declared alternative keys

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,135 @@
...
+  "id": "1a89ade6-9f59-4fea-a139-23f84e3aef66"
+}
+```
+
+3. Requesting a resource through an alternate key which yields more than a result
+
+```http
+GET https://graph.microsoft.com/v1.0/users(name='Bob')
+
+400 Bad Request
```
PetRich-MSFT, line 132: Shouldn't this be 422 to match the above documentation
```suggestion
422 Unprocessable Entity
```

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,135 @@
...
+```
```
PetRich-MSFT, line 133: Should we add an example of the case where multiple properties are required to form an alternative key e.g. https://graph.microsoft.com/v1.0/users(email='bob@contoso.com',ssn='1234')

graph/patterns/alternate-key.md
PetRich-MSFT: FYI the [Odata spec](https://github.com/OData/vocabularies/blob/master/OData.Community.Keys.V1.md#223key-value-uniqueness) says

> Alternate Keys are intended to allow callers to refer to a specific entity instance. Therefore, services must allow 1 and only 1 record in an entity collection to match any set of values for the set of properties defined for each key.

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,135 @@
...
+  "mobilePhone": "+1 425 555 0109",
+  "officeLocation": "18/2111",
+  "preferredLanguage": "en-US",
+  "surname": "Vance",
+  "userPrincipalName": "bob@contoso.com",
+  "id": "1a89ade6-9f59-4fea-a139-23f84e3aef66"
+}
+```
+
+3. Requesting a resource through an alternate key which yields more than a result
```
PetRich-MSFT:
```suggestion
3. Requesting a resource through an alternate key which yields more than one result
```

graph/patterns/alternate-key.md
XVincentX: It should be 400 per previous conversation. Changed the documentation above

graph/patterns/alternate-key.md
XVincentX: You are correct. I was not aware that alternative keys must be declared in advance in the CSDL file

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Problem
+
+---
+
+The resources exposed in Graph are identified through a Primary Key - which guarantees uniqueness inside the same resource type. Often though, that same resource can also be uniquely identified by an alternative, more convenient property (or set of properties) that provides a better developer experience.
+
+Take a look at the `user` resource: while the `id` remains a perfectly valid way to get the resource details, the `mail` address is also an unique property that could be used to identify it.
+
+While it is still possible to use the OData filter, such as
```
dkershaw10: I wonder if we even need to mention OData here.  Can we just say "use the $filter query parameter"

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,130 @@
...
+
+`https://graph.microsoft.com/v1.0/users?$filter=mail eq 'bob@contoso.com'`, the returned result is wrapped in an array that needs to be unpacked.
```
dkershaw10: Needs a GET

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Solution
+
+---
+
+OData offers resource addressing via an alternate key using the same parentheses-style convention as for the canonical key, with one difference: single-part alternate keys MUST specify the key property name to unambiguously determine the alternate key. (Note: this is a hypothetical sample)
```
dkershaw10: Again, I think we can drop the OData here.
Can we say: "Resource addressing via an alternative key can be achieved..."

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,130 @@
...
+
+https://graph.microsoft.com/v1.0/users(0) - Retrieves the employee with ID = 0
+https://graph.microsoft.com/v1.0/users(email='bob@contoso.com') Retrieves the employee with the email matching `bob@contoso.com`
+
+## When to Use this Pattern
+
+---
+
+This pattern works and makes sense when the alternate key is good enough to identify a single resource and provides an useful alternative to the client; while it does not work if the resultset has more than one element. In such case, the workload SHOULD return `400`
```
dkershaw10: I see what you are saying - but in my mind this isn't quite correct.
A property can only be a candidate for an alternate key if it only ever uniquely identifies one single resource.  At that point a service can decide (based on scenarios/customer feedback) whether to offer that property as an alternate key.  This is an explicit decision through schema annotations.  There should not be a world where GET by alternate key returns more than one object.  
Hence the 400 is more about the service not supporting the property used in the request as an alternate key

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,130 @@
...
+## Example
+
+---
+
+The same user identified via the alternate key SSN, the canonical (primary) key ID using the non-canonical long form with specified key property name, and the canonical short form without key property name
+
+Declare `mail` and `ssn` as alternate keys on an entity:
+
+```xml
+<EntityType Name="User">
```
dkershaw10: Name="user"
We should follow our own naming convention rules in our examples.

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,130 @@
...
+         </Record>
+      </Collection>
+   </Annotation>
+</EntityType>
+```
+
+1. Get a specific resource through `$filter`:
+
+```http
+GET https://graph.microsoft.com/v1.0/users/?$filter=(ssn eq '123-45-6789')
```
dkershaw10: No parentheses - please remove.

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,130 @@
...
+      "preferredLanguage": "en-US",
+      "surname": "Vance",
+      "userPrincipalName": "bob@contoso.com",
+      "id": "1a89ade6-9f59-4fea-a139-23f84e3aef66"
+    }
+  ]
+}
+```
+
+2. Get a specific resource through its primary key, and then through the two alternate keys:
```
dkershaw10: Get a specific resource either through its primary key, or through the two alternate keys

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,130 @@
...
+
+```http
+GET https://graph.microsoft.com/v1.0/users/1a89ade6-9f59-4fea-a139-23f84e3aef66
```
dkershaw10, line 103: I think somewhere you should note that this is our recommended "key-as-segment" addressing approach when using primary keys, and that key-as-segment pattern is NOT supported for alternated key pattern.  The latter think could go in an Issues section (take a look at the topic template).

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,130 @@
...
+  "mobilePhone": "+1 425 555 0109",
+  "officeLocation": "18/2111",
+  "preferredLanguage": "en-US",
+  "surname": "Vance",
+  "userPrincipalName": "bob@contoso.com",
+  "id": "1a89ade6-9f59-4fea-a139-23f84e3aef66"
+}
+```
+
+3. Requesting a resource through an alternate key which yields more than one result
```
dkershaw10: Per earlier comment - this should never happen.  The service should never make "name" as a valid alternate key.
Hence I think the 400 error is that "name" is not supported as an alternate key.

I would title this "Requesting a resource for an unsupported alternate key property"  or some such.

XVincentX: Is this ensured by AGS or by the workload? In any case that is correct, I will reword this part

dkershaw10: That's a good question.  AGS has the schema and annotations, so *maybe* it could (and does) catch the use of a property that is not marked as an alternate key and save the round trip to the workload (fail fast).  But I don't know the answer.

graph/patterns/alternate-key.md
XVincentX: I understand. My original rationale was more that the concept of the key in a backing store is in relational databases, but not in other ones - so the concept of an alternate key is something that the workload would need to handle - including returning an error in case the query yields more than a result.

I will remove this.

graph/patterns/alternate-key.md
XVincentX: Added a note. Out of curiosity - does the primary key also work as alternate key automatically? There would be no explicit CSDL annotation

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,132 @@
...
+  ]
+}
+```
+
+2. Get a specific resource either through its primary key, or through the two alternate keys:
+
+```http
+GET https://graph.microsoft.com/v1.0/users/1a89ade6-9f59-4fea-a139-23f84e3aef66
+GET https://graph.microsoft.com/v1.0/users(ssn='123-45-6789')
+GET https://graph.microsoft.com/v1.0/users(email='bob@contoso.com')
```
PetRich-MSFT: Should be mail for consistency with other examples
```suggestion
GET https://graph.microsoft.com/v1.0/users(mail='bob@contoso.com')
```

graph/patterns/alternate-key.md
```diff
@@ -0,0 +1,132 @@
...
+    }
+  ]
+}
+```
+
+2. Get a specific resource either through its primary key, or through the two alternate keys:
+
+```http
+GET https://graph.microsoft.com/v1.0/users/1a89ade6-9f59-4fea-a139-23f84e3aef66
+GET https://graph.microsoft.com/v1.0/users(ssn='123-45-6789')
```
PetRich-MSFT, line 104: Minor nitpick: the example response does not include a ```ssn``` property

PetRich-MSFT (commented):
Added a couple of minor nitpicks

PetRich-MSFT (approved):

dkershaw10 (approved):
LGTM

OlgaPodo (approved):

PR #349, opened 2022-05-02
Change Status pattern

XVincentX:
This PR introduces the Change Status Pattern discussed internally. This is supposed to be proposed in the next API Council when I get enough people to accept the proposal.


graph/patterns/change-status.md
```diff
@@ -0,0 +1,84 @@
...
+
+Instead of returning an empty body with `204` status code, depending on the situation, the response should contain the status of the request, if can be satisfied in a reasonable time:
+
+```json
+{
+  "newStatus": "confirmed"
+}
+```
+
+If the response triggers a long running operation, the response should contain a reference to query its status. The status on the resource shall stay the same until the long running operation is completed:
```
corranrogue9: For long running operations, the [existing guidance](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211713/Long-running-operations) suggests always using an entity to track the progress of the operation, and that the entity should be created (and therefore the request for the operation made) with a `POST` or a `PUT`. I think that makes a lot of sense for the case where the operation isn't revolving around changing the status of another entity, but I get your point here. I might have something like a "VM" resource that can be started, shutdown, or stopped. In that case, starting and shutting down might be long running operations, but stopping should be pretty immediate. It might be worth briefly addressing that here, and maybe updating the long running operations guidance to refer here for cases like that.

graph/patterns/change-status.md
```diff
@@ -0,0 +1,84 @@
...
+
+```xml
+<EnumType Name="riskyStatus">
+    <Member Name="confirmed" Value="0"/>
+    <Member Name="dismissed" Value="1"/>
+    <Member Name="unknown" Value="2"/>
+</EnumType>
+
+<EntityType Name="riskAssessment" >
+    <Property Name="status" Type="riskyStatus" />
```
corranrogue9: Just to fully illustrate the point, it might be worth it here to make sure the `riskAssessment` somehow refers to the resource whose status is actually being updated.

XVincentX: I am not too sure I understand the suggestion. Assuming the `riskAssessment` belongs to the `riskyUser` resource, what would you like to see here?

`curl --json '{"status": "confirmed", "duration": 3600}' https://graph.microsoft.com/v1.0/riskyUsers/a57dc75f-24b5-47ce-b5e1-44822f5d4729/riskAssessments`

graph/patterns/change-status.md
XVincentX: I have on purpose not articulated the long running operation scenario - the idea was more to convey the message that this works with long running operation as well, but still suggesting to follow the existing guidance we have in place.

I have added a link to the existing guidance for reference and modified the returned the payload. Can you take another look and see if that makes more sense now?

graph/patterns/change-status.md
```diff
@@ -0,0 +1,86 @@
...
+
+_The change status pattern is an effective alternative to model the ability to trigger side effects while keeping an agnostic, data driven approach_
+
+## Problem
+
+In some cases, exposed resources can trigger side effects that modify the system behavior. For instance, a [`riskyUser`](https://docs.microsoft.com/en-us/graph/api/resources/riskyuser?view=graph-rest-1.0) can be [`confirmed`](https://docs.microsoft.com/en-us/graph/api/riskyuser-confirmcompromised?view=graph-rest-1.0&tabs=http) or [`dismissed`](https://docs.microsoft.com/en-us/graph/api/riskyuser-confirmcompromised?view=graph-rest-1.0&tabs=http).
+
+These situation are often modeled as a `POST` request to a specific endpoint in the same URI space of the resource, triggering the side effect directly and not returning any data. While it might be a quick and easy solution, there are some considerations:
+
+- The API should be exposing resources and allow their manipulation, while Actions resemble an object in [OOP][1]
```
dkershaw10: Would it be better to say this is more like a remote procedure call?

graph/patterns/change-status.md
```diff
@@ -0,0 +1,86 @@
...
+- It couples the request with the side effect, making more difficult to test, log, replay and reason about what is happening in the system
+- It forces the consumer to follow the [OOP][1] approach at the API level. We do not want to pursue any specific style and the developers writing the client might be working in other ways
```
dkershaw10, line 15: Maybe it forces a non-standard pattern, where generally POST is used to create resources?

graph/patterns/change-status.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Solution
+
+API Designers can prefer the usage of a `status` property or a datafyed version of the intended action to represent the **intent** of triggering a side effect rather than starting the execution directly through the HTTP request.
```
dkershaw10: It seems like with this proposal there's a currentState and a desiredState, with desiredState capturing the intended target state?

graph/patterns/change-status.md
```diff
@@ -0,0 +1,86 @@
...
+
+Taking in consideration the example of the `riskyUser` above, it is possible to add a `status` property to the resource:
+
+```xml
+<EnumType Name="riskyStatus">
+    <Member Name="confirmed" Value="0"/>
+    <Member Name="dismissed" Value="1"/>
+    <Member Name="unknown" Value="2"/>
```
dkershaw10: unknownFutureValue

graph/patterns/change-status.md
```diff
@@ -0,0 +1,86 @@
...
+
+Instead of returning an empty body with `204` status code, depending on the situation, the response should contain the status of the request, if can be satisfied in a reasonable time:
+
+```json
+{
+  "newStatus": "confirmed"
+}
+```
+
+If the response triggers a long running operation, the response should contain a reference to query its status, following the [existing guidance][2]. The status on the resource shall stay the same until the long running operation is completed:
```
dkershaw10: The link should be updated to the LRO topic that we should have on this site, rather than the internal site.

graph/patterns/change-status.md
```diff
@@ -0,0 +1,86 @@
...
+
+```json
+{
```
dkershaw10, line 58: Is this a new complex type? Is there a name for this complex type?
Be good to show the full response - including the HTTP response code and any relevant headers and the full object that is returned.

graph/patterns/change-status.md
```diff
@@ -0,0 +1,86 @@
...
+```xml
+<EnumType Name="riskyStatus">
+    <Member Name="confirmed" Value="0"/>
+    <Member Name="dismissed" Value="1"/>
+    <Member Name="unknown" Value="2"/>
+</EnumType>
+
+<EntityType Name="riskAssessment" >
+    <Property Name="status" Type="riskyStatus" />
+    <Property Name="duration" Type="Edm.Int32" />
```
dkershaw10: We should use Edm.Duration here, following the ISO standard for durations.

graph/patterns/change-status.md
```diff
@@ -0,0 +1,86 @@
...
+</EntityType>
+```
+
+```bash
+curl --json '{"status": "confirmed", "duration": 3600}' https://graph.microsoft.com/v1.0/riskyUsers/a57dc75f-24b5-47ce-b5e1-44822f5d4729/riskAssessments
```
dkershaw10: Again show the full response here (and how this is related to riskyUsers.

graph/patterns/change-status.md
```diff
@@ -0,0 +1,86 @@
+# Change Status & Action Request
+
+Microsoft Graph API Design Pattern
+
+_The change status pattern is an effective alternative to model the ability to trigger side effects while keeping an agnostic, data driven approach_
+
+## Problem
+
+In some cases, exposed resources can trigger side effects that modify the system behavior. For instance, a [`riskyUser`](https://docs.microsoft.com/en-us/graph/api/resources/riskyuser?view=graph-rest-1.0) can be [`confirmed`](https://docs.microsoft.com/en-us/graph/api/riskyuser-confirmcompromised?view=graph-rest-1.0&tabs=http) or [`dismissed`](https://docs.microsoft.com/en-us/graph/api/riskyuser-confirmcompromised?view=graph-rest-1.0&tabs=http).
```
dkershaw10: While this is a good and simple example, we point to something that is GA with the anti-pattern.  Maybe we should pick another example where the correct pattern has been adopted instead?

XVincentX: I would be all for it, but I have not found any - and I am afraid that this is the way things have been modeled in Graph since forever. I will take another look, but I am not confident I can find something

graph/patterns/change-status.md
XVincentX: That value is not meant to recall the evolvable enum feature, but rather the fact that the user has been detected as potentially risky, but nor confirmed or dismissed. I have changed the string value in 7491f89

graph/patterns/change-status.md
XVincentX: 9df0d2d

graph/patterns/change-status.md
XVincentX: Done in 874176d

graph/patterns/change-status.md
XVincentX: Not necessarily - the problem is that it uses POST to trigger a side effect, while what we want is to create a resource which will trigger the side effect somehow. I can put emphasis on this if you want!

graph/patterns/change-status.md
XVincentX: This is the same data structure that is used in the [LRO](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#1323-post-hybrid-model) document. I have not been able to find a name - is it specified somewhere?

graph/patterns/change-status.md
XVincentX: I have been wondering about it myself. The initiating user can always query the operation since they have a reference, but other people would not. I have added the proposal in - let me know what you think: d85a901

graph/patterns/change-status.md
dkershaw10: That makes sense to me.  I believe that Olga is working on a new long running operations topic, but you should link to the Microsoft REST API guidelines for now (rather than an internal wiki) - https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#13-long-running-operations

graph/patterns/change-status.md
dkershaw10: Yes I should have been more clear. What I was trying to say is that typically a POST is used to directly create/insert a resource record (by POSTing to the resource collection). This is what I would class as a "traditional" pattern.

Using POST to do other operations through an action may be confusing, when there are "traditional" patterns through POST/PATCH/DELETE.  Actions should really be a last resort when there are no clear alternatives.

graph/patterns/change-status.md
dkershaw10: Actually thinking about this again, I think it's good to use this example to show how it should have been done.  [Aside: Potentially the API can be updated without a breaking change, and the actions can be deprecated.]

If we are using this as an example, then I would recommend showing more of the request and response in your examples below.

```json
PATCH /identityProtection/riskyUsers/{id}
{
   "desiredRiskState": "confirmedCompromised"  <--- new additional property
}

200 OK
{
   "id": "{id}",
   "riskState": "confirmedCompromised",
   "riskDetail": "adminConfirmedUserCompromised",
   "riskLastUpdatedDateTime":  "{now-in-utc-time}",
   "desiredRiskState": "confirmedCompromised",
 }
```

That's not all the properties, but just showing the ones that are relevant.
[You can do the example using curl like you have below]

This would be the "synchronous" operation.
Alternatively, we could just say that riskState itself is updateable, but only for setting dismissed or confirmedCompromised values.  All other values are system set values and hence are read-only - setting those would be a 400 error. 

The challenge here is that the existing actions are also bulk operations (changing the state of multiple users in one request).  We should put this to one side for the time-being IMHO.

We might actually want to have 2 examples (using riskyUsers) - one where the operation is synchronous, and one where the operation is long-running.  While we can point to the LRO docs, I think it's good to see how these might be combined.

```json
PATCH /identityProtection/riskyUsers/{id}
{
   "desiredRiskState": "confirmedCompromised"  <--- new additional property
}

200 OK
{
   "id": "{id}",
   "riskState": "atRisk",
   "riskDetail": "none",
   "riskLastUpdatedDateTime":  "{some-time-days-in-the-past-in-utc-time}",
   "desiredRiskState": "confirmedCompromised",
 }
```

You can then poll until:
```json
GET /identityProtection/riskyUsers/{id}

200 OK
{
   "id": "{id}",
   "riskState": "confirmedCompromised",
   "riskDetail": "adminConfirmedUserCompromised",
   "riskLastUpdatedDateTime":  "{now-in-utc-time}",
   "desiredRiskState": "confirmedCompromised",
 }
```

The other example to point to (which is much more complex) is the model that PIM and Access Reviews use - where there is a request object for creating the actual object.  https://docs.microsoft.com/en-us/graph/api/resources/privilegedidentitymanagementv3-overview?view=graph-rest-1.0 and https://docs.microsoft.com/en-us/graph/api/resources/accesspackageassignmentrequest?view=graph-rest-1.0

graph/patterns/change-status.md
dkershaw10: OK - so given that object already has "riskyState" any additional property is effectively a desiredState

graph/patterns/change-status.md
```diff
@@ -0,0 +1,98 @@
...
+  "status": "disabling"
+}
+```
+
+The status on the resource shall stay the same until the long running operation is completed, but the workload can optionally return a `desiredStatus` property along with the other resource's properties to signal that it is working on a request:
+
+curl https://graph.microsoft.com/v1.0/riskyUsers/a57dc75f-24b5-47ce-b5e1-44822f5d4729
+
+```json
+{ "status": "potential", "desiredStatus": "confirmed" }
```
dkershaw10: I think status here should be the existing riskyState.  Not sure about the "potential" value though - it should probably show the current state until it's updated.  The object also has a riskLastModifiedDateTime, and riskDetail that would both be updated as a side-effect of the riskyState updating.

graph/patterns/change-status.md
XVincentX: I did change the link. Resolving this

graph/patterns/change-status.md
XVincentX: Oh I see what you mean now - point out that usually POST is used to create resources, not to trigger side effects.

I have added such point in 7f3a2c4

graph/patterns/change-status.md
XVincentX: That is correct. I would be fine to "mandate" to call it `desiredState` if you want, but at the moment I do not see any automated tooling that might use that property to do anything useful, thus leaving the naming to the implementer.

graph/patterns/change-status.md
XVincentX: > I think status here should be the existing riskyState

Isn't it the case? We are asking to return back the current status for the property - in this case the `riskyState = potential`

>  it should probably show the current state until it's updated

Isn't it what the guideline is saying?

> The status on the resource shall stay the same until the long running operation is completed

---

> The object also has a riskLastModifiedDateTime, and riskDetail that would both be updated as a side-effect of the riskyState updating

Good point. Maybe we can change the guideline to say "you should also return all the appropriate properties changed because of the request of change of status", what do you think?

graph/patterns/change-status.md
XVincentX: > Potentially the API can be updated without a breaking change, and the actions can be deprecated

If this gets passed, I would love to work with some "early adopter" workload to try this out and see if that fits the use case better!

> The challenge here is that the existing actions are also bulk operations (changing the state of multiple users in one request). We should put this to one side for the time-being IMHO.

I would be open to work on this use case as well, but to be honest I would need some evidence that the system is either significantly slower if doing X requests instead of a single request with an array of X user ids or some good scenario/customer complain.

To expand a little bit my line of thinking, if you have an operation on a single object, you don’t need to define another version just to operate on a collection of those objects.
 

I give the developer `X`, and they can do `map collection Y over X`. That is not something I should be providing.

Having a separate definition for processing a group of things implies that there is something special about processing a group instead of a single item. If that is the case (for example, a more efficient batch implementation.) I might (not) be open to consider it. Let me know if that makes sense or not.

graph/patterns/change-status.md
```diff
@@ -0,0 +1,107 @@
+# Change Status & Action Request
+
+Microsoft Graph API Design Pattern
+
+_The change status pattern is an effective alternative to model the ability to trigger side effects while keeping an agnostic, data driven approach_
+
+## Problem
+
+In some cases, exposed resources can trigger side effects that modify the system behavior. For instance, a [`riskyUser`](https://docs.microsoft.com/en-us/graph/api/resources/riskyuser?view=graph-rest-1.0) can be [`confirmedCompromised`](https://docs.microsoft.com/en-us/graph/api/riskyuser-confirmcompromised?view=graph-rest-1.0&tabs=http) or [`dismissed`](https://docs.microsoft.com/en-us/graph/api/riskyuser-confirmcompromised?view=graph-rest-1.0&tabs=http).
+
```
OlgaPodo, line 10: I think we need to provide some example of system behavior modification here.

graph/patterns/change-status.md
```diff
@@ -0,0 +1,107 @@
...
+These situation are often modeled as a `POST` request to a specific endpoint in the same URI space of the resource, triggering the side effect directly and not returning any data. While it might be a quick and easy solution, there are some considerations:
```
OlgaPodo, line 11: POST is not an idempotent operation in Graph thus side effects are expected. here is a ref https://github.com/microsoft/api-guidelines/blob/graph/Guidelines.md#74-supported-methods

graph/patterns/change-status.md
```diff
@@ -0,0 +1,107 @@
...
+
+```xml
+<EnumType Name="riskState">
+    <Member Name="confirmedCompromised" Value="0"/>
+    <Member Name="dismissed" Value="1"/>
+    <Member Name="none" Value="2"/>
+</EnumType>
+
+<EntityType Name="riskyUser" >
+    <Property Name="riskState" Type="riskState" />
```
OlgaPodo, line 32: I think if you suggest adding additional properties such as desiredRiskState you need to show it in the entity model. We recommend strongly typed entities and recently made a decision not to allow open types at all for new APIs.

graph/patterns/change-status.md
OlgaPodo: please refer to the latest WIP LRO  https://github.com/microsoft/api-guidelines/blob/f3371075fd1884758238fb1b40b0766f0a9441db/graph/patterns/longRunningOperations.md

graph/patterns/change-status.md
```diff
@@ -0,0 +1,107 @@
...
+```xml
+<EnumType Name="riskState">
+    <Member Name="confirmedCompromised" Value="0"/>
+    <Member Name="dismissed" Value="1"/>
+    <Member Name="unknown" Value="2"/>
+</EnumType>
+
+<EntityType Name="riskAssessment" >
+    <Property Name="riskState" Type="riskState" />
+    <Property Name="duration" Type="Edm.Duration" />
```
OlgaPodo, line 88: I'm not sure what to do with duration property. For richLongRunningOperation we have percentageComplete. Does duration serve a similar role?

graph/patterns/change-status.md
```diff
@@ -0,0 +1,107 @@
...
+
+<EntityType Name="riskyUser" >
+    <Property Name="riskState" Type="riskState" />
+</EntityType>
+```
+
+To confirm the user as compromised, we can send a `PATCH` request on the resource:
+
+```bash
+curl -X PATCH --json '{"riskState": "confirmedCompromised"}' https://graph.microsoft.com/v1.0/identityProtection/riskyUsers/a57dc75f-24b5-47ce-b5e1-44822f5d4729
```
OlgaPodo: could you please use  request examples without reference to any tool as in other patterns?

graph/patterns/change-status.md
```diff
@@ -0,0 +1,107 @@
...
+
+_The change status pattern is an effective alternative to model the ability to trigger side effects while keeping an agnostic, data driven approach_
+
+## Problem
+
+In some cases, exposed resources can trigger side effects that modify the system behavior. For instance, a [`riskyUser`](https://docs.microsoft.com/en-us/graph/api/resources/riskyuser?view=graph-rest-1.0) can be [`confirmedCompromised`](https://docs.microsoft.com/en-us/graph/api/riskyuser-confirmcompromised?view=graph-rest-1.0&tabs=http) or [`dismissed`](https://docs.microsoft.com/en-us/graph/api/riskyuser-confirmcompromised?view=graph-rest-1.0&tabs=http).
+
+These situation are often modeled as a `POST` request to a specific endpoint in the same URI space of the resource, triggering the side effect directly and not returning any data. While it might be a quick and easy solution, there are some considerations:
+
+- The API should be exposing resources and allow their manipulation, while Actions resemble a remote procedure call on an object, typical of the [OOP][1]
```
OlgaPodo, line 13: i would say that a procedure call is typical for RPC interface style not necessarily OOP

graph/patterns/change-status.md
```diff
@@ -0,0 +1,107 @@
...
+- It couples the request with the side effect, making it more difficult to test, log, replay and reason about what is happening in the system
+- It forces the consumer to follow the [OOP][1] approach at the API level. We do not want to pursue any specific style and the developers writing the client might be working in other ways
```
OlgaPodo, line 15: If we want to refer to a particular style it should be an API style which is not necessarily a language style

graph/patterns/change-status.md
XVincentX: When is this going to be merged? That link is to a specific commit (so we would lose future updates). The link I have is on the branch which is a moving target.

graph/patterns/change-status.md
XVincentX: Idempotency and side effects are two different things. I fail to understand the point you're trying to make

graph/patterns/change-status.md
XVincentX: Absolutely good point, I have added it.

graph/patterns/change-status.md
XVincentX: This duration property has nothing to do with the Long Running Operations; it is an hypothetical example on how a `riskState` could be help up to that time. The general idea is to show how to model situation where the status change requires an additional parameter.

Does that explain?

graph/patterns/change-status.md
XVincentX: I am not too sure I am following your point.

What I am trying to communicate here is that the current way we are modeling this situation is forcing the SDK to also be modeled in such way - while there are developers working in different manners.

One of the advantages of representing status changes through data is that the SDK is more agnostic and does not clash with anybody.

Let me know whether this makes sense or not

graph/patterns/change-status.md
XVincentX: That is correct. My sentence is indeed more specific:

> Actions resemble a remote procedure call _on an object_

In this particular case:

`POST /riskyUsers/{id}/confirmCompromised` = `riskyUser.confirmCompromised();`

Does that clarify what I meant?

graph/patterns/change-status.md
XVincentX: Can you elaborate? The only way I could get to the specifics is that I would need to dig into the `riskyUser` system source code and see what they are doing under the hood. Or maybe you have something else in mind?

graph/patterns/change-status.md
OlgaPodo: I hope to be able to merge next week by 5/26

XVincentX: Perfect. In this case I will keep the current link so that it will automatically be updated when done

graph/patterns/change-status.md
```diff
@@ -0,0 +1,112 @@
...
+POST https://graph.microsoft.com/v1.0/identityProtection/riskyUsers/a57dc75f-24b5-47ce-b5e1-44822f5d4729/riskAssessments
+
+{"riskState": "confirmedCompromised", "duration": 3600}
+
+HTTP 200/OK
+
+{
+  "riskState": "confirmedCompromised",
+  "riskDetail": "adminConfirmedUserCompromised",
+  "riskLastUpdatedDateTime": "2015-06-19T12-01-03.4Z",
```
mikepizzo, line 102: the "riskLasUpdatedDateTime property is not in the definition of the entity, and duration is missing from the response. Is that intentional?

graph/patterns/change-status.md
```diff
@@ -0,0 +1,112 @@
...
+    <Property Name="riskState" Type="riskState" />
+</EntityType>
+```
+
+To confirm the user as compromised, we can send a `PATCH` request on the resource:
+
+```bash
+PATCH https://graph.microsoft.com/v1.0/identityProtection/riskyUsers/a57dc75f-24b5-47ce-b5e1-44822f5d4729
+
+{"riskState": "confirmedCompromised"}
```
mikepizzo, line 41: we sometimes define actions to make state transitions more discoverable -- for example, I can't arbitrarily set a status to completed, i have to go through created/inreview/approved/completed.  actions don't protect against calling in the wrong state, but may be more discoverable.

graph/patterns/change-status.md
```diff
@@ -0,0 +1,112 @@
...
+
+These situation are often modeled as a `POST` request to a specific endpoint in the same URI space of the resource, triggering the side effect directly and not returning any data. While it might be a quick and easy solution, there are some considerations:
+
+- The API should be exposing resources and allow their manipulation, while Actions resemble a remote procedure call on an object, typical of the [OOP][1]
+- It couples the request with the side effect, making it more difficult to test, log, replay and reason about what is happening in the system
+- It forces the consumer to follow the [OOP][1] approach at the API level. We do not want to pursue any specific style and the developers writing the client might be working in other ways
+- A `POST` request is used to create a resource record. Using it to do other operations through an action is confusing, and should really be a last resort when there are no clear alternatives.
+
+## Solution
```
mikepizzo, line 18: Another pattern where we might use actions would be where you are describing a change (i.e., increment the value of a property) rather than a state (i.e., the final value of the property). The former may be better in a concurrent scenario than the later.

graph/patterns/change-status.md
XVincentX: Do you this is something that the state diagram in the public documentation might help mitigate?

graph/patterns/change-status.md
XVincentX: Can you elaborate more? What would be the advantages over a status change or increment the property?

graph/patterns/change-status.md
OlgaPodo: https://developer.mozilla.org/en-US/docs/Glossary/Idempotent

PR #351, opened 2022-05-11
graph LRO

OlgaPodo:
Graph specific LRO updates


ghost:
Hi Olga, the best way I can help you with this review is to have you make all your changes and publish them, and then let me know the files that need an edit pass, and then I will create a branch to edit those files. Make sense? Let me know if you have any questions.

OlgaPodo:
I got the problem now 😊
Here is a link to the only file api-guidelines/longRunningOperations.md at op_LRO · microsoft/api-guidelines (github.com)<https://github.com/microsoft/api-guidelines/blob/op_LRO/graph/patterns/longRunningOperations.md> I edited and it’s in a separate working branch op_LRO.

Could you please review and provide your comments? I hope you can do it in your current role.
I wasn’t sure if I can grant Write permission on this repo because it’s shared with the other teams let us start with comments.

Greatly appreciate your help!
olga

graph/patterns/longRunningOperations.md
```diff
@@ -0,0 +1,198 @@
+# Pattern Name
+
+Microsoft Graph API Design Pattern
```
ghost: Hi Olga,
I've looked over this file. Here are my comments:
- Is the title of this page supposed to be Pattern Name, or should it be Microsoft Graph API Design Pattern or even Long Running Operations (LRO) Pattern?
- Does "Pattern" have to be capitalized? If not, it's better to lowercase it (pattern).
- For any headings, please use Sentence case capitalization instead of Title Case capitalization. For example, "When to Use this Pattern" should be "When to use this pattern"
- For numbered lists, please use bullets instead of numbers unless they are steps that need to be completed in sequence.
- Under Examples, what about using the imperative verb form instead; for example: "Create a new resource...", "Poll on..."
- I see some minor punctuation/grammatical errors throughout that I can only fix if I get edit perms to the file after you publish it. Otherwise, it's a tedious process to comment on each update that's needed.

Overall, I think the organization of these docs is good. Let me know if you have any questions.

graph/patterns/longRunningOperations.md
```diff
@@ -0,0 +1,192 @@
...
+<!-- markdownlint-enable MD033 -->
+
+The RELO pattern is the preferred pattern for long running operations and should be
+used wherever possible. The pattern avoids complexity, and consistent resource
+presentation makes things simpler for our users and tooling chain.
+
+In general, Microsoft Graph API guidelines for LRO follow [Microsoft REST API
+Guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#13-long-running-operations).
+
+A single deviation from the base guidelines is that Microsoft Graph API standards require that you use the following response headers:
```
PetRich-MSFT: Do we also want to call out that we are not allowing API producers to expose a tenant-level /operations resource (as the main API guidelines allow in https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#1324-operations-resource)

graph/patterns/longRunningOperations.md
```diff
@@ -0,0 +1,192 @@
...
+1.  If a service can create a resource with a minimal latency and continue
+    updating its status according to the well-defined and stable state
+    transition model until completion, then the RELO model is the best choice.
+
+2.  Otherwise, a service should follow the Stepwise Operation pattern. 
+ 
+
+## Issues and considerations
+
+- One or more clients MUST be able to monitor and operate on the same resource at the same time.
```
PetRich-MSFT: Should we use API consumer instead of client here, since the consumers of an API may themselves be services

```suggestion
- One or more API consumers MUST be able to monitor and operate on the same resource at the same time.
```

graph/patterns/longRunningOperations.md
```diff
@@ -0,0 +1,192 @@
...
+}
+```
+
+The API responds synchronously that the request has been accepted and includes
+the Location header with an operation resource for further polling.
+
+```
+HTTP/1.1 202 Accepted
+
+Location: https://graph.microsoft.com/v1.0/operations/123
```
PetRich-MSFT: We may want to avoid the creation of a root level operations resource since we don't have a way to do this for multiple workloads in Graph

```suggestion
Location: https://graph.microsoft.com/v1.0/databases/operations/123
```

graph/patterns/longRunningOperations.md
```diff
@@ -0,0 +1,192 @@
...
+```
+HTTP/1.1 200 OK
+
+{
+"createdDateTime": "2015-06-19T12-01-03.45Z",
+"lastActionDateTime": "2015-06-19T12-06-03.0024Z",
+"status": "succeeded",
+"resourceLocation": "https://graph.microsoft.com/v1.0/databases/db1"
+}
+```
```
PetRich-MSFT: Do we want an example of how to cancel and operation for both resource-based and stepwise

graph/patterns/longRunningOperations.md
```diff
@@ -0,0 +1,192 @@
...
+    etc. Following a canceled operation, the API should return a consistent state that allows
+    continued service.
+
+-  A recommended minimum retention time for a stepwise operation is 24 hours.
+    Operations SHOULD transition to "tombstone" for an additional period of time
+    prior to being purged from the system.
+    
+-  Services that provides a new operation resource MUST support GET semantics on the operation.
+
+
```
PetRich-MSFT: Should we allow the case where an API producer sometimes immediately completes an operation and other times returns an LRO?

darrelmiller: :-) I think so, but Gareth disagrees.  We should have a longer conversation about this.

graph/patterns/longRunningOperations.md
OlgaPodo: Thank you! good point. will update the pattern

graph/patterns/longRunningOperations.md
OlgaPodo: I think that if API response takes long time for P99 then it should be LRO and edge cases may be treated as errors

PetRich-MSFT: In my opinion for the resource-based pattern it's OK if sometimes the operation completes immediately and other times it's a LRO as in the case the shape of the API response is the same. For the stepwise pattern, my thinking is if an API can return a LRO then it should always return an LRO (even if the LRO is created in the completed state) that way API consumers don't have to deal with two different shapes of response.

We may even want to consider a metadata annotation to make any API that uses the step-wise LRO pattern

graph/patterns/longRunningOperations.md
```diff
@@ -0,0 +1,193 @@
...
+</p>
+<!-- markdownlint-enable MD033 -->
+
+- The returned resource is a new API resource called 'Stepwise Operation' and
+    is created to track the status. This LRO solution is similar to the concept
+    of Promises or Futures in other programming languages.
+
+<!-- markdownlint-disable MD033 -->
+<p align="center">
+  <img src="LRO.gif" alt="The status monitor LRO flow"/>
```
dkershaw10: Can't add comments to the diagrams - so adding here.
In the Location header it should be /operations (not /operation)

Personally I find the mix of "Resource" and /databases/db1 inconsistent.  Should the swimlane say "database"?

graph/patterns/longRunningOperations.md
```diff
@@ -0,0 +1,193 @@
...
+POST https://graph.microsoft.com/v1.0/databases/
+
+{
+"id": "db1",
+}
+```
+
+The API responds synchronously that the database has been created and indicates
+that the provisioning operation is not fully completed by including the
+Operation-Location header and status property in the response payload.
```
dkershaw10: Content-Location

graph/patterns/longRunningOperations.md
```diff
@@ -0,0 +1,193 @@
...
+}
+```
+
+The API responds synchronously that the request has been accepted and includes
+the Location header with an operation resource for further polling.
+
+```
+HTTP/1.1 202 Accepted
+
+Location: https://graph.microsoft.com/v1.0/databases/operations/123
```
dkershaw10: OK - this is problematic for me.  databases is a collection, so really this should be /databases/db1/operations/123.
Otherwise (with key-as-segment) is "operations" an ID for the databases collection?

But then that means that the database record is (at least partially) created.  In which case just use RELO...

I think for this mechanism /operations needs to be somewhere related but cannot be on the resource.

graph/patterns/longRunningOperations.md
```diff
@@ -0,0 +1,193 @@
...
+that the provisioning operation is not fully completed by including the
+Operation-Location header and status property in the response payload.
+
+```
+HTTP/1.1 201 Created
+Content-Location: https://graph.microsoft.com/v1.0/databases/db1
+
+{
+"id": "db1",
+"Status": "Provisioning",
```
dkershaw10: lowerCamelCase - feels rude saying that to another API reviewer :)

dkershaw10: So my nit with RELO is that we have this status property living on the resource in perpetuity, when really it's only of any use during the beginning of the resource's lifecycle.  I guess callers should use $select :)

graph/patterns/longRunningOperations.md
```diff
@@ -0,0 +1,193 @@
...
+    resource is no longer active. Clients MAY issue a GET on some resource to
+    determine the state of a long running operation.
+
+-  The long running operations pattern SHOULD work for clients looking to "fire and forget"
+    and for clients looking to actively monitor and act upon results.
+
+-  The long running operations pattern might be supplemented by the [change notification
+    pattern](change-notification.md).
+
+-  Cancellation of a long running operation does not explicitly mean a rollback. On a per API-defined case, it
```
dkershaw10: Can you provide examples of what this looks like for RELO?  
I have no idea what the request looks like to cancel RELO.  The Microsoft API Guidelines describe cancellation for LRO but not for RELO...

graph/patterns/longRunningOperations.md
dkershaw10: Stepwise cancellation is documented in the Microsoft API Guidelines.  RELO is not - I added the same comment earlier.

graph/patterns/longRunningOperations.md
OlgaPodo: It's a good recommendation. Will add to the pattern

graph/patterns/longRunningOperations.md
OlgaPodo: Fixed images with correct JSON but hesitant to change Resource Endpoint to Database because the pattern name is RELO

graph/patterns/longRunningOperations.md
OlgaPodo: fixed

graph/patterns/longRunningOperations.md
OlgaPodo: added a top-level node in front of databases sot it will be /storage/databases/db1 or /storage/operations/123

graph/patterns/longRunningOperations.md
OlgaPodo: Added a new RELO cancelation example

graph/patterns/longRunningOperations.md
```diff
@@ -0,0 +1,243 @@
...
+-  A recommended minimum retention time for a stepwise operation is 24 hours.
+    Operations SHOULD transition to "tombstone" for an additional period of time
+    prior to being purged from the system.
+    
+-  Services that provides a new operation resource MUST support GET semantics on the operation.
+-  Services that returns a new operation MUST always return an LRO (even if the LRO is created in the completed state) that way API consumers don't have to deal with two different shapes of response.
+
+
+
+## Examples
```
darrelmiller: Add an example for a hybrid LRO.

graph/patterns/longRunningOperations.md
darrelmiller: Boo ;-)

graph/patterns/longRunningOperations.md
OlgaPodo: will omit for simplicity for now

graph/patterns/longRunningOperations.md
OlgaPodo: Provided

graph/patterns/longRunningOperations.md
OlgaPodo: @darrelmiller while attempting to add hybrid model we identified that because the response headers contain both tatget resource URL and operation URL there is a need for the 'status' property on both resources. Otherwise a client can't determine if the target resource isn't ready by using GET on target URL. duplication of status property is an antipattern and would require additional sync work. Therefore we (Peter, Dan, Gareth, Olga) collectively decided not to include hybrid model in the pattern

PR #352, opened 2022-05-11
Drop guidance on version discovery

mikekistler:
This PR removes the guidance for version discovery features. 

This was discussed in a recent API Stewardship board session and while there was general agreement that some mechanisms are need to identifying deprecated api-versions, the current guidance does not address this is a usable fashion.


JeffreyRichter (approved):

markweitzel (approved):
Created issue #356 to capture the requirement.

PR #353, opened 2022-05-12
Update Azure API Guidelines for new LRO pattern

### LINKED ISSUE #317: Clarify guidance when a service has both long-running and "sync" versions of the same operations
> What is the recommendation when we have both long-running and "sync" versions of the same operations? The sync operations may return various 4xx. But the long-running version would need to return 200 and embed the status code in the inner error. Thus, should the top level error code more or less reflect HTTP status codes? Or should we have more specific errors?
>
>Specifically, for Form Recognizer, should the top-level error codes be InvalidRequest/InvalidArgumen/NotFound/... or ContentSourceNotAccessible/InvalidContentSourceFormat/ModelNotFound/...?

**Proposed Guidance**
Include the status code as an additional property inside the 200's error property. We do that in many of our batching APIs for a similar example that's worked well. If folks agree, I think we'd want that recommendation in the LRO section.

[Initial Discussion](https://github.com/microsoft/api-guidelines/pull/311#discussion_r769177040)

mikekistler:
Update Azure API Guidelines for new LRO pattern.


azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,197 @@ cannot collide with a resource path that contains user-specified resource ids.
 Long-running operations are an API design pattern that should be used when the processing of
 an operation may take a significant amount of time -- longer than a client will want to block
 waiting for the result.
-Azure allows for two forms of this design pattern: resource-based long-running operations (RELO),
-which is the preferred pattern, and long-running operations with a status monitor.
+The request that initiates a long-running operation generally returns a `202 Accepted` response with a
+response body that contains a representation of a _status monitor_, which is an ephemeral resource that
+will track the status and final result of the operation.
+The status monitor resource is distinct from the target resource and specific to the individual
```
```suggestion
The status monitor resource is distinct from the target resource (if any) and specific to the individual
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,197 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+    Client->>Status Monitor: GET
+    Status Monitor->>Client: HTTP/1.1 200 Ok<br/>Retry-After: 5<br/>{ "id": "22", "status": "Succeeded" }
+```
 
-In both patterns, the processing of the operation is initiated by one API call and the client
-obtains the results of the operation from a subsequent API call.
-Here we illustrate the sequence of API calls involved in each of these patterns.
+1. The client sends the request to initiate the long-running operation.
+The initial request could be a PUT, POST, or DELETE method.
+PATCH should not be used for LROs -- it should be reserved for simple resource updates.
```
```suggestion
PATCH must never be used for LROs -- it should be reserved for simple resource updates.
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,197 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+If a long-running update is required it should be implemented with POST.
+The request may contain an `operation-id` header that the service will use as the ID of the status monitor created for the operation.
```
```suggestion
The request may contain an `operation-id` header that the service uses as the ID of the status monitor created for the operation.
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,197 @@ cannot collide with a resource path that contains user-specified resource ids.
...
 
-### Resource-based long-running operations
+2. The resource validates the request and initiates the operation processing.
```
```suggestion
2. The service validates the request and initiates the operation processing.
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,197 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+It sends a response to the client with a `202-Accepted` HTTP status code.
+The response body is status monitor for the operation including the ID, either from the request header or generated by the service.
```
```suggestion
The response body is the status monitor for the operation including the ID, either from the request header or generated by the service.
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,197 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+The response also includes a `Retry-after` header telling the client a minimum time to wait (in seconds)
```
```suggestion
When returning a status monitor whose status is not in a terminal state, the response must also include a `retry-after` header indicating the minimum number of seconds the client should wait 
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,197 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+before sending a request to the status monitor URL.
```
```suggestion
before polling (GETting) the status monitor URL again for an update.
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,197 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+For backward compatibility, the response may also include an `Operation-Location` header containing the absolute URL
```
JeffreyRichter, line 188: may or must? I think we want must, right?

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,197 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+of the status monitor resource (without an api-version query parameter).
 
-In the RELO pattern, the resource that is the target of the operation contains a `status` field
-that holds the status of an outstanding or last completed operation.
-This means that the client can use a standard "get" operation on the resource to determine the
-status of an operation it initiated. The flow looks like this:
+3. After waiting at least the amount of time specified by the previous response's `Retry-after` header,
+the client issues a GET request to the status monitor using the ID in the body of the initial response.
+The GET operation for the status monitor is documented in the REST API definition and the ID
+is a path parameter.
```
```suggestion
is the last URL path segment.
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,197 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+4. The status monitor responds with information about the operation including its current status,
+which should be represented as one of a fixed set of string values in a field named `status`.
+If the operation is still being processed, the status field will contain a "non-terminal" value, like `NotStarted` or `Running`.
 
-1. The client sends the initial request to the resource to initiate the long-running operation.
-This initial request could be a PUT, PATCH, POST, or DELETE method.
+5. After the operation processing completes, a GET request to the status monitor returns the status monitor with a status field set to a terminal value -- `Succeeded`, `Failed`, or `Canceled` -- that indicates the result of the operation.
+If the status is `Failed` or `Canceled`, the status monitor resource contains an `error` field with a `code` and `message` that describes the failure.
+If the status is `Succeeded` and the LRO is an Action operation, the operation results will be returned in the `results` field of the status monitor.
+If the status is `Succeeded` and the LRO is a CRUD operation on a resource, the client can perform a GET on the resource
```
JeffreyRichter: CRUD is a bit much; it can only be create, update or delete; not read.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,197 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+to observe the result of the operation if desired.
 
-2. The resource validates the request and initiates the operation processing.
-It sends a response to client with a `200-OK` HTTP status code (or `201-Created` if the operation
-is a create operation) and a representation of the resource where the `status` field is set
-to a value indicating that the operation processing has been started.
+6. There may be some cases where a long-running operation can be completed before the response to the initial request.
+In these cases, the operation should still return a `202 Accepted` but the `status` in the response body be set to the
```
```suggestion
In these cases, the operation should still return a `202 Accepted` with the `status` property set to the appropriate terminal state.
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,197 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+terminal value `Succeeded` so that the client does not need to make an additional network request.
```
JeffreyRichter: I'd delete this. The operation can fail synchronously too.

azure/ConsiderationsForServiceDesign.md
mikekistler: I said "may" because Operation-Location is redundant with the information in the status monitor and thus not really necessary.  Our current SDK patterns use it so _for backward compatibility_ it is good to have but hopefully we'll update the generators soon so that they operate on the status monitor information and not the `Operation-Location`.

azure/ConsiderationsForServiceDesign.md
mikekistler: Agreed.  I think "CRUD" is unnecessary here.
```suggestion
If the status is `Succeeded` and the LRO is an operation on a resource, the client can perform a GET on the resource
```

azure/ConsiderationsForServiceDesign.md
mikekistler: Agreed.
```suggestion
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,196 @@ cannot collide with a resource path that contains user-specified resource ids.
...
-obtains the results of the operation from a subsequent API call.
-Here we illustrate the sequence of API calls involved in each of these patterns.
+1. The client sends the request to initiate the long-running operation.
+The initial request could be a PUT, POST, or DELETE method.
+PATCH must never be used for LROs -- it should be reserved for simple resource updates.
+If a long-running update is required it should be implemented with POST.
+The request may contain an `operation-id` header that the service uses as the ID of the status monitor created for the operation.
 
-### Resource-based long-running operations
+2. The service validates the request and initiates the operation processing.
```
markweitzel, line 182: Should we say something about what happens if the request is invalid? For example..
> If the request is not valid, the service will return a `400` level error, ideally with the reason the request was invalid. It is important to realize that this occurs prior to the LRO starting, and is therefore, NOT a failure of the long running operation.
I might be over thinking this, but for folks that are not familiar LROs, it might be good to ensure they realize that the incoming request could fail.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,196 @@ cannot collide with a resource path that contains user-specified resource ids.
...
-</p>
-<!-- markdownlint-enable MD033 -->
+4. The status monitor responds with information about the operation including its current status,
+which should be represented as one of a fixed set of string values in a field named `status`.
+If the operation is still being processed, the status field will contain a "non-terminal" value, like `NotStarted` or `Running`.
 
-1. The client sends the initial request to the resource to initiate the long-running operation.
-This initial request could be a PUT, PATCH, POST, or DELETE method.
+5. After the operation processing completes, a GET request to the status monitor returns the status monitor with a status field set to a terminal value -- `Succeeded`, `Failed`, or `Canceled` -- that indicates the result of the operation.
+If the status is `Failed` or `Canceled`, the status monitor resource contains an `error` field with a `code` and `message` that describes the failure.
```
markweitzel: Is `Canceled` an error? I think it is not. IMHO, `Canceled` is the result of a deliberate action, either by a user or another system. Because it is a deliberate action, it's not an error. Could we treat this as another terminal state but include a string for "reason" ? 

Since canceling an LRO is a POST, we could potentially pass in a reason in the body, e.g. 
POST /<status-monitor-url>:cancel
{ "reason":"i got tired of waiting"}

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,196 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+```
 
-3. After waiting at least the amount of time specified by the previous response's `Retry-after` header,
-the client issues a GET request to the status monitor URL.
+### Create (PUT) with additional long-running processing
 
-4. The status monitor URL responds with information about the operation including its current status,
-which should be represented as one of a fixed set of string values in a field named `status`.
-If the operation is still being processed, the status field will contain a "non-terminal" value, like `Processing`.
+A special case of long-running operation that occurs often is a PUT operation to create a resource
```
markweitzel: We should give an example of this to make this a bit more clear.

azure/ConsiderationsForServiceDesign.md
JeffreyRichter: The LRO "failed" to complete due to cancellation. So, since the LRO is not successful, I think the error should indicate that it failed due to cancellation.

mikekistler: I discussed with Jeffrey today and he is willing to reconsider, so we'll drop cancelled for now and see if we get feedback on this.

azure/Guidelines.md
```diff
@@ -134,7 +135,7 @@ GET    | Read (i.e. list) a resource collection | `200-OK`
 GET    | Read the resource | `200-OK`
 DELETE | Remove the resource | `204-No Content`\; avoid `404-Not Found`
 
-:white_check_mark: **DO** return status code `202-Accepted` and follow the guidance in [Long-Running Operations & Jobs](#long-running-operations--jobs) when a PUT, PATCH, POST, or DELETE method completes asynchronously.
+:white_check_mark: **DO** return status code `202-Accepted` and follow the guidance in [Long-Running Operations & Jobs](#long-running-operations--jobs) when a PUT, POST, or DELETE method completes asynchronously.
```
JeffreyRichter, line 142: I think we should add subheadings to the LRO section to be more "what I'm trying to do" focused:
- Initiating an action via POST or deleting a resource: These both return 202s and a status monitor
- Creating a resource via PUT (or POST [note: I put POST in parens because we discourage POST for resource creation]): These both return 200/201 and the resource itself (not the status monitor)
- Polling an LRO
- Controlling an LRO
- Update an resource via POST-action; do not make LRO PATCH operations

It is OK to repeat some bullets in each section as devs will only read the section relevant to them. There has been so many problems with LROs that I think structuring the bullets this way will really help bring the improvements we seek.

need to change the structure of these LRO guidelines a tiny bit. I think the first part should describe the guidelines for POST/DELETE

mikekistler: I've restructured roughly along the lines you suggest.  We need to settle on the pattern to use for POST to create a resource -- I currently have that grouped in the "standard LRO pattern" but can move it if necessary.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,201 @@ cannot collide with a resource path that contains user-specified resource ids.
...
 an operation may take a significant amount of time -- longer than a client will want to block
 waiting for the result.
-Azure allows for two forms of this design pattern: resource-based long-running operations (RELO),
-which is the preferred pattern, and long-running operations with a status monitor.
 
-In both patterns, the processing of the operation is initiated by one API call and the client
-obtains the results of the operation from a subsequent API call.
-Here we illustrate the sequence of API calls involved in each of these patterns.
+The request that initiates a long-running operation generally returns a `202 Accepted` response with a
+response body that contains a representation of a _status monitor_, which is an ephemeral resource that
```
johanste: The 202 + status monitor response body should only be returned for a POST and DELETE operation. So I'm not sure if the "generally" vs. "special case" terminology is the right choice. 2 vs. 1 may not be a 100% fair fight, but it also doesn't reach my "generally" bar :). 

I think we are also falling into the same trap that the old guidance did in that in order to correctly understand the guidance, you have to understand HTTP terminology. While it is true that the status monitor is a resource (in HTTP terminology), my question is how likely it is that our target audience will correctly interpret, or if their understanding of a "resource" is different. I've heard "it's not a resource, it is just a URL or location" plenty of times in my discussions with service teams.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,201 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+
+This diagram illustrates how a long-running operation with a status monitor is initiated and then how the client
+determines it has completed and obtains its results:
+
+```mermaid
+sequenceDiagram
+    participant Client
+    participant API Endpoint
+    participant Status Monitor
+    Client->>API Endpoint: PUT/POST/DELETE
```
johanste: Per above, PUT should not respond with a 202 .

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +142,201 @@ cannot collide with a resource path that contains user-specified resource ids.
...
-2. The resource validates the request and initiates the operation processing.
-It sends a response to client with a `200-OK` HTTP status code (or `201-Created` if the operation
-is a create operation) and a representation of the resource where the `status` field is set
-to a value indicating that the operation processing has been started.
+6. There may be some cases where a long-running operation can be completed before the response to the initial request.
+In these cases, the operation should still return a `202 Accepted` with the `status` property set to the appropriate terminal state.
 
-3. The client then issues a GET request to the resource to determine if the operation processing
-has completed.
+7. The service will auto-purge status monitor resource after completion (at least 24 hours).
```
johanste: This sentence is broken.

azure/ConsiderationsForServiceDesign.md
mikekistler: Added.

azure/ConsiderationsForServiceDesign.md
mikekistler: Now called "POST or DELETE LRO".

azure/Guidelines.md
```diff
@@ -800,68 +804,90 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
 
+:no_entry: **DO NOT** implement PATCH as an LRO.  If LRO update is required it should be implemented with POST.
+
 In rare instances where an operation may take a _very long_ time to complete, e.g. longer than 15 minutes,
 it may be better to expose this as a first class resource of the API rather than as an operation on another resource.
 
-There are two basic patterns that can be used for long-running operations:
-1. Resource-based long-running operations (RELO)
-2. Long-running operations with status monitor
+There are two basic patterns for long-running operations in Azure. In the first pattern is used for a POST or DELETE
```
```suggestion
There are two basic patterns for long-running operations in Azure. The first pattern is used for a POST and DELETE
```

azure/Guidelines.md
```diff
@@ -800,68 +804,90 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
+operation that initiates the LRO returns a `202 Accepted` response with a status monitor in the response body.
```
```suggestion
operations that initiate the LRO. These return a `202 Accepted` response with a JSON status monitor in the response body.
```

azure/Guidelines.md
```diff
@@ -800,68 +804,90 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
+The second pattern is applies only in the case of a PUT operation to create a resource that also involves additional long-running processing.
```
```suggestion
The second pattern applies only in the case of a PUT operation to create a resource that also involves additional long-running processing.
```

azure/Guidelines.md
```diff
@@ -800,68 +804,90 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
+These are described in the following two sections.
 
-:white_check_mark: **DO** use the RELO pattern when the operation is on a resource that contains a "status" property that can be used to obtain the outcome of the operation.
+#### POST or DELETE LRO pattern
 
-:ballot_box_with_check: **YOU SHOULD** only use the status monitor LRO pattern when the RELO pattern is not applicable.
+A POST or DELETE long-running operation accepts a request from the client to initiate the operation processing and returns
+a [status monitor](https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.3) that will report the operation results.
```
```suggestion
a [status monitor](https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.3) that reports the operation's progress.
```

azure/Guidelines.md
```diff
@@ -800,68 +804,90 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
 
-#### Resource-based long-running operations
+:white_check_mark: **DO** allow the client to pass an `Operation-Id` header with a ID for the status monitor for the operation.
```
```suggestion
:white_check_mark: **DO** allow the client to pass an `Operation-Id` header with an ID for the operation's status monitor.
```

azure/Guidelines.md
```diff
@@ -800,68 +804,90 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
 
-Some common situations where the RELO pattern should be used:
-1. A "create" operation (PUT, PATCH, or POST) for a resource where the basic structure of the resource is created immediately and includes a status field that indicates when the create has completed, e.g. "provisioning" -> "active".
-2. An action operation for a resource where both the initiation of the action and the completion of the action cause a change to the "status" property of the resource.
+:white_check_mark: **DO** generate an ID (typically a GUID) of the status monitor if the `Operation-Id` header was not passed by the client.
```
```suggestion
:white_check_mark: **DO** generate an ID (typically a GUID) for the status monitor if the `Operation-Id` header was not passed by the client.
```

azure/Guidelines.md
```diff
@@ -800,68 +804,90 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
+#### Create (PUT) operation with additional long-running processing
 
-:white_check_mark: **DO** return the status monitor URL in the `Operation-Location` response header.
+For a create (PUT) with additional long-running processing:
 
-:white_check_mark: **DO** support the `get` method on the status monitor endpoint that returns a `200-OK` response with a response body that contains the completion status of the operation with sufficient information to diagnose any potential failures.
+:white_check_mark: **DO** allow the client to pass an `Operation-Id` header with a ID for the status monitor for the operation.
 
-:white_check_mark: **DO** include a field in the status monitor resource named `status` indicating the operation's status. This field should be a string with well-defined values. Indicate the terminal state using "Succeeded", "Failed", or "Canceled".
+:white_check_mark: **DO** generate an ID (typically a GUID) of the status monitor if the `Operation-Id` header was not passed by the client.
```
```suggestion
:white_check_mark: **DO** generate an ID (typically a GUID) for the status monitor if the `Operation-Id` header was not passed by the client.
```

JeffreyRichter (approved):

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,205 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+{
+   "id": "22",
+   "status": "Succeeded",
+   "result": { ... }
+}
+```
 
-3. After waiting at least the amount of time specified by the previous response's `Retry-after` header,
-the client issues a GET request to the status monitor URL.
+### Create (PUT) with additional long-running processing
```
haocornell-ms: I might be missing something. But in practice, PUT is often used to either create and update. Is there a different behavior  for updating existing resources with a long-running operation? I feel the guideline can be more prescriptive on it and looks like it should be consistent with create case, i.e. updated content is applied first before LRO is finished.

JeffreyRichter: We do NOT consider PUT to create or update; we consider PUT to be create or wholesale replace. We consider PATCH to be create or update. The reason for this is due to Azure's versioning policies. I do have a video describing this [here](https://www.youtube.com/watch?v=PFK6RRTwXCg).

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,205 @@ cannot collide with a resource path that contains user-specified resource ids.
...
-obtains the results of the operation from a subsequent API call.
-Here we illustrate the sequence of API calls involved in each of these patterns.
+The request that initiates a long-running operation returns a response that points to or embeds
+a _status monitor_, which is an ephemeral resource that will track the status and final result of the operation.
+The status monitor resource is distinct from the target resource (if any) and specific to the individual
+operation request.
+
+A POST or DELETE operation returns a `202 Accepted` response with the status monitor in the response body.
+A long-running POST should not be used for resource create -- use PUT as described below.
+PATCH must never be used for LROs -- it should be reserved for simple resource updates.
```
weidongxu-microsoft: nit, before we use `LRO`, we probably want to make sure reader knows that `LROs == long-running operation`

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,205 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+1. The client sends the request to initiate the long-running operation.
+The initial request could be a PUT, POST, or DELETE method.
+The request may contain an `operation-id` header that the service uses as the ID of the status monitor created for the operation.
+
+2. The service validates the request and initiates the operation processing.
+If there are any problems with the request, the service responds with a `4xx` status code and error response body.
+Otherwise the service responds with a `202-Accepted` HTTP status code.
+The response body is the status monitor for the operation including the ID, either from the request header or generated by the service.
+When returning a status monitor whose status is not in a terminal state, the response must also include a `retry-after` header indicating the minimum number of seconds the client should wait
+before polling (GETting) the status monitor URL again for an update.
```
weidongxu-microsoft:
```suggestion
before polling (GETing) the status monitor URL again for an update.
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,205 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+    API Endpoint->>Client: HTTP/1.1 202 Accepted<br/>Retry-After: 5<br/>{ "id": "22", "status": "NotStarted" }
+    Client->>Status Monitor: GET
+    Status Monitor->>Client: HTTP/1.1 200 Ok<br/>Retry-After: 5<br/>{ "id": "22", "status": "Running" }
+    Client->>Status Monitor: GET
+    Status Monitor->>Client: HTTP/1.1 200 Ok<br/>{ "id": "22", "status": "Succeeded" }
+```
 
-### Resource-based long-running operations
+1. The client sends the request to initiate the long-running operation.
+The initial request could be a PUT, POST, or DELETE method.
```
weidongxu-microsoft: It is a bit strange that
`The remainder of this section describes the pattern for long-running POST and DELETE operations.`
then a few lines later
`The initial request could be a PUT, POST, or DELETE method.`

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,205 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+The request may contain an `operation-id` header that the service uses as the ID of the status monitor created for the operation.
```
weidongxu-microsoft: Should we specify what is a valid `operation-id`? I guess we don't want customer to inject an URL to service (to the status monitor URL). (SDK likely know to use a GUID)

Is it "MUST" in POST? If we lost the response, without it, we seems not able to get the result.
If customer uses REST API, they had to know how to handle the status monitor. So it might not be too much to ask them always provides the `operation-id` in request?

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,205 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+
+2. The service validates the request and initiates the operation processing.
+If there are any problems with the request, the service responds with a `4xx` status code and error response body.
+Otherwise the service responds with a `202-Accepted` HTTP status code.
+The response body is the status monitor for the operation including the ID, either from the request header or generated by the service.
```
weidongxu-microsoft, line 185: Should we say that the response "MUST" contains an `id` property (if 202)?

Is `status` also a "MUST" in response of initial request?

weidongxu-microsoft: I am not sure if it is just me, but `id` in response of initial request of POST could be confusing with the same `id` in response of initial request of PUT. The former is the ID for status monitor, the latter is the ID for the resource.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,205 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+```text
+HTTP/1.1 200 Ok
+ 
+{
+   "id": "22",
+   "status": "Succeeded"
+}
+```
+
+If the additional processing failed, the service may delete the original resource if it is not usable in this state.
```
weidongxu-microsoft: the "MAY" here might be ambiguous for customer. Seems the only option for them would be always manually check and delete, if service "may" delete the resource.

azure/ConsiderationsForServiceDesign.md
mikekistler:
```suggestion
PATCH must never be used for long-running operations -- it should be reserved for simple resource updates.
```

azure/ConsiderationsForServiceDesign.md
mikekistler:
```suggestion
The initial request could be a POST or DELETE method.
```

azure/ConsiderationsForServiceDesign.md
mikekistler: We have debated the optional vs required for `operation-id` and (maybe grudgingly) decided on optional. We could reopen the discussion ...

azure/ConsiderationsForServiceDesign.md
mikekistler: POST should never be used to create resources and especially not an LRO POST, so there should be no confusion about the `id` in the response body of LRO POST.

LRO PUT will return the resource in the body and the operation ID in the `Operation-Id` header, so I think this is also clear which ID is for which.

I think this is enough to avoid confusion for most users, but you want to suggest something that would make it even clearer please feel free.

azure/ConsiderationsForServiceDesign.md
mikekistler: I think the service would need to clearly document this behavior. I suppose we could make that explicit here but I feel that that's generally implied across all topics, so adding it here might mean we have to be explicit everywhere else too.

azure/ConsiderationsForServiceDesign.md
haocornell-ms: Jeffrey, "update" might be a bit misleading but it is used in many Azure API docs for PUT. One example is "[Create Or Update](https://docs.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update)" . But my point is to make the guideline specific on the behavior of PUT in the case of replacement. I would expect it is the same as create case except response code would be 202 instead of 201.

JeffreyRichter: To update a resource via an LRO, we prescribe the POST-action LRO pattern.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,205 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+```
 
-3. After waiting at least the amount of time specified by the previous response's `Retry-after` header,
-the client issues a GET request to the status monitor URL.
+### Create (PUT) with additional long-running processing
 
-4. The status monitor URL responds with information about the operation including its current status,
-which should be represented as one of a fixed set of string values in a field named `status`.
-If the operation is still being processed, the status field will contain a "non-terminal" value, like `Processing`.
+A special case of long-running operation that occurs often is a PUT operation to create a resource
```
mikekistler:
```suggestion
A special case of long-running operation that occurs often is a PUT operation to create or replace a resource
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,205 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+{
+   "id": "22",
+   "status": "Succeeded",
+   "result": { ... }
+}
+```
 
-3. After waiting at least the amount of time specified by the previous response's `Retry-after` header,
-the client issues a GET request to the status monitor URL.
+### Create (PUT) with additional long-running processing
```
mikekistler:
```suggestion
### PUT with additional long-running processing
```

azure/Guidelines.md
```diff
@@ -757,68 +761,92 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
+:white_check_mark: **DO** include a `Retry-After` header in the response to the initiating request if the operation is not complete. The value of this header should be an integer number of seconds to wait before making the first request to the status monitor.
 
-:white_check_mark: **DO** return a `202-Accepted` status code from the request that initiates an LRO with status monitor if the processing of the operation was successfully initiated.
+:ballot_box_with_check: **YOU SHOULD** include an `Operation-Location` header in the response with the absolute URL of the status monitor for the operation, but do not include an api-version query parameter.
 
-:white_check_mark: **DO** perform as much validation of the initial request as practical and return an error response immediately when appropriate (without starting the operation).
+:white_check_mark: **DO** return a status monitor in the response body as described in [Status and results of long-running operations](#status-and-results-of-long-running-operations).
 
-:white_check_mark: **DO** return the status monitor URL in the `Operation-Location` response header.
+#### Create (PUT) operation with additional long-running processing
```
mikekistler:
```suggestion
#### PUT operation with additional long-running processing
```

azure/Guidelines.md
```diff
@@ -757,68 +761,92 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
 
-:white_check_mark: **DO** support the `get` method on the status monitor endpoint that returns a `200-OK` response with a response body that contains the completion status of the operation with sufficient information to diagnose any potential failures.
+For a create (PUT) with additional long-running processing:
```
mikekistler:
```suggestion
For a create or replace (PUT) with additional long-running processing:
```

azure/Guidelines.md
```diff
@@ -757,68 +761,92 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
 
-:white_check_mark: **DO** include a field in the status monitor named `error` to contain error information -- minimally `code` and `message` fields -- when an operation fails.
+:white_check_mark: **DO** generate an ID (typically a GUID) for the status monitor if the `Operation-Id` header was not passed by the client.
 
-:white_check_mark: **DO** retain the status monitor resource for some documented period of time (at least 24 hours) after the operation completes.
+:white_check_mark: **DO** fail a request with a `400-BadRequest` if the `Operation-Id` header that matches an outstanding operation unless the request is identical to the prior request (a retry scenario).
+
+:white_check_mark: **DO** perform as much validation as practical when initiating the operation to alert clients of errors early.
+
+:white_check_mark: **DO** return a `201-Created` status code from the initial request with a representation of the resource if the resource was created successfully.
```
mikekistler:
```suggestion
:white_check_mark: **DO** return a `201-Created` status code, or `200-OK` for replace, from the initial request with a representation of the resource if the resource was created or replaced successfully.
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,205 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+PUT /items/FooBar&api-version=2022-05-01
+Operation-Id: 22
+
+{
+   "prop1": 555, 
+   "prop2": "something"
+}
+```
+
+In this case the response to the initial request is a `201 Created` to indicate that the resource has been created.
```
mikekistler:
```suggestion
In this case the response to the initial request is a `201 Created` to indicate that the resource has been created, or `200 OK` when the resource was replaced.
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,205 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+
+```mermaid
+sequenceDiagram
+    participant Client
+    participant API Endpoint
+    participant Status Monitor
+    Client->>API Endpoint: POST/DELETE
+    API Endpoint->>Client: HTTP/1.1 202 Accepted<br/>Retry-After: 5<br/>{ "id": "22", "status": "NotStarted" }
+    Client->>Status Monitor: GET
+    Status Monitor->>Client: HTTP/1.1 200 Ok<br/>Retry-After: 5<br/>{ "id": "22", "status": "Running" }
```
heaths: Nit: All `Ok` -> `OK`.

azure/ConsiderationsForServiceDesign.md
heaths: I wouldn't say *never*. Consider a Key Vault certificate. The request to create is significantly different from the resource created just by the very nature of CSRs vs. certificates. From meetings, I thought we agree this is rare and maybe doesn't need to be detailed, but shouldn't we at least acknowledge it is a (less desirable) possibility lest we have different teams coming up with different solutions?

heaths: ...in fact, it's documented below and acknowledged by Jeffrey.

azure/ConsiderationsForServiceDesign.md
heaths: If you make this change, don't forget to change the link text and bookmark on or around line 159.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,205 @@ cannot collide with a resource path that contains user-specified resource ids.
...
-for example a "reboot" operation on a VM that is in the process of rebooting, then the service should
-accept the request. The status field of the resource should then report the completion status of _both_
-operations.
+The operation is initiated with a POST operation and the operation path ends in `:action`.
 
-Note: The RELO pattern should not be used in cases where the completion status of individual operations
-may be important to users, as opposed to simply learning that an operation of the type they requested
-(e.g. create a resource with a specific name) has successfully completed.
+```text
+POST /<service-or-resource-url>:action
```
heaths: Should include an api-version like we do for other examples.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,205 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+
+When the delete operation completes successfully, a client must be able to create new resource with same name without conflicts.
+
+### Controlling a long-running operation
+
+It might be necessary to support some control action on a long-running operation, such as cancel.
+This is implemented as a POST on the status monitor endpoint with `:action` added.
+
+```text
+POST /<status-monitor-url>:cancel
```
heaths: Needs api-version (consistency).

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,205 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+```
+
+A successful response to a control operation should be a `200 Ok` with a representation of the status monitor.
```
heaths: Do we want to prescribe whether this is the same status monitor as the original being changed? For example, if you already have a separate or embedded status monitor, canceling might return the same one and the status monitor property e.g., `provisioningState` would just eventually change to "canceled". Or is that really more of an implementation detail we needn't worry about?

azure/Guidelines.md
```diff
@@ -1,8 +1,12 @@
 # Microsoft Azure REST API Guidelines
+
+<!-- cspell:ignore autorest, etag, idempotency, maxpagesize, innererror -->
```
heaths: Might be worth just putting a .vscode/cspell.json file in the repo.

azure/Guidelines.md
```diff
@@ -757,68 +761,92 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
-2. An action operation for a resource where both the initiation of the action and the completion of the action cause a change to the "status" property of the resource.
+:white_check_mark: **DO** allow the client to pass an `Operation-Id` header with an ID for the operation's status monitor.
 
-:white_check_mark: **DO** return a `200-OK` response, `201-Created` for create operations, from the request that initiates the operation.  The response body should contain a representation of the resource that clearly indicates that the operation has been accepted or started.
+:white_check_mark: **DO** generate an ID (typically a GUID) for the status monitor if the `Operation-Id` header was not passed by the client.
 
-:white_check_mark: **DO** support a get method on the resource that returns a representation of the resource including the status field that indicates when the operation has completed.
+:white_check_mark: **DO** fail a request with a `400-BadRequest` if the `Operation-Id` header that matches an outstanding operation unless the request is identical to the prior request (a retry scenario).
```
heaths: Wording is off here. Not sure what you were going for. Perhaps get rid of "that" in "that matches"?

azure/Guidelines.md
```diff
@@ -757,68 +761,92 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
 
-:white_check_mark: **DO** define the "status" field of the resource as an enum with all the values it may contain including the "terminal" values "Succeeded", "Failed", and "Canceled". See [Enums & SDKs](#enums--sdks-client-libraries).
+:white_check_mark: **DO** perform as much validation as practical when initiating the operation to alert clients of errors early.
 
-:ballot_box_with_check: **YOU SHOULD** use the name `status` for the "status" field of the resource.
+:white_check_mark: **DO** return a `202-Accepted` status code from the request that initiates an LRO if the processing of the operation was successfully initiated (except for "create (PUT) with additional processing" type LRO).
```
heaths: Would you want to change this to just "PUT with additional processing" as well?

heaths (approved):
Just a few suggestions - nothing blocking.

azure/ConsiderationsForServiceDesign.md
weidongxu-microsoft: No, we don't need to reopen the discussion (yes, remember discussion on whether to merge it with `Repeatability-Request-ID`).

What I think to be important is the specification on what value of `operation-id` is acceptable (e.g. GUID only?). This way, both user and backend developer would be clear on what to send and what to expect/validate (I assume user cannot send `!@#$%^&*(` as `operation-id`), and we prefer it to be consistent across services.
I know this is not an RFC, but still want it to be precise.

azure/ConsiderationsForServiceDesign.md
weidongxu-microsoft: Appears that I am the only one confused about the `id`. Let's drop it.

---

The POST to create resource would be a different topic.

I assume if we do that, the (created) resource ID (and maybe the full URL of the resource) would be provided in the `result` property, in the success case of status monitor poll (I could be wrong, as we didn't get to the discussion on this scenario in detail).
```
HTTP/1.1 200 Ok

{
   "id": "22",
   "status": "Succeeded",
   "result": { ... }
}
```

weidongxu-microsoft:
IMHO, one important supporting material would be a reference swagger on the new LRO (in future maybe CADL as well).
(I know I expect too much, but a mock backend and a client to play with would be best)

It is not required in this PR.

azure/ConsiderationsForServiceDesign.md
MushMal: Perhaps an example of what an operation-id can be (whether GUID) and how it's used from the clients perspective and what are the implications of it not being present.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,205 @@ cannot collide with a resource path that contains user-specified resource ids.
...
-2. The resource validates the request and initiates the operation processing.
-It sends a response to client with a `200-OK` HTTP status code (or `201-Created` if the operation
-is a create operation) and a representation of the resource where the `status` field is set
-to a value indicating that the operation processing has been started.
+6. There may be some cases where a long-running operation can be completed before the response to the initial request.
+In these cases, the operation should still return a `202 Accepted` with the `status` property set to the appropriate terminal state.
 
-3. The client then issues a GET request to the resource to determine if the operation processing
-has completed.
+7. The service will auto-purge the status monitor resource after completion (at least 24 hours).
```
MushMal: Perhaps we should include a short paragraph with examples illustrating the edge cases/negative cases. Ex: client fails to receive the response for LRO and retries with the same ID or when the ID is missing, etc.. What are the other interesting edge cases?

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,205 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+
+### Long-running delete operation
+
+A long-running delete operation follows the general pattern of a long-running operation --
+it returns a `202 Accepted` with a status monitor which the client uses to determine the outcome of the delete.
+
+The resource being deleted should remain visible (returned from a GET) until the delete operation completes successfully.
+
+When the delete operation completes successfully, a client must be able to create new resource with same name without conflicts.
+
```
MushMal, line 328: Perhaps we should mention what would happen if a client issues a POST/PATCH while the LRO delete operation is ongoing and the resource is still visible. This would be good for the implementation.

azure/ConsiderationsForServiceDesign.md
MushMal: I see there is some wording around an operation-id that it can be a GUID in the Guidelines.md file

azure/ConsiderationsForServiceDesign.md
MushMal: The case with the same ID is covered in Guidelines.md but I wonder if the reader would benefit from a paragraph of examples with edge cases

MushMal (approved):
Really, the only suggestion is for a paragraph describing edge cases to help with the implementation

azure/ConsiderationsForServiceDesign.md
mikekistler: When we update the Azure API Style guide I think we can say that the operation-id header should document allowed characters (with `pattern`) and maximum length, which is already documented for ids passed in path parameters.

https://github.com/Azure/azure-api-style-guide/blob/main/openapi-style-guide.md#path-parameters

weidongxu-microsoft: OK.

Though I would suggest that we say here that GUID "must" be supported, whatever other allowed pattern.
This would give user the minimal guarantee that if they use a GUID, it will be accepted, whatever service one is communicating (another advantage is that user be able to use the same value of `Repeatability-Request-ID` as `operation-id`).
I know we probably not having lots of LRO in data-plane, but if user uses data-plane REST API from multiple services, we could save them the time to read doc on every service to decide which value can be used (there is the difference with resource ID, which user probably had to read the doc on each service they use).

azure/ConsiderationsForServiceDesign.md
mikekistler: I changed the wording a bit to make clear that the "PUT" pattern applies to both "create" and "replace" (since they really aren't separable).

azure/ConsiderationsForServiceDesign.md
mikekistler: I added a statement in Guidelines.md:

> :no_entry: **DO NOT** use a long-running POST to create a resource -- use PUT as described below.

If some service insists on using POST rather than PUT we'll probably guide them into a design that looks like the PUT pattern, but we shouldn't offer this as an option in the Guidelines.

MushMal (approved):

weidongxu-microsoft (approved):

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,206 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+This diagram illustrates how a long-running operation with a status monitor is initiated and then how the client
+determines it has completed and obtains its results:
+
+```mermaid
+sequenceDiagram
+    participant Client
+    participant API Endpoint
+    participant Status Monitor
+    Client->>API Endpoint: POST/DELETE
+    API Endpoint->>Client: HTTP/1.1 202 Accepted<br/>Retry-After: 5<br/>{ "id": "22", "status": "NotStarted" }
```
johanste: The `Retry-After` header here is a bit ambiguous. It could easily be interpreted as "retry the POST".

azure/ConsiderationsForServiceDesign.md
johanste: I think anything other than using a GUID is inviting trouble since most well behaved services would use a GUID (or possibly some form of hash for the request - but see below) in order to avoid name conflicts (which are really bad for operations). The GUID-to-string conversion is sensitive to a couple of different things; with-or-without curly braces, with-or-without dashes and case sensitivity.

azure/ConsiderationsForServiceDesign.md
johanste: There are valid uses cases for POST for resource creation. It should be avoided, yes, and certainly not used for service implementation convenience, but for use cases where it is imperative that the service owns the uniqueness of the identity of a resource (possibly global uniqueness), POST would be the preferred pattern. I don't mind this verbiage in the guidelines as long as we, as a stewardship board, understand when it is and when it is not appropriate to use POST for resource creates.

azure/ConsiderationsForServiceDesign.md
johanste: I would strongly push for MUST include the Operation-Location. The operation location is not redundant in the response for anything that wants to act in a hyperlinked way. The status monitor resource by id only requires out-of-band knowledge in order to correctly form the URL. And while hyperlinking is not super commonly used in the general case, for single redirects etc. it has certainly proven to be a valuable concept in HTTP.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,206 @@ cannot collide with a resource path that contains user-specified resource ids.
...
-<!-- markdownlint-enable MD033 -->
+4. The status monitor responds with information about the operation including its current status,
+which should be represented as one of a fixed set of string values in a field named `status`.
+If the operation is still being processed, the status field will contain a "non-terminal" value, like `NotStarted` or `Running`.
 
-1. The client sends the initial request to the resource to initiate the long-running operation.
-This initial request could be a PUT, PATCH, POST, or DELETE method.
+5. After the operation processing completes, a GET request to the status monitor returns the status monitor with a status field set to a terminal value -- `Succeeded`, `Failed`, or `Canceled` -- that indicates the result of the operation.
+If the status is `Failed`, the status monitor resource contains an `error` field with a `code` and `message` that describes the failure.
+If the status is `Succeeded` and the LRO is an Action operation, the operation results will be returned in the `results` field of the status monitor.
```
johanste:
```suggestion
If the status is `Succeeded` and the LRO is an Action operation, the operation results will be returned in the `result` field of the status monitor.
```

azure/ConsiderationsForServiceDesign.md
MushMal: > 

I believe there are two aspects here:

1) The 'operation-id' being recommended and the edge cases covered here. 
2) The 'operation-id' being a type that's in-depth covered in the style guideline with recommendations around the normalization and code generation

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,206 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+a _status monitor_, which is an ephemeral resource that will track the status and final result of the operation.
+The status monitor resource is distinct from the target resource (if any) and specific to the individual
+operation request.
+
+A POST or DELETE operation returns a `202 Accepted` response with the status monitor in the response body.
+A long-running POST should not be used for resource create -- use PUT as described below.
+PATCH must never be used for long-running operations -- it should be reserved for simple resource updates.
+If a long-running update is required it should be implemented with POST.
+
+There is a special form of long-running operation initiated with PUT that is described
```
markweitzel, line 158: This sentence and the next are distracting at this point in the document. Because this is the main pattern we want people to follow, I would recommend striking this and simply introducing the PUT LRO section later.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,206 @@ cannot collide with a resource path that contains user-specified resource ids.
...
-2. The resource validates the request and initiates the operation processing.
-It sends a response to client with a `200-OK` HTTP status code (or `201-Created` if the operation
-is a create operation) and a representation of the resource where the `status` field is set
-to a value indicating that the operation processing has been started.
+6. There may be some cases where a long-running operation can be completed before the response to the initial request.
+In these cases, the operation should still return a `202 Accepted` with the `status` property set to the appropriate terminal state.
 
-3. The client then issues a GET request to the resource to determine if the operation processing
-has completed.
+7. The service will auto-purge the status monitor resource after completion (at least 24 hours).
```
markweitzel:
```suggestion
7. The service is responsible for purging the status-monitor resource. It should auto-purge the status monitor resource after completion (at least 24 hours).
```

azure/ConsiderationsForServiceDesign.md
markweitzel: @MushMal -- How different are the examples you mentioned from "normal" REST calls? For example, if I fail to get a response from a "regular" call, wouldn't this work the same way? I'm thinking we can simply point to other parts of the guidelines and "demystify" LROs a bit more, e.g. "they are just like other REST calls"...

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,206 @@ cannot collide with a resource path that contains user-specified resource ids.
...
-for example a "reboot" operation on a VM that is in the process of rebooting, then the service should
-accept the request. The status field of the resource should then report the completion status of _both_
-operations.
+The operation is initiated with a POST operation and the operation path ends in `:action`.
 
-Note: The RELO pattern should not be used in cases where the completion status of individual operations
-may be important to users, as opposed to simply learning that an operation of the type they requested
-(e.g. create a resource with a specific name) has successfully completed.
+```text
+POST /<service-or-resource-url>:action?api-version=2022-05-01
```
markweitzel: Are we suggesting that what goes at the end is the actual word *action*, instead of an "action verb"?

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,206 @@ cannot collide with a resource path that contains user-specified resource ids.
...
 
-4. The status monitor URL responds with information about the operation including its current status,
-which should be represented as one of a fixed set of string values in a field named `status`.
-If the operation is still being processed, the status field will contain a "non-terminal" value, like `Processing`.
+A special case of long-running operation that occurs often is a PUT operation to create or replace a resource
+that involves some additional long-running processing.
+One example is a resource requires physical resources (e.g. servers) to be "provisioned" to make the resource functional.
+In this case, the request may contain an `operation-id` header that the service will use as
+the ID of the status monitor created for the operation.
+
```
markweitzel, line 267: The text refers to `operation-id` but the example has `operation-Id`, capital I. We should be consistent.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,206 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+```
 
-3. After waiting at least the amount of time specified by the previous response's `Retry-after` header,
-the client issues a GET request to the status monitor URL.
+### PUT with additional long-running processing
 
-4. The status monitor URL responds with information about the operation including its current status,
-which should be represented as one of a fixed set of string values in a field named `status`.
-If the operation is still being processed, the status field will contain a "non-terminal" value, like `Processing`.
+A special case of long-running operation that occurs often is a PUT operation to create or replace a resource
```
markweitzel, line 262: Re-reading this again,  it's not 100% clear when I would use this pattern. Or maybe said a slightly different way... If someone is reading this, how will they recognize that they have this pattern themselves? Maybe we can give an example from a couple of different domains. Some ideas...

"You may encounter this pattern when you have a one resource that either represents a physical resource. For example, in the Internet of Things domain, it is common to have a logical representation of a physical device. You can provision the logical device, but it is dependent on its real world counterpart to be installed, connected, and turned on. Similarly, you may have a resource that acts as a 'factory', creating other resources. An example of this could be a model used to train a cognitive service. The model itself needs to be created, but it will also produce results, e.g. a document with analyzed text.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,206 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+If the additional processing failed, the service may delete the original resource if it is not usable in this state.
+
+### Long-running delete operation
+
+A long-running delete operation follows the general pattern of a long-running operation --
+it returns a `202 Accepted` with a status monitor which the client uses to determine the outcome of the delete.
+
+The resource being deleted should remain visible (returned from a GET) until the delete operation completes successfully.
+
+When the delete operation completes successfully, a client must be able to create new resource with same name without conflicts.
```
markweitzel, line 327: The *must* in here feels a bit strong. If I was using a guid, must I really be able to create a new thing with an old guid?

azure/Guidelines.md
```diff
@@ -757,68 +761,92 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
 
+:no_entry: **DO NOT** implement PATCH as an LRO.  If LRO update is required it should be implemented with POST.
+
 In rare instances where an operation may take a _very long_ time to complete, e.g. longer than 15 minutes,
 it may be better to expose this as a first class resource of the API rather than as an operation on another resource.
 
-There are two basic patterns that can be used for long-running operations:
-1. Resource-based long-running operations (RELO)
-2. Long-running operations with status monitor
+There are two basic patterns for long-running operations in Azure. The first pattern is used for a POST and DELETE
```
markweitzel, line 777: It would be valuable if we could agree on a name for these two patterns. I'm thinking of the [Gang of Four](https://www.amazon.com/Design-Patterns-Elements-Reusable-Object-Oriented/dp/0201633612/ref=sr_1_1?keywords=9780321700698&linkCode=qs&qid=1654610601&s=books&sr=1-1) where they name things like, `flywheel`, `state` etc. Although it sounds simple, I think this will help us start to develop a vocabulary in our API reviews.

azure/Guidelines.md
```diff
@@ -757,68 +761,92 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
 operation is being processed. The client obtains the outcome of the operation at some later time
 through another API call.
 See the [Long Running Operations section](./ConsiderationsForServiceDesign.md#long-running-operations) in
-Considerations for Service Design for an introduction to the design of long running operations.
+Considerations for Service Design for an introduction to the design of long-running operations.
 
 :white_check_mark: **DO** implement an operation as an LRO if the 99th percentile response time is greater than 1s.
 
+:no_entry: **DO NOT** implement PATCH as an LRO.  If LRO update is required it should be implemented with POST.
```
```suggestion
:no_entry: **DO NOT** implement PATCH as an LRO.  If LRO update is required it must be implemented with POST.
```
Let's have the language match--DO NOT = MUST (as compared to the weaker 'should')

azure/Guidelines.md
```diff
@@ -757,68 +761,92 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
+
 In rare instances where an operation may take a _very long_ time to complete, e.g. longer than 15 minutes,
 it may be better to expose this as a first class resource of the API rather than as an operation on another resource.
 
-There are two basic patterns that can be used for long-running operations:
-1. Resource-based long-running operations (RELO)
-2. Long-running operations with status monitor
+There are two basic patterns for long-running operations in Azure. The first pattern is used for a POST and DELETE
+operations that initiate the LRO. These return a `202 Accepted` response with a JSON status monitor in the response body.
+The second pattern applies only in the case of a PUT operation to create a resource that also involves additional long-running processing.
```
markweitzel, line 779:
```suggestion
The second pattern applies only in the case of a PUT operation to create a resource that also involves additional long-running processing. For guidance on when to use a specific pattern, please refer to [Considerations for Service Design, Long Running Operations](./ConsiderationsForServiceDesign.md#long-running-operations)
```

azure/Guidelines.md
```diff
@@ -757,68 +761,92 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
-:white_check_mark: **DO** return the same value for **both** headers.
+:white_check_mark: **DO** return a status monitor in the response body that conforms with the following structure:
 
-:white_check_mark: **DO** look for **both** headers in client code, preferring the `Operation-Location` header.
+**OperationStatus** : Object
+
+Property | Type        | Required | Description
+-------- | ----------- | :------: | -----------
+`id`     | string      | true     | The unique id of the operation
+other    |             | true     | Other values needed for a GET request to the status monitor (e.g. location)
```
markweitzel: Wondering if we should be more clear about this, e.g. if it should be an object, e.g. `additionalProperties`. Does this help with breaking changes, e.g. if I need to add something in the future?

markweitzel (commented):
Re-reading this again--it all looks great. I had a couple of minor thoughts along the way that we can discuss. Overall, it is a significant improvement to the existing guidance and I'm excited to formalize this in our guidelines.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -1,9 +1,12 @@
 # Considerations for Service Design
 
+<!-- cspell:ignore autorest, etag, idempotency -->
+
 ## History
 
 | Date        | Notes                                                          |
 | ----------- | -------------------------------------------------------------- |
+| 2022-May-20 | Update guidance on long-running operations                     |
```
tg-msft: _(Nit - We've rolled over to June already)_

azure/Guidelines.md
```diff
@@ -757,68 +761,92 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
+:white_check_mark: **DO** generate an ID (typically a GUID) for the status monitor if the `Operation-Id` header was not passed by the client.
 
-:white_check_mark: **DO** retain the status monitor resource for some documented period of time (at least 24 hours) after the operation completes.
+:white_check_mark: **DO** fail a request with a `400-BadRequest` if the `Operation-Id` header that matches an outstanding operation unless the request is identical to the prior request (a retry scenario).
+
+:white_check_mark: **DO** perform as much validation as practical when initiating the operation to alert clients of errors early.
+
+:white_check_mark: **DO** return a `201-Created` status code for create or `200-OK` for replace from the initial request with a representation of the resource if the resource was created successfully.
+
+:white_check_mark: **DO** include an `Operation-Id` header in the response with the ID of the status monitor for the operation.
```
tg-msft, line 824: Many of these are the same.  Would it be more helpful to aggregate and only call out the differences based on verb (i.e., how much do you want to bet someone will miss the 201 vs. 202 because the first few guidelines were identical so their brain short circuited the rest)?

azure/Guidelines.md
tg-msft: Does it mean literally `additionalProperties: true`?  Or do we mean extra well defined values?  I'd assume extra well defined values?

azure/Guidelines.md
```diff
@@ -757,68 +761,92 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
-:white_check_mark: **DO** look for **both** headers in client code, preferring the `Operation-Location` header.
+**OperationStatus** : Object
+
+Property | Type        | Required | Description
+-------- | ----------- | :------: | -----------
+`id`     | string      | true     | The unique id of the operation
+other    |             | true     | Other values needed for a GET request to the status monitor (e.g. location)
+`status` | string      | true     | enum that includes terminal values "Succeeded", "Failed", "Canceled"
+`error`  | ErrorDetail |          | Error object that describes the error when status is "Failed"
+`result` | object      |          | Only for POST action-type LRO, the results of the operation when completed successfully
```
tg-msft, line 849: How does this work for partial success in batched APIs?  Should we call out that those all go in the `result` and `error` is only for entire batch failure?

azure/Guidelines.md
```diff
@@ -757,68 +761,92 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
+
+:white_check_mark: **DO** include a `Retry-After` header in the response to GET requests to the status monitor if the operation is not complete. The value of this header should be an integer number of seconds to wait before making the next request to the status monitor.
+
+:white_check_mark: **DO** include the `result` property (if any) in the status monitor for a POST action-type long-running operation when the operation completes successfully.
+
+:no_entry: **DO NOT** include a `result` property in the status monitor for a long-running operation that is not a POST action-type long-running operation.
+
+:white_check_mark: **DO** retain the status monitor resource for some documented period of time (at least 24 hours) after the operation completes.
```
```suggestion
:white_check_mark: **DO** retain the status monitor resource for some publicly documented period of time (at least 24 hours) after the operation completes.
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,206 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+If the status is `Succeeded` and the LRO is an Action operation, the operation results will be returned in the `results` field of the status monitor.
+If the status is `Succeeded` and the LRO is an operation on a resource, the client can perform a GET on the resource
+to observe the result of the operation if desired.
 
-2. The resource validates the request and initiates the operation processing.
-It sends a response to client with a `200-OK` HTTP status code (or `201-Created` if the operation
-is a create operation) and a representation of the resource where the `status` field is set
-to a value indicating that the operation processing has been started.
+6. There may be some cases where a long-running operation can be completed before the response to the initial request.
+In these cases, the operation should still return a `202 Accepted` with the `status` property set to the appropriate terminal state.
```
tg-msft, line 207: "and no retry-after header"

azure/ConsiderationsForServiceDesign.md
tg-msft: +1 for something like `<action>` or `<action-name>` if that's the templating style we're using everywhere

azure/ConsiderationsForServiceDesign.md
```diff
@@ -141,94 +144,206 @@ cannot collide with a resource path that contains user-specified resource ids.
...
+A successful response to a control operation should be a `200 OK` with a representation of the status monitor.
+
+```text
+HTTP/1.1 200 OK 
+ 
+{
+   "id": "22",
+   "status": "Canceled"
+}
+```
```
tg-msft, line 347: Do we want prescriptive guidance on the path structure for listing all active operations?  Either globally or per operation?  Or is that coming in a future update?

tg-msft (approved):
Looks great

azure/ConsiderationsForServiceDesign.md
mikekistler: Our guidance will be to avoid long-running cancel and we'll see if that works.

azure/ConsiderationsForServiceDesign.md
mikekistler: We'll watch for patterns in edge cases in API reviews and feed those back into these guidelines.

azure/ConsiderationsForServiceDesign.md
mikekistler: We should leave Retry-After off of the response to the operation that initiates an LRO.  Then the client would initiate the get on the status monitor immediately, and the status monitor would then give the Retry-After in its response.

azure/Guidelines.md
mikekistler: I think we should cover this when we add guidance on batch operations.

weidongxu-microsoft:
One question, do we still require `x-ms-long-running-operation` on the API?

azure/Guidelines.md
```diff
@@ -757,68 +761,91 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
+:warning: **YOU SHOULD NOT** return any other `2xx` status code from the initial request of an LRO -- return `202-Accepted` and a status monitor even if processing was completed before the initiating request returns.
 
-#### Resource-based long-running operations
+:ballot_box_with_check: **YOU SHOULD** include an `Operation-Location` header in the response with the absolute URL of the status monitor for the operation, but do not include an api-version query parameter.
 
-Some common situations where the RELO pattern should be used:
-1. A "create" operation (PUT, PATCH, or POST) for a resource where the basic structure of the resource is created immediately and includes a status field that indicates when the create has completed, e.g. "provisioning" -> "active".
-2. An action operation for a resource where both the initiation of the action and the completion of the action cause a change to the "status" property of the resource.
+:white_check_mark: **DO** return a status monitor in the response body as described in [Status and results of long-running operations](#status-and-results-of-long-running-operations).
```
heaths: Looks like the heading changed to "Obtaining status and results of long-running operations" so the bookmark needs to change as well. To keep these stable, you can use A.name in the heading and just reference that everywhere:

```markdown
See [here](#there).

## Some heading <a name="there"></a>
```

heaths:
As being discussed in https://github.com/Azure/azure-sdk-for-net/pull/29221, do we need to define some way that the correct status monitor can be determined (through explicit reference or by strict convention) in order to properly codegen pollers?

/cc @annelo-msft

azure/Guidelines.md
```diff
@@ -757,68 +761,91 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
+:white_check_mark: **DO** perform as much validation as practical when initiating the operation to alert clients of errors early.
 
-:white_check_mark: **DO** use the RELO pattern when the operation is on a resource that contains a "status" property that can be used to obtain the outcome of the operation.
+:white_check_mark: **DO** return a `202-Accepted` status code from the request that initiates an LRO if the processing of the operation was successfully initiated (except for "PUT with additional processing" type LRO).
 
-:ballot_box_with_check: **YOU SHOULD** only use the status monitor LRO pattern when the RELO pattern is not applicable.
+:warning: **YOU SHOULD NOT** return any other `2xx` status code from the initial request of an LRO -- return `202-Accepted` and a status monitor even if processing was completed before the initiating request returns.
 
-#### Resource-based long-running operations
+:ballot_box_with_check: **YOU SHOULD** include an `Operation-Location` header in the response with the absolute URL of the status monitor for the operation, but do not include an api-version query parameter.
```
darrelmiller: How does a client know it is supposed to add back the api-version?

azure/Guidelines.md
```diff
@@ -757,68 +761,91 @@ implemented as a _long-running operation (LRO)_. This allows clients to continue
...
+For a PUT (create or replace) with additional long-running processing:
 
-:white_check_mark: **DO** define the "status" field of the resource as an enum with all the values it may contain including the "terminal" values "Succeeded", "Failed", and "Canceled". See [Enums & SDKs](#enums--sdks-client-libraries).
+:white_check_mark: **DO** allow the client to pass an `Operation-Id` header with a ID for the status monitor for the operation.
 
-:ballot_box_with_check: **YOU SHOULD** use the name `status` for the "status" field of the resource.
+:white_check_mark: **DO** generate an ID (typically a GUID) for the status monitor if the `Operation-Id` header was not passed by the client.
 
-#### Long-running operations with status monitor
+:white_check_mark: **DO** fail a request with a `400-BadRequest` if the `Operation-Id` header that matches an outstanding operation unless the request is identical to the prior request (a retry scenario).
```
darrelmiller: What happens if the operation completes before the retry is processed?  Is there an expectation that the service keep tombstones of operation-ids?

mikekistler: Yes.  I will clarify that. "outstanding" is the wrong word here.

mikekistler: I think the right word is "existing". We recommend that the operation status resource be retained for at least 24 hours before it is garbage collected. This gives a guarantee of idempotency for at least 24 hours, which should be sufficient for the vast majority of cases.

azure/Guidelines.md
```diff
@@ -698,7 +698,11 @@ Azure services need to change over time. However, when changing a service, there
 
 :white_check_mark: **DO** review any API changes with the Azure API Stewardship Board
 
-:white_check_mark: **DO** use an `api-version` query parameter with a `YYYY-MM-DD` date value, with a `-preview` suffix for a preview service.
+Clients specify the version of the API to be used in every request to the service, even requests to an `Operation-Location` or `nextLink` URL returned by the service.
```
heaths, line 701: Should this explanation go below the **DO** directive right below it? IIRC, I thought we put explanations after the requirement.

heaths (approved):

azure/Guidelines.md
mikekistler: I would be okay to move it, but I think this sentence "provides context" for the guidelines below rather than "explaining" them. That's why I put it above.

heaths: Okay. Just a nit. Works for me either way. :)

PR #361, opened 2022-05-25
Create subsets.md

ghost:
Moved *Modeling collection subsets* content from wiki page to GitHub

### DIFF
--- a/graph/patterns/subsets.md
+++ b/graph/patterns/subsets.md
@@ -6,17 +6,17 @@ Microsoft Graph API Design Pattern
 
 ## Problem
 
-A common pattern is to apply a policy or state to a collection of resources. With this, there also comes the question of how to model cases where we want to apply to 'all' or 'none' without having to special case these values within the collection set or introduce cross-property dependencies. Likewise, we'd like to model it in a way where it is easy to understand and interpret usage from just looking at the schema.
+A common pattern is to apply a policy or state to a collection of resources. With this, there also comes the question of how to model cases where we want to apply to `all` or `none` without having to special case these values within the collection set or introduce cross-property dependencies. Likewise, we'd like to model it in a way where it is easy to understand and interpret usage from just looking at the schema.
 
 An example is where you have a policy that you need to be able to apply to users in an organization. You might want to support the default **None**, enablement for **All**, or enablement for **Select** users where you only grant it to a few users.
 
-Existing patterns for this either have special-cased 'strings' or have tightly coupled dependencies between two independent properties. Neither is intuitive, both require reading documentation, and neither can be inferred from the schema or within client libraries.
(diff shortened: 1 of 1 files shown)


ghost:
@OlgaPodo I moved the content from the wiki and did an edit pass. If you're okay with it, let me know and either I can merge my branch into vNext or you can. Thanks.

OlgaPodo (approved):

ghost:
I've finished any updates I'm going to make to this page. It appears that you might like others to review it, so please merge it when you're ready.

graph/patterns/subsets.md
```diff
@@ -0,0 +1,173 @@
+# Modeling collection subsets
+
+Microsoft Graph API Design Pattern
+
+*The modeling collection subsets pattern is the modeling state associated to a collection that may include all instances, an included subset, an excluded subset, no instances, or any combinations of the preceding items.*
+
+## Problem
+
+A common pattern is to apply a policy or state to a collection of objects. With this, there also comes the question of how to model cases where we want to apply to 'all' or 'none' without having to special case these values within the collection set or introduce cross-property dependencies. Likewise, we'd like to model it in a way where it is easy to understand and interpret usage from just looking at the schema.
```
gareth-a-jones:
```suggestion
A common pattern is to apply a policy or state to a collection of resources. With this, there also comes the question of how to model cases where we want to apply to 'all' or 'none' without having to special case these values within the collection set or introduce cross-property dependencies. Likewise, we'd like to model it in a way where it is easy to understand and interpret usage from just looking at the schema.
```

graph/patterns/subsets.md
```diff
@@ -0,0 +1,173 @@
...
+    <ComplexType Name="excludedMembership" BaseType="graph.membership">
+      <Property Name="members" Type="Collection(Edm.String)"/>
+    </ComplexType>
+```
+
+Be aware that the name values and types in the preceding examples are just examples and can be replaced with your scenario equivalent values. For example, types don't really need to be memberships. The collection doesn't have to be a collection at all; it can be singular and doesn't have to be a string.
+
+```xml
+    <ComplexType Name="<base-type>" IsAbstract="true">
+      <Property Name="<enum-name-for-type>" Type="graph.<enum-type>"/>
```
gareth-a-jones: I know this wasn't in the original doc but I think we should standardize the relationship between the name of the abstract base type and the name of the enumeration type.

ghost: @garethj-msft Is there a specific change that you wanted here?

ghost:
@darrelmiller , @dkershaw10 ---

If you have comments you'd like to make on this topic, please let us know today. Otherwise, we will go ahead and publish it.  This topic was on the wiki, and we're migrating it to GitHub.

graph/patterns/subsets.md
OlgaPodo: Agree. @garethj-msft what do you think about Base suffix, ie membership, membershipBase?

ghost:
Olga, this is ready for you to merge.

OlgaPodo (approved):

PR #363, opened 2022-05-26
Improved alternate key pattern content.

gareth-a-jones:
Updated the content for clarity, directness, minor corrections on the scope of key uniqueness, fuller error examples, and being specific with guidance against multi-part alternate keys.
Also introduced the concept of key parentheses for those Graph users who've never encountered them, given slashes are the known public format.

### DIFF
--- a/graph/patterns/alternate-key.md
+++ b/graph/patterns/alternate-key.md
@@ -2,29 +2,51 @@
 
 Microsoft Graph API Design Pattern
 
-*The alternate key pattern provides the ability to query for a single, specific resource identifiable through an alternative set of properties that is not its primary key.*
+*The alternate key pattern provides the ability to query for a single, specific resource identifiable via one of an alternative set of properties that is not its primary key.*
 
 ## Problem
 
-The resources exposed in Microsoft Graph are identified through a primary key, which guarantees uniqueness inside the same resource type. Often though, that same resource can also be uniquely identified by an alternative, more convenient property (or set of properties) that provides a better developer experience.
+The resources exposed in Microsoft Graph are identified through a primary key, which guarantees uniqueness inside the same resource collection. Often though, that same resource can also be uniquely identified by an alternative, more convenient property that provides a better developer experience.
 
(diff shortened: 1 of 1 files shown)


graph/patterns/alternate-key.md
```diff
@@ -2,29 +2,51 @@
 
 Microsoft Graph API Design Pattern
 
-*The alternate key pattern provides the ability to query for a single, specific resource identifiable through an alternative set of properties that is not its primary key.*
+*The alternate key pattern provides the ability to query for a single, specific resource identifiable via one of an alternative set of properties that is not its primary key.*
 
 ## Problem
 
-The resources exposed in Microsoft Graph are identified through a primary key, which guarantees uniqueness inside the same resource type. Often though, that same resource can also be uniquely identified by an alternative, more convenient property (or set of properties) that provides a better developer experience.
+The resources exposed in Microsoft Graph are identified through a primary key, which guarantees uniqueness inside the same resource collection. Often though, that same resource can also be uniquely identified by an alternative, more convenient property that provides a better developer experience.
```
XVincentX, line 9: I am assuming composite keys are not a thing in MS Graph?

graph/patterns/alternate-key.md
```diff
@@ -2,29 +2,51 @@
...
 
+```http
 https://graph.microsoft.com/v1.0/users(email='bob@contoso.com') Retrieves the employee with the email matching `bob@contoso.com`.
+```
+
+In the same way as requesting a resource via the canonical key, if a resource cannot be located that matches the alternate key, then a 404 must be returned.
+
+> **Note:** When requesting a resource via alternate keys, the simple slash-delimited URL style does not work.
+
+> **Note:** Do not use multi-part alternate keys.   Feedback has been that customers find multi-part keys confusing.
```
XVincentX, line 44: What is a multi part key?

XVincentX (approved):
Great improvements, thanks!

graph/patterns/alternate-key.md
gareth-a-jones: correct

graph/patterns/alternate-key.md
gareth-a-jones: composite key

PR #365, opened 2022-05-31
Document the Default Properties pattern

PetRich-MSFT:


XVincentX (commented):
I am confused when this pattern should be used VS handling default values in the workload. 

Why would I do this at the AGS level VS my workload? Is there any scenario (like a shared type, for instance) where this would make sense? 

If so, it should be stated here so that I get what is the value of moving this to AGS.

graph/patterns/default-properties.md
```diff
@@ -0,0 +1,154 @@
...
+
+--------
+
+This pattern should be used when an API producer has a requirement to omit a property from an API response unless the API consumer explicitly requests the property.
+
+## Issues and Considerations
+
+--------
+
+- API Consumers should clearly document when a property is not returned by default in their API documentation, as at present the public MS Graph metadata does not indicate whether a property is a default property.
```
OlgaPodo, line 55: I think it should be API producer.

graph/patterns/default-properties.md
```diff
@@ -0,0 +1,154 @@
...
+- API producers should not mix the use of `ags:Default="false"` and `ags:Default="true"` annotations, For clarity it is recommended to only use the `ags:Default="false"` annotation.
+- Unless explicitly marked as not supporting `$filter` and/or `$orderby` a non-default property can be used in a `$filter` or `$orderby` clause even if the property is not included in a `$select` clause
```
OlgaPodo, line 57: I think that this consideration contradicts the example above

OlgaPodo:
This feature is specific for the gateway implementation and is moved to the internal wiki for reference.

PR #366, opened 2022-06-01
Update subtypes.md

gareth-a-jones:
Clean up language a little to refer to resources rather than objects and drop some unnecessary references to OData.   Also note that base types do not need to be abstract.

### DIFF
--- a/graph/patterns/subtypes.md
+++ b/graph/patterns/subtypes.md
@@ -13,8 +13,8 @@ of new variants without breaking changes.
 
 ## Solution
 
-API designers may use a *type hierarchy*, where there is one base
-type (which may be abstract) with a few shared properties representing the common concept and one
+API designers might use a *type hierarchy*, where there is one base
+type (which might be abstract) with a few shared properties representing the common concept and one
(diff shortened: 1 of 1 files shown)


OlgaPodo (approved):
@garethj-msft Thank you for the corrections!

graph/patterns/subtypes.md
```diff
@@ -2,29 +2,30 @@
...
 that can be further grouped into *mutually exclusive variants* with specific
 properties and behaviors. The API design should be evolvable and allow the addition
 of new variants without breaking changes.
 
 ## Solution
 
-API designers can use OData *type hierarchy*, where there is one abstract base
-type with a few shared properties representing the common concept and one
-subtype for each variant of the entity. In the hierarchy, the interdependencies of properties, that is, which properties are relevant for which variants, is fully captured in metadata, and client code can potentially leverage that to construct and/or validate requests.
+API designers may use a *type hierarchy*, where there is one base
```
ghost: Please update all instances of "may" with "might". "May" implies permission, while "might" implies possibility, which is what you mean here.

PR #376, opened 2022-07-03
Add guidance on including api-version in nextLink

### LINKED ISSUE #371: Document whether service should include api-version in nextLink or Operation-Location
The Azure SDK arch board has decided SDKs will handle `api-version` in `nextLink` and `Operation-Location` as described in this issue: https://github.com/Azure/azure-sdk/issues/3817#issuecomment-1033226508

However, I believe it is still undocumented what services should actually implement, whether these should include `api-version` or not.  This issue tracks adding this guidance to the REST API guidelines.

mikekistler:
This PR adds explicit guidance on whether `api-version` and similar required query parameters should be included in the value of `nextLink` in pageable responses.

I also added a brief overview of pagination in the Considerations for Service Design because this is an important API design pattern that was missing from that document.

### DIFF
--- a/azure/ConsiderationsForServiceDesign.md
+++ b/azure/ConsiderationsForServiceDesign.md
@@ -394,6 +394,8 @@ exist in the collection.
 The value of the `nextLink` property should be an opaque absolute URL that will return the next page of results.
 The absence of a `nextLink` property means that no additional pages are available.
 Since `nextLink` is an opaque URL it should include any query parameters required by the service, including `api-version`.
+The service should honor a request to a URL derived from `nextLink` by replacing the value for the `apl-version` query parameter
+with a different but valid api version. The service may reject the request if any other element of `nextLink` was modified.
 
 The service determines how many items to include in the response and may choose a different number for different collections and even for different pages of the same collection.
 An operation may allow the client to specify a maximum number of items in a response with an optional `maxpagesize` parameter.
--- a/azure/ConsiderationsForServiceDesign.md
+++ b/azure/ConsiderationsForServiceDesign.md
@@ -376,7 +376,7 @@ Finally, write sample code for your service's workflow and add the code you'd wa
 
 ## Pagination
 
-Operations that return a collection of resources may need should support pagination.
+Operations that return a collection of resources must consider pagination.
 There are hard limits to the payload size of HTTP responses, and when the size of a collection or the resources themselves
 can grow arbitrarily large there is the risk of exceeding this limit if the operation does not support pagination.
 Further, adding support for pagination is a breaking change so it should be supported in the initial GA of the service
@@ -385,6 +385,7 @@ if there is any possibility that it will eventually be needed.
 There are two forms of pagination that MAY be supported by RESTful APIs.
 Server-driven paging mitigates against denial-of-service attacks by forcibly paginating a request over multiple response payloads.
 Client-driven paging enables clients to request only the number of resources that it can use at a given time.
+Services should almost always support server-driven paging and may optionally support client-driven paging.
 
 ### Server-driven paging
 
--- a/azure/ConsiderationsForServiceDesign.md
+++ b/azure/ConsiderationsForServiceDesign.md
@@ -374,6 +374,37 @@ All responses should include the `x-ms-request-id` header with a unique id for t
 
 Finally, write sample code for your service's workflow and add the code you'd want customers using to gracefully recover from errors.  Is it actually graceful?  Is it something you'd be comfortable asking most customers to write?  We also highly encourage reaching out to customers during private preview and asking them for code they've written against your service.  Their error handling might match your expectations, you might find a strong need for better documentation, or you might find important opportunities to improve the errors you're returning.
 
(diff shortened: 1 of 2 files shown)


azure/ConsiderationsForServiceDesign.md
```diff
@@ -258,6 +258,37 @@ All responses should include the `x-ms-request-id` header with a unique id for t
 
 Finally, write sample code for your service's workflow and add the code you'd want customers using to gracefully recover from errors.  Is it actually graceful?  Is it something you'd be comfortable asking most customers to write?  We also highly encourage reaching out to customers during private preview and asking them for code they've written against your service.  Their error handling might match your expectations, you might find a strong need for better documentation, or you might find important opportunities to improve the errors you're returning.
 
+## Pagination
+
+Operations that return a collection of resources may need should support pagination.
```
```suggestion
Operations that return a collection of resources must consider pagination.
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -258,6 +258,37 @@ All responses should include the `x-ms-request-id` header with a unique id for t
...
+There are hard limits to the payload size of HTTP responses, and when the size of a collection or the resources themselves
+can grow arbitrarily large there is the risk of exceeding this limit if the operation does not support pagination.
+Further, adding support for pagination is a breaking change so it should be supported in the initial GA of the service
+if there is any possibility that it will eventually be needed.
+
+There are two forms of pagination that MAY be supported by RESTful APIs.
+Server-driven paging mitigates against denial-of-service attacks by forcibly paginating a request over multiple response payloads.
+Client-driven paging enables clients to request only the number of resources that it can use at a given time.
+
```
JeffreyRichter, line 389: Services almost always want to support server-driven paging and can optionally support client-driven paging as well.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -258,6 +258,37 @@ All responses should include the `x-ms-request-id` header with a unique id for t
...
+### Server-driven paging
+
+In server-driven paging, the service includes a `nextLink` property in the response to indicate that additional elements
+exist in the collection.
+The value of the `nextLink` property should be an opaque absolute URL that will return the next page of results.
```
JeffreyRichter, line 394: Can we say "practically opaque" or "almost opaque" or maybe "semi-opaque" URL because it is not completely opaque.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -258,6 +258,37 @@ All responses should include the `x-ms-request-id` header with a unique id for t
...
+The absence of a `nextLink` property means that no additional pages are available.
+Since `nextLink` is an opaque URL it should include any query parameters required by the service, including `api-version`.
```
JeffreyRichter, line 396: Do you not want to say anything about clients replacing the api-version query parameter? I think we should so that service teams know what the expected client usage is.

JeffreyRichter (approved):

azure/ConsiderationsForServiceDesign.md
mikekistler, line 389:
```suggestion
Services should almost always support server-driven paging and may optionally support client-driven paging.

```

azure/ConsiderationsForServiceDesign.md
mikekistler: We probably should say something about the service behavior when a client changes _something_ in the nextLink URL. We could say:
> The service should honor a request to a URL derived from `nextLink` by replacing the value for the `apl-version` query parameter with a different but valid api version.  The service may reject the request if any other element of `nextLink` was modified.

What do you think of that?

JeffreyRichter: The service MAY reject the nextLink request if any element of nextLink was modified by the client except for the api-version query parameter as the client should change this to reflect the version that it expects. Of course, the service must reject the request if the client passes an api-version value not supported by the service.

mikekistler: I said "should" rather than "must" because I think there may be cases where changing the api-version creates an invalid request.  For example, suppose a service adds support for "filter" on a pageable GET operation in V2. A V2 client makes a request to this endpoint with `filter=something`. The `nextLink` in the response has `?filter=something&continuationToken=xxxyyy&api-version=v2`.  Then the client modifies the URL to have `api-version=v1` -- which does not support `filter`. How could we allow the service to return a successful response for this?

JeffreyRichter: We're designing our SDKs to always replace the api-version. How would this work with our SDKs? If the filter were embedded in the continuationToken, then the service could make it work transparently to the client. Alternatively, the V1-service could support the V2-filter in a non-breaking way. The end-result is that the service HAS TO make this work transparently to the client via some opaque thing (like a continuationToken or the opaque part of the URL/query params).

This is more reason to prefer continuationTokens over nextLink. Weren't you going to write a proposal for this?

markweitzel (approved):
LGTM. Thank you.

PR #378, opened 2022-07-13
Operations pattern

OlgaPodo:


darrelmiller:
https://github.com/microsoft/api-guidelines/blob/27aa5a2983d162edf8f4a58360019d4efedbb5dd/graph/patterns/operations.md?plain=1#L5
Operations do not necessarily need to affect multiple resources.  e.g. Creating a resource via POSTing to a collection may be a LRO that only creates a single resource.

darrelmiller:
I'm confused as to why there is a suggestion that actions/functions are related to performing operations on multiple resources.  I don't see any correlation to how many resources are affected.

Actions are implemented via an unsafe data processing resource that where the behaviour does not naturally map to the native HTTP methods and for usability reasons the API designer wants to expose that behaviour as a first class resource.  e.g. MarkAsUnread, SendMail

Functions are a way to create read-only resources that are often some kind of convenience composite of other resources.

as we discussed on 07/27/22 we will elevate resource definition for HTTP level to Domain model or OData definition

PetRich-MSFT:
https://github.com/microsoft/api-guidelines/blob/27aa5a2983d162edf8f4a58360019d4efedbb5dd/graph/patterns/operations.md?plain=1#L24

What is the scenario for a null binding parameter on an action/function?


@PetRich-MSFT  honestly i copied it from the wiki but i can imagine that NULL might signify ANY type/collection including recently discussed navigation properties.

OlgaPodo:
> 



> I'm confused as to why there is a suggestion that actions/functions are related to performing operations on multiple resources. I don't see any correlation to how many resources are affected.
> 
> Actions are implemented via an unsafe data processing resource that where the behaviour does not naturally map to the native HTTP methods and for usability reasons the API designer wants to expose that behaviour as a first class resource. e.g. MarkAsUnread, SendMail
> 
> Functions are a way to create read-only resources that are often some kind of convenience composite of other resources.

@darrelmiller  The intent is to emphasize that usually operations in opposite to HTTP Methods are applied to a set of resources where the set can't be expressed via a URL. I agree that quantity not a good differentiator, another point is side effects even though a function doesn't produce them. We need to provide clarity on what conditions are right to make behavior the first class citizen. IMO MarkAsUnread is a good candidate for PATCH but sendMail a good candidate for an action because it has side effects. While a function is a read-only resource the function result may not be an addressable resource

graph/patterns/operations.md
```diff
@@ -2,16 +2,16 @@
 
 Microsoft Graph API Design Pattern
 
-*The operations pattern provides the ability to model a change that impacts multiple resources and can't be effectively modeled by using HTTP methods.*
+*The operations pattern provides the ability to model a change that might impact multiple resources and can't be effectively modeled by using HTTP methods.*
 
 ## Problem
 
-Sometimes when modeling a complex business domain, API designers need to model a business operation that effects multiple resources and needs to be performed as a single unit. Modeling the operation via HTTP methods on each individual resource might be either ineffective or not reflect how it's processed by the backend service. In addition, the operation might produce observable side effects.
+Sometimes when modeling a complex business domain, API designers need to model a business operation that effects one or multiple resources and has additional semantic meaning that cannot be expressed by HTTP methods. Modeling the operation via HTTP methods on each individual resource might be either ineffective or expose internal implementation details. In addition, the operation might produce observable side effects.
```
corranrogue9: > expose internal implementation details

I have a feeling that workloads' use of actions is actually because the resources are actually leaking implementation details, i.e. that the resource boundaries are the same as internal team boundaries, and because of this there is a lack of cohesion within and between resources, requiring workloads to force that cohesion through an action. 

Probably nothing I just wrote is actionable, but it might be something to think about

corranrogue9: > ineffective 

What do we mean by "ineffective" here? Was that meant to be "inefficient"?

graph/patterns/operations.md
```diff
@@ -2,16 +2,16 @@
...
 
 ## Solution
 
-To address these use cases, API designers might use operational resources such as functions or actions.
-If the operation doesn't have any side effects and MUST return a single instance of a type or a collection of instances, then the designer SHOULD use the OData function; otherwise, the designer can model the operation as an action.
+To address these use cases, API designers can use operational resources such as functions or actions.
+If the operation doesn't have any side effects and MUST return a single instance of a type or a collection of instances, then the designer SHOULD use OData functions; otherwise, the designer can model the operation as an action.
```
corranrogue9:
```suggestion
If the operation doesn't have any side effects and MUST return a single instance of a type or a collection of instances, then the designer MUST use OData functions; otherwise, the designer can model the operation as an action.
```

I think that the guidance should be more strict around the use of functions. An operation that has no side-effects and can be modeled as a function should really need to be extraordinary to be modeled as an action instead in my opinion.

graph/patterns/operations.md
```diff
@@ -33,32 +34,34 @@ You can consider related patterns such as [long running operations](./long-runni
 
 - Microsoft Graph supports the use of optional parameters. The optional parameter annotation can be used instead of creating function or action overloads when unnecessary.
 
-- API designer **MUST** use POST to call operations on resources.
+- API designer **MUST** use POST to call actions on resources.
+- API designer **MUST** use GET to call functions on resources.
 
 - The addition of a new mandatory not-nullable parameter to an existing action or function is a breaking change and is not allowed without proper versioning.
```
corranrogue9:
```suggestion
- The addition of a new mandatory not-nullable parameter to an existing action or function is a breaking change and is not allowed without proper versioning that is in accordance with our [deprecation guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/graph/deprecation.md).
```

If there's something about versioning on parameters, I think we should add it to this document, but I think it's just our standard deprecation process

graph/patterns/operations.md
OlgaPodo: Conway's law in practice

OlgaPodo: yes it's inefficient

graph/patterns/operations.md
```diff
@@ -2,26 +2,34 @@
 
 Microsoft Graph API Design Pattern
 
-*The operations pattern provides the ability to model a change that impacts multiple resources and can't be effectively modeled by using HTTP methods.*
+*The operations pattern provides the ability to model a change that might impact multiple resources and can't be effectively modeled by using HTTP methods.*
 
 ## Problem
 
-Sometimes when modeling a complex business domain, API designers need to model a business operation that effects multiple resources and needs to be performed as a single unit. Modeling the operation via HTTP methods on each individual resource might be either ineffective or not reflect how it's processed by the backend service. In addition, the operation might produce observable side effects.
+Sometimes when modeling a complex business domain, API designers need to model a business operation that effects one or multiple resources and has additional semantic meaning that cannot be expressed by HTTP methods. Modeling the operation via HTTP methods on each individual resource might be either inefficient or expose internal implementation details. In addition, the operation might produce observable side effects.
```
XVincentX: I know people are going to hate me - but my response to this would be "Get your domain model straight" :)

Ok ok, I will shut up.

graph/patterns/operations.md
```diff
@@ -2,26 +2,34 @@
...
 
 ## Solution
 
-To address these use cases, API designers might use operational resources such as functions or actions.
-If the operation doesn't have any side effects and MUST return a single instance of a type or a collection of instances, then the designer SHOULD use the OData function; otherwise, the designer can model the operation as an action.
+To address these use cases, API designers can use operational resources such as functions or actions.
+If the operation doesn't have any side effects and MUST return a single instance of a type or a collection of instances, then the designer MUST use OData functions; otherwise, the designer can model the operation as an action.
```
XVincentX: I can see somebody having some confusion about this, and we should clarify operation vs (odata) function vs action

graph/patterns/operations.md
```diff
@@ -2,26 +2,34 @@
...
+- complex processing logic that shouldn't be exposed to the client
+- operation parameters might convey a restricted set of option (i.e., a report that can has to specify a date range)
+- the operation leverage some service-side data not exposed to (or easily retrieved in context by) the user.
+
 You can consider related patterns such as [long running operations](./long-running-operations.md) and [change tracking](./change-tracking.md).
 
 ## Issues and considerations
 
-- Microsoft Graph does NOT support unbound actions or functions. Bound actions and functions are invoked on resources matching the type of the binding parameter. The binding parameter can be of any type, and it MAY be Nullable. For Microsoft Graph, actions and functions must have the `isBound="true"` attribute. The first parameter is the binding parameter.
+- Microsoft Graph does NOT support unbound actions or functions. Bound actions and functions are invoked on resources matching the type of the binding parameter. The binding parameter can be of any type, and parameter value MAY be Nullable.
```
XVincentX: What is an unbound function? Somebody reading this for the first time might not get it (I did not)

graph/patterns/operations.md
```diff
@@ -33,32 +41,34 @@ You can consider related patterns such as [long running operations](./long-runni
 
 - Microsoft Graph supports the use of optional parameters. The optional parameter annotation can be used instead of creating function or action overloads when unnecessary.
 
-- API designer **MUST** use POST to call operations on resources.
+- API designer **MUST** use POST to call actions on resources.
+- API designer **MUST** use GET to call functions on resources.
 
-- The addition of a new mandatory not-nullable parameter to an existing action or function is a breaking change and is not allowed without proper versioning.
+- The addition of a new mandatory not-nullable parameter to an existing action or function is a breaking change and is not allowed without proper versioning that is in accordance with our [deprecation guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/graph/deprecation.md).
```
XVincentX, line 43: I am surprised we need to point this out.

graph/patterns/operations.md
OlgaPodo: actions and functions are operations or operational resources.

graph/patterns/operations.md
OlgaPodo: Line #32 provides a definition of the bound operations. I would think that unbound by definition would be what isn't bound. Do you think more explanation needed?

graph/patterns/operations.md
dkershaw10: For the last statement - why is an observable side-effect  an issue using regular HTTP methods?  Do our customers *really* know OData that well that they could say - "aha, this is not a regular HTTP method, it's a service action, so it's fine for it to have an observable side effect"?

graph/patterns/operations.md
dkershaw10: We have examples of operations that have no side-effects that are modelled as actions, purely because it's much simpler to pass a larger payload (like collections) as parameters with an action.  I'm not saying it's right or wrong, but there's definitely precedence.

Maybe we are better saying the designer SHOULD rather than MUST?

graph/patterns/operations.md
```diff
@@ -2,26 +2,34 @@
...
 
 ## Issues and considerations
 
-- Microsoft Graph does NOT support unbound actions or functions. Bound actions and functions are invoked on resources matching the type of the binding parameter. The binding parameter can be of any type, and it MAY be Nullable. For Microsoft Graph, actions and functions must have the `isBound="true"` attribute. The first parameter is the binding parameter.
+- Microsoft Graph does NOT support unbound actions or functions. Bound actions and functions are invoked on resources matching the type of the binding parameter. The binding parameter can be of any type, and parameter value MAY be Nullable. 
+For Microsoft Graph, actions and functions must have the `isBound="true"` attribute. The first parameter is the binding parameter.
 
 - Both actions and functions support overloading, meaning a schema might contain multiple actions or functions with the same name. The overload rules as per the OData [standard](http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_FunctionOverloads) apply when adding parameters to actions and functions.
```
dkershaw10, line 30: I think that's true for REST.  But not sure that overloading works on all of our SDK platforms.  Please check with @baywet

baywet: now that we switched to using a parameter set class for actions/functions in Java (almost 2 years ago now), I believe all of our SDKs support overloading. And all Kiota based SDKs have been designed with that requirement in mind. Thanks for double checking.

dkershaw10: That is awesome! Thanks for the update.

graph/patterns/operations.md
```diff
@@ -33,32 +41,34 @@ You can consider related patterns such as [long running operations](./long-runni
...
 - Microsoft Graph supports the use of optional parameters. The optional parameter annotation can be used instead of creating function or action overloads when unnecessary.
 
-- API designer **MUST** use POST to call operations on resources.
+- API designer **MUST** use POST to call actions on resources.
+- API designer **MUST** use GET to call functions on resources.
 
-- The addition of a new mandatory not-nullable parameter to an existing action or function is a breaking change and is not allowed without proper versioning.
+- The addition of a new mandatory not-nullable parameter to an existing action or function is a breaking change and is not allowed without proper versioning that is in accordance with our [deprecation guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/graph/deprecation.md).
 
 ## Example
```
dkershaw10: Nit: Examples

dkershaw10: I feel like this section is a bit incomplete.
It would be great to add to each of these:

1. A 1-2 sentence scenario explanation, with why it makes sense as a function/action
2. Show the actual request and response to go with the XML
3. Strip out the AGS annotations - they are confusing and not required for defining these functions/actions.

graph/patterns/operations.md
XVincentX: It's nitpicking, but I would rephrase this to define bound actions first, and then talk about unbound ones

graph/patterns/operations.md
OlgaPodo: Agree. will remove side effects because it is not a differentiator.

graph/patterns/operations.md
OlgaPodo: @dkershaw10 Do you think this is a good set of examples?

graph/patterns/operations.md
```diff
@@ -33,32 +37,112 @@ You can consider related patterns such as [long running operations](./long-runni
...
+```
+HTTP/1.1 202 Accepted
+
+    "cache-control": "private",
+    "client-request-id": "ca2d0416-a2c1-05af-df60-0921547a86e9",
+    "content-length": "0",
+    "request-id": "8b53016f-cc2b-4d9f-9818-bd6f0a5e3cd0"
+```
+
+`forward` operation is modeled as an asynchronous action bound to the Graph `event` entity type because the operation represents a complex business logic processed on the server side.
```
PetRich-MSFT: Nitpick: The text and JSON example talks about formwarding a message, however the CSDL example has action bound to event

PR #381, opened 2022-07-15
Navigation property pattern

darrelmiller:
- [x] Probably needs more examples of one-to-many
- [x] Missing discussion of containment, although I can't decide how important that is.


graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,185 @@
...
+*A navigation property is used to identify a relationship between resources.*
+
+## Problem
+--------
+
+It is often valuable to represent a relationship between resources in an API. This may be a many-to-one or a one-to-many relationship. 
+
+Relationships between resources are often implicitly represented by a property contained in one of the resources that identifies the other related resource. Usually that information returned in a representation as an id value and the property is named using a convention that identifies the target type of related resource. e.g. userId 
+
+In many-to-one relationships, for a client to access the related resource it must request the primary resource, read the id value of the related resource and then construct a URL to the related resource using the Id value and knowledge of how the property name maps to the related resource. This requires at least two round trips and requires the client know how to construct the URL to the related resource.
```
OlgaPodo: I believe phrase 'In many-to-one relationships' is not necessary because if relationships is defined just by ref ID then two round trips are needed. Also it creates a perception that for one-to-many relationships it's not the same.

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,185 @@
...
+e.g. /user/{userId}/manager  # many-to-one relationship
+     /user/{userId}/messages # one-to-many relationship
+
+Additionally, using the OData Expand query parameter, related entities can be transcluded into the primary entity so both can be retrieved in a single round trip.
+
+## Issues and Considerations
+-------------------------
+
+In the current Microsoft Graph implementation, support for navigation properties is limited to entities within the same backend service or the user entity.  
+
```
OlgaPodo, line 34: It seems that `user` entity violates this rule. may be it's just not recommended ?

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,185 @@
...
+Additionally, using the OData Expand query parameter, related entities can be transcluded into the primary entity so both can be retrieved in a single round trip.
+
+## Issues and Considerations
+-------------------------
+
+In the current Microsoft Graph implementation, support for navigation properties is limited to entities within the same backend service or the user entity.  
+ 
+Navigation properties defined within an entity are not returned when retreiving the representation of an entity.  
+
```
OlgaPodo, line 41: are not returned by default

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,185 @@
...
+One other use case is when child resources appear in a non-contained collection and there is a desire to access the canonical parent:
+
+`/me/pinnedChannels/{channelId}/team`
+
+## Example
+-------
+
+### Retrieving a related entity
+
```
OlgaPodo, line 95: I think we need to include CSDL representation.  Our devs often don't know how to describe a particular request/response

graph/patterns/navigation-property.md
OlgaPodo, line 39: "userSettings" entity also has navigation properties to other workloads. May be we should say that navigation properties to other workloads are not recommended?

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,185 @@
...
+------------------------
+
+### "has a" relationships
+
+The use of navigation properties is preferred over including an Id field to reference the related entity in a many-to-one relationship.  Id values require a client to make two round trips to retrieve the details of a related entity.  With a navigation property a client can retrieve a related entity in a single round trip. 
+
+Navigation properties are also useful when clients sometimes want to retrieve both the primary entity and the related entity in a single round trip.  The `expand` query parameter makes this possible.
+
+### "parent-child" relationships
+
```
OlgaPodo, line 73: IMO it's better to call containment relationships since parent-child in OOP is usually inheritance so all c# devs will be confused.

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,185 @@
...
+Create a new user that references an existing manager
+```http
+POST /users/{id}
+Content-Type: application/json
+
+{
+    "displayName": "Bob",
+    "manager@bind": "https://graph.microsoft.com/v1.0/users/{managerId}"
+}
```
OlgaPodo: Love this example. but probably user id in the URL isn't needed for POST?

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,185 @@
...
+
+Requiring two round trips to access this information is inefficient for some applications and the lack of a formally described relationship limits the ability for tooling to take advantage of the relationship to improve developer experience.
+
+## Solution
+--------
+
+Navigation properties are an [OData convention](https://docs.microsoft.com/en-us/odata/webapi/model-builder-untyped#navigation-property) that allows an API designer to describe a special kind of property in a model that references an related entity. In the HTTP API this property name translates to a path segment that can be appended to the URL of the primary resource in order to access a representation of the related resource. This prevents the client from needing to know any additional information on how to construct the URL to the related resource and the client does not need to retrieve the primary resource if it is only interested in the related resource.  It is the responsibility of the API implementation to determine the Id of the related resource and return the representation of the related entity.
```
XVincentX: I was about to suggest to use a link in the [official spec](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html) but surprisingly speaking there is no Navigation Property definition. The term is used without being defined. Is that on purpose? Or maybe it is in a different document?

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,185 @@
...
+
+e.g. /user/{userId}/manager  # many-to-one relationship
+     /user/{userId}/messages # one-to-many relationship
+
+Additionally, using the OData Expand query parameter, related entities can be transcluded into the primary entity so both can be retrieved in a single round trip.
+
+## Issues and Considerations
+-------------------------
+
+In the current Microsoft Graph implementation, support for navigation properties is limited to entities within the same backend service or the user entity.
```
XVincentX: I might be wrong, but I thought that AGS was able to fan out requests to the necessary workloads?

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,185 @@
...
+Navigation properties defined within an entity are not returned when retreiving the representation of an entity.  
+
+Implementing support for accessing the "$ref" of a navigation property allows a caller to return just the URL of related resource. e.g. `/user/23/manager/$ref`. This is useful when a client wishes to identity the related resource but doesn't need all of its properties.
+
+## When to Use this Pattern
+------------------------
+
+### "has a" relationships
+
+The use of navigation properties is preferred over including an Id field to reference the related entity in a many-to-one relationship.  Id values require a client to make two round trips to retrieve the details of a related entity.  With a navigation property a client can retrieve a related entity in a single round trip.
```
XVincentX, line 49: Would it make sense to also tell when it is appropriate *not* to put a navigation property, but rather leave an ID?

graph/patterns/navigation-property.md
XVincentX: I argue that parent child is not inheritance - and if devs get confused by that I argue it is their problem and they should get the terminology right. Inheritance is generalization/specialization, `is a` relationship

graph/patterns/navigation-property.md
XVincentX: Out of curiosity, would `POST /users/{id}/manager` work?

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,185 @@
...
+```
+
+Create a relationship between the user and the existing manager.
+
+```http
+PUT /users/{id}/manager/$ref
+Content-Type: application/json
+
+{
+    "@OData.Id": "https://graph.microsoft.com/v1.0/users/{managerId}"
```
XVincentX: Lowercase?

```suggestion
    "@odata.Id": "https://graph.microsoft.com/v1.0/users/{managerId}"
```

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,185 @@
...
+## Problem
+--------
+
+It is often valuable to represent a relationship between resources in an API. This may be a many-to-one or a one-to-many relationship. 
+
+Relationships between resources are often implicitly represented by a property contained in one of the resources that identifies the other related resource. Usually that information returned in a representation as an id value and the property is named using a convention that identifies the target type of related resource. e.g. userId 
+
+In many-to-one relationships, for a client to access the related resource it must request the primary resource, read the id value of the related resource and then construct a URL to the related resource using the Id value and knowledge of how the property name maps to the related resource. This requires at least two round trips and requires the client know how to construct the URL to the related resource.
+
+For both many-to-one and one-to-many relationships in order to retrieve information from both resources on both sides of the relationship requires at minimum two requests.
```
dkershaw10: We seem to be repeating the same info here.  Can this merge into line 14 with "For relationships, for a client..."
Or at worst "For both many-to-one and one-to-many relationships, for a client...

graph/patterns/navigation-property.md
dkershaw10: Yeah - this statement is incorrect.
For non-contained nav properties it's currently limited to within backend services or the target being an entity set.  

We're working on the target being a nested entity collection (as long as it's not nested under another entity collection).  So target could be /topLevelSegment/2ndLevelSegment/entityCollection
The limitations are covered in [this document](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/354352/Cross-workload-navigations)

graph/patterns/navigation-property.md
dkershaw10: See comment above.  We support x-workload navigations, and we're investing to support more scenarios, and do so without requiring any extra work for the target workload.  I don't think we should be discouraging workloads from doing this.

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,185 @@
...
+Additionally, using the OData Expand query parameter, related entities can be transcluded into the primary entity so both can be retrieved in a single round trip.
+
+## Issues and Considerations
+-------------------------
+
+In the current Microsoft Graph implementation, support for navigation properties is limited to entities within the same backend service or the user entity.  
+ 
+Navigation properties defined within an entity are not returned when retreiving the representation of an entity.  
+
+Implementing support for accessing the "$ref" of a navigation property allows a caller to return just the URL of related resource. e.g. `/user/23/manager/$ref`. This is useful when a client wishes to identity the related resource but doesn't need all of its properties.
```
dkershaw10:
```suggestion
Implementing support for accessing the "$ref" of a navigation property allows a caller to return just the URL of related resource. e.g. `/user/23/manager/$ref`. This is useful when a client wishes to identify the related resource but doesn't need all of its properties.
```

graph/patterns/navigation-property.md
dkershaw10, line 49: Why are we just talking about many-to-one relationships?
Why not one-to-many relationships?  We do support navigation property collections.

graph/patterns/navigation-property.md
dkershaw10: I'm OK with parent-child relationship.  I also like to point out the lifecycle implications.  The child entities do not have an independent lifecycle - when the parent is deleted, it's children are too.

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,185 @@
...
+
+{
+    "displayName": "Bob",
+    "manager@bind": "https://graph.microsoft.com/v1.0/users/{managerId}"
+}
+
+201 Created
+```
+
+Create a new user and the users manager and create a relationship between the two.
```
dkershaw10: Yeah - not sure this is supported by directory, but as a pattern this is sound.

graph/patterns/navigation-property.md
dkershaw10: `PUT /users/{id}/manager/$ref` would work to just update the manager.  It's a PUT because manager is a singleton.  I think Darrel has the example below.

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,185 @@
...
+### Creating an entity with a reference to a related entity
+
+Create a new user that references an existing manager
+```http
+POST /users/{id}
+Content-Type: application/json
+
+{
+    "displayName": "Bob",
+    "manager@bind": "https://graph.microsoft.com/v1.0/users/{managerId}"
```
dkershaw10: BTW the syntax isn't correct - it should be: "manager@odata.bind"
For OData 4.01 this got simplified so that the odata bind annotation is not needed.

```json
   "manager" { "@id": "https://graph.microsoft.com/v1.0/users/{managerId}" }
```

See https://docs.oasis-open.org/odata/odata/v4.01/os/part1-protocol/odata-v4.01-os-part1-protocol.html#sec_LinktoRelatedEntitiesWhenCreatinganE

graph/patterns/navigation-property.md
dkershaw10: Id needs to be lower case too
```suggestion
    "@odata.id": "https://graph.microsoft.com/v1.0/users/{managerId}"
```
Also in 4.01 the "odata" is not required.  Would need to check if AGS supports annotations without "odata".

dkershaw10 (changes requested):
Mainly looks good to me.  Some fixes needed.

graph/patterns/navigation-property.md
OlgaPodo: Thank you for the confirmation! Agree this is one of Graph benefits

graph/patterns/navigation-property.md
darrelmiller: It is defined in the CSDL spec here https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#_Toc38530365  I added a link to the document.

graph/patterns/navigation-property.md
darrelmiller: Changed the comment to be unspecific about what works and doesn't and provided suggestion to verify that a particular scenario is supported.

graph/patterns/navigation-property.md
darrelmiller: Fixed

graph/patterns/navigation-property.md
darrelmiller: There is a section on "parent-child" relationships.  Those are one-to-many.  I noticed that I don't have any examples of one-to-many relationships.  I will add some of those.

graph/patterns/navigation-property.md
darrelmiller: @dkershaw10   Except for when they don't.  You could have a one-to-many relationship where deleting the parent doesn't delete the children.  OneDrive Bundles is an example. Terms in TermSet is another. Perhaps we could argue that "parent-child" infers containsTarget=true, however, we have even stated that containsTarget doesn't define lifetime completely.  A non-contained collection could cause its children to be deleted when it is deleted.

I have tried to avoid bringing up the topic of lifetime because I think it deserves a section all to itself, but I am torn about the right way to explain this.

graph/patterns/navigation-property.md
darrelmiller: I added one and I include a link to the relevant section of the CSDL spec.

graph/patterns/navigation-property.md
darrelmiller: It may be wrong, but it works  :-)  I tested all the calls before adding them to the doc.  I updated the example to match what the spec says.

graph/patterns/navigation-property.md
darrelmiller: Ok, so maybe I didn't test this one.

graph/patterns/navigation-property.md
OlgaPodo: we are trying to avoid just ID because it requires to know a type of addressable resource to construct a graph URL. and an additional roundtrip

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,190 @@
...
+POST /users/{id}
+Content-Type: application/json
+
+{
+    "displayName": "Jim James",
+    "manager": {
+        "displayName": "Bob Boyce"
+    }
+}
+
```
OlgaPodo, line 159: this example doesn't work on my test tenant. probably the URL should be POST /users?

graph/patterns/navigation-property.md
XVincentX: So are we saying that there is no situation in which it would be preferable? I can't think of any, I was more wondering if @darrelmiller has some on top of his mind

graph/patterns/navigation-property.md
darrelmiller: Fixed

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,174 @@
...
+### "parent-child" relationships
+
+Resources that contain a parent Id property in a child resource can utilize a navigation property in the parent resource that is declared as a collection of child resources. If desirable, a parent navigation property can also be created in the child resource to the parent resource. This is usually not necessary as the parent URL is a subset of child resource URL. The main use of this would be when retrieving child resources and choosing to expand properties of the parent resource so that both can be retrieved in a single request.  
+
+`/invoice/{invoiceId}/items/{itemId}?expand=parentInvoice(select=invoiceDate,Customer)`
+
+### "child-parent" relationships
+
+One other use case is when child resources appear in a non-contained collection and there is a desire to access the canonical parent:
```
mikepizzo: You use "non-contained" here, but you don't use "contained" in the previous section, so it's unclear what this is contrasting with.  Also, in the previous section you call out the fact that a child can have a navigation to the parent, so unclear what this is adding.

Personally, I would include the term "contained" in the previous section in order to introduce the terminology, since that's what they will see when the go to the CSDL.

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,174 @@
...
+Implementing support for accessing the "$ref" of a navigation property allows a caller to return just the URL of related resource. e.g. `/user/23/manager/$ref`. This is useful when a client wishes to identify the related resource but doesn't need all of its properties.
+
+## When to Use this Pattern
+------------------------
+
+### "has a" relationships
+
+The use of navigation properties is preferred over including an Id field to reference the related entity in a many-to-one relationship.  Id values require a client to make two round trips to retrieve the details of a related entity.  With a navigation property a client can retrieve a related entity in a single round trip. 
+
+Navigation properties are also useful when clients sometimes want to retrieve both the primary entity and the related entity in a single round trip.  The `expand` query parameter makes this possible.
```
mikepizzo: should we say "...want to retrieve *properties of* both the primary entity and the related entity...", in order to emphasize that clients need not (and typically should not) return the full entities?

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,174 @@
...
+  "@odata.type": "#microsoft.graph.user",
+  "id": "6b3ee805-c449-46a8-aac8-8ff9cff5d213",
+  "displayName": "Bob Boyce"
+}
+```
+
+This navigation property could be described with the following CSDL: 
+```xml
+<EntityType Name="user">
+  <NavigationProperty Name="manager" Type="user" ContainsTarget="false" >
```
mikepizzo: false is the default for ContainsTarget. Do we want to include it anyway for clarity? If so, should we add a comment that it is the default and typically not returned (so that, when they look at the CSDL, they understand that nav props not marked with ContainsTarget are non-contained?)

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,174 @@
...
+GET /users/{id}/manager/$ref
+
+200 OK
+Content-Type: application/json
+
+{
+  "@odata.id": "https://graph.microsoft.com/v1.0/directoryObjects/6b3ee805-c449-46a8-aac8-8ff9cff5d213/Microsoft.DirectoryServices.User"
+}
+```
+Note: Currently the base URL returned in $ref results are incorrect.  In order to process these URLs the client will need to convert the URL to a Graph URL.
```
mikepizzo, line 127: It's really unfortunate that we have this bug in graph. I wonder if we could pick an example that wasn't broken.

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,174 @@
...
+
+`/me/pinnedChannels/{channelId}/team`
+
+## Example
+-------
+
+### Retrieving a related entity
+
+```http
+GET /users/{id}/manager
```
mikepizzo: I wonder if we should start including $select in all of our examples, to emphasize the point that request should always include select?

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,174 @@
...
+
+200 OK
+Content-Type: application/json
+
+{
+  "@odata.type": "#microsoft.graph.user",
```
mikepizzo: type is only required because, apparently, in MS graph one could be managed by a device (i.e., when AI gets to the point where our computers become sentient and take over the world).  Since your simplified CSDL below types manager as user, the type cast is not required in the response.

graph/patterns/navigation-property.md
mikepizzo: Type should be "graph.user"

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,174 @@
...
+### Creating an entity with a reference to a related entity
+
+Create a new user that references an existing manager
+```http
+POST /users
+Content-Type: application/json
+
+{
+    "displayName": "Bob",
+    "manager@odata.bind": "https://graph.microsoft.com/v1.0/users/{managerId}"
```
mikepizzo, line 157: At what point do we start migrating to the (much cleaner) 4.01 syntax:
{
    "displayName": "Bob",
    "manager": 
     {
          "@odata.id":"https://graph.microsoft.com/v1.0/users/{managerId}"
     }
}

or:

{
    "displayName": "Bob",
    "manager": 
     {
          "id":"{managerId}"
     }
}

mikepizzo (commented):
I think this is awesome!  Added some comments.

graph/patterns/navigation-property.md
darrelmiller: Let's discuss this in API council.

graph/patterns/navigation-property.md
darrelmiller: I documented it as a correct URL.  I'd rather have the set of examples be consistent around the manager navigation property.

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,174 @@
...
+}
+```
+
+This navigation property could be described with the following CSDL: 
+```xml
+<EntityType Name="user">
+  <NavigationProperty Name="manager" Type="user" ContainsTarget="false" >
+</EntityType>
+```
+
```
darrelmiller:
```suggestion
`ContainsTarget` is set to false for clarity, this is the default value when the attribute is omitted.
```

graph/patterns/navigation-property.md
darrelmiller: Done

darrelmiller:
```suggestion
  <NavigationProperty Name="manager" Type="graph.user" ContainsTarget="false" >
```

graph/patterns/navigation-property.md
darrelmiller:
```suggestion
```

graph/patterns/navigation-property.md
darrelmiller: I do not have any clear scenarios where we should avoid navigation properties.  Today there are some negative effects of navigation properties in SDKs, but that should be resolved via SDK work, not by avoiding navigation properties.

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,173 @@
+# Navigation Property
+
+Microsoft Graph API Design Pattern
+
+*A navigation property is used to identify a relationship between resources.*
+
+## Problem
+--------
+
+It is often valuable to represent a relationship between resources in an API. This may be a many-to-one or a one-to-many relationship.
```
PetRich-MSFT: it can also be 1-1 as well

graph/patterns/navigation-property.md
darrelmiller: We will change our guidance once WebAPI supports this.

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,208 @@
...
+Implementing support for accessing the "$ref" of a navigation property allows a caller to return just the URL of related resource. e.g. `/user/23/manager/$ref`. This is useful when a client wishes to identify the related resource but doesn't need all of its properties.
+
+## When to Use this Pattern
+------------------------
+
+### "Many-to-one" relationships  
+
+The use of navigation properties is preferred over including an Id field to reference the related entity in a many-to-one relationship.  Id values require a client to make two round trips to retrieve the details of a related entity.  With a navigation property a client can retrieve a related entity in a single round trip. 
+ 
+Many-to-one relationships are always non-contained relationships as the lifetime of the target cannot depend on the source.
```
OlgaPodo, line 51: Nit: Can we make "Many-to-one relationships are always non-contained relationships as the lifetime..." a little more precise for example "Many-to-one relationships are often modeled as non-contained relationships so that the lifetime..."?

graph/patterns/navigation-property.md
```diff
@@ -0,0 +1,208 @@
...
+    "manager@odata.bind": "https://graph.microsoft.com/v1.0/users/{managerId}"
+}
+
+204 No Content
+```
+
+Create a relationship between the user and the existing manager.
+
+```http
+PUT /users/{id}/manager/$ref
```
OlgaPodo: If we have two ways of achieving the same result what is our preference? What do we ask from our workloads to implement one or both?

graph/patterns/navigation-property.md
darrelmiller: Done.

graph/patterns/navigation-property.md
darrelmiller: Removed the example that requires supporting the $ref endpoint.

PR #383, opened 2022-07-26
Add lifting guidelines for Complex Types

XVincentX:


gareth-a-jones (changes requested):
IMO, we should not have this pattern. It's an antipattern for Graph.

At Microsoft, we should be extremely resistant to API changes for the sake of incremental improvement.

Breaking changes should bring *major* value to customers, and this lifting or grouping is adding minor value at the cost of pushing work onto our customers at some time in the next three years.

I strongly propose that we do not merge this pattern.

XVincentX:
Let's discuss at the next possible meeting. This was a response/decision made from my original argument, but pretty sure we can revisit and find another solution.

PR #387, opened 2022-08-18
Guidelines for 'azure-deprecating' response header

### LINKED ISSUE #386: Add guidance for deprecation
In working with service teams, we need to add a way for them to indicate an API or part of their service has been deprecated. The API case is quite straightforward. However the "part of the service" is referring to cognitive services where they may deprecate a model. 

We are looking at [draft deprecation header](https://datatracker.ietf.org/doc/html/draft-ietf-httpapi-deprecation-header) and 
[sunset header](https://www.rfc-editor.org/rfc/rfc8594) as prior art.

Current proposal is:
> "azure-deprecating": "<description>; <date>; <url>" 

Header: `azure-deprecating`
Value: string.
The value string should be a semi-colon delimited list that contains the following:
`description`: a human readable description of what is being deprecated
`date`: the target date that this will be deprecated. This should be expressed following the format in [RFC7231, Section 7.1.1.1](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.1.1), e.g. "Sun, 06 Nov 1994 08:49:37 GMT".
`url`: a fully qualified url that the user can follow to learn more about what is being deprecated.

JeffreyRichter:


azure/Guidelines.md
```diff
@@ -747,6 +749,20 @@ While removing a value from an enum is a breaking change, adding value to an enu
 
 > :ballot_box_with_check: **You SHOULD** use extensible enums unless you are positive that the symbol set will **NEVER** change over time.
 
+### Deprecating Behavior Notification
+
+When the [API Versioning](#API-Versioning) guidance above cannot be followed and the [Azure Breaking Change Reviewers](mailto:azbreakchangereview@microsoft.com) approve a breaking change to the operation/service, the operation/service that is being deprecated must add the `azure-deprecating` response header with a semicolon-delimited string notifying the caller what is being deprecated, when it will no longer function, and a URL linking to more information. The purpose is to inform customers (when debugging/logging responses) that they must take action to modify their call to the service's operation or their call will soon stop working entirely. It is not expected that client code will examine/parse this header's value in any way; it is purely informational to a human being.
```
MushMal: Do we want to include the breaking change reviewers email to the public doc? Also, recommend including a sentence about services using public communication channels/announcements to reach out to the customers (couldn't find it in the doc)

JeffreyRichter: We already had this email in the API Versioning section above so I just repeated it here. I agree that it's not ideal to have the email in the public doc but I think it's useful to have the reference in these guidelines. If we remove it, then it's harder for people to get in touch with us. I can say this: to date, this hasn't been a problem.

The notification mechanism is spelled out well in the Azure Breaking Change policy document and is somewhat involved/complicated - this we do not make public.

MushMal (approved):

azure/Guidelines.md
```diff
@@ -747,6 +749,20 @@ While removing a value from an enum is a breaking change, adding value to an enu
...
+:white_check_mark: **DO** add the 'azure-deprercating' header with a string value to all service operations in your service's contract file (cadl/swagger).
+
+:white_check_mark: **DO** include this header in the operation's response _only if_ the operation will stop working in the future and the client _must take_ action in order for it to keep working. NOTE: We do not want to scare customers with this header.
+
+:white_check_mark: **DO** make the header's value a semicolon-delimited string indicating a set of deprecations where each one indicates what is deprecating, when it is deprecating, and a URL to more information. For example: 
+
+```text
+azure-deprecating: api-version=2009-27-07 will stop working on 2022-12-01 (https://azure.microsoft.com/en-us/updates/video-analyzer-retirement);TLS 1.0 & 1.1 will stop working on 2020-10-30 (https://azure.microsoft.com/en-us/updates/azure-active-directory-registration-service-is-ending-support-for-tls-10-and-11/)
+```
+
```
tomkerkhove, line 767: Let's make sure people are aware that they should not do this without both of these, even if this is already mentioned above:
```suggestion

:no_entry: **DO NOT** introduce this header without this pattern without approval from [Azure Breaking Change Reviewers](mailto:azbreakchangereview@microsoft.com) and an official deprecation notice on [Azure Updates](https://azure.microsoft.com/updates/).
```

azure/Guidelines.md
```diff
@@ -747,6 +749,20 @@ While removing a value from an enum is a breaking change, adding value to an enu
 
 > :ballot_box_with_check: **You SHOULD** use extensible enums unless you are positive that the symbol set will **NEVER** change over time.
 
+### Deprecating Behavior Notification
+
+When the [API Versioning](#API-Versioning) guidance above cannot be followed and the [Azure Breaking Change Reviewers](mailto:azbreakchangereview@microsoft.com) approve a breaking change to the operation/service, the operation/service that is being deprecated must add the `azure-deprecating` response header with a semicolon-delimited string notifying the caller what is being deprecated, when it will no longer function, and a URL linking to more information. The purpose is to inform customers (when debugging/logging responses) that they must take action to modify their call to the service's operation or their call will soon stop working entirely. It is not expected that client code will examine/parse this header's value in any way; it is purely informational to a human being.
+
+:white_check_mark: **DO** add the 'azure-deprercating' header with a string value to all service operations in your service's contract file (cadl/swagger).
```
tomkerkhove: Fixed a typo:
```suggestion
:white_check_mark: **DO** add the 'azure-deprecating' header with a string value to all service operations in your service's contract file (cadl/swagger).
```

However, I'm more of a fan of `azure-api-deprecated` which is more clear

azure/Guidelines.md
```diff
@@ -747,6 +749,20 @@ While removing a value from an enum is a breaking change, adding value to an enu
...
+
+:white_check_mark: **DO** include this header in the operation's response _only if_ the operation will stop working in the future and the client _must take_ action in order for it to keep working. NOTE: We do not want to scare customers with this header.
+
+:white_check_mark: **DO** make the header's value a semicolon-delimited string indicating a set of deprecations where each one indicates what is deprecating, when it is deprecating, and a URL to more information. For example: 
+
+```text
+azure-deprecating: api-version=2009-27-07 will stop working on 2022-12-01 (https://azure.microsoft.com/en-us/updates/video-analyzer-retirement);TLS 1.0 & 1.1 will stop working on 2020-10-30 (https://azure.microsoft.com/en-us/updates/azure-active-directory-registration-service-is-ending-support-for-tls-10-and-11/)
```
tomkerkhove: This shows a bit of my doubt with the single header as we don't enforce structure which will be a pain for customers given every service will use its own flavor. I've seen enough occasions of this while maintaining [Azure Deprecation Notices](https://azure-deprecation-notices.cloud/) where some teams do them properly, others not so much.

I believe we should be more explicit in what the required format is and required information that must be provided is.

Another thing that comes to mind - Should they open a review with us to get approval on the header content? Given they will have to update their specs we should also enforce them to provide the samples they want to emit.

tomkerkhove: Also this example is a bit misleading - Is the intent of this header to announce deprecations for the whole API version or just one or more operations of that API version?

tomkerkhove:
````suggestion
:white_check_mark: **DO** make the header's value a semicolon-delimited string indicating a set of deprecations where each one indicates what is deprecating, when it is deprecating, and a URL to more information.

Deprecations should use the following pattern:
```text
<description> will retire on <date> (`url`)
```

Where the following placeholders should be provided:
- `description`: a human-readable description of what is being deprecated
- `date`: the target date that this will be deprecated. This should be expressed following the format in [ISO 8601](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.1.1), e.g. "2022-10-31".
- `url`: a fully qualified url that the user can follow to learn more about what is being deprecated, preferably to Azure Updates.

For example: 
```text
azure-deprecating: API version 2009-27-07 will retire on 2022-12-01 (https://azure.microsoft.com/en-us/updates/video-analyzer-retirement);TLS 1.0 & 1.1 will retire on 2020-10-30 (https://azure.microsoft.com/en-us/updates/azure-active-directory-registration-service-is-ending-support-for-tls-10-and-11/)
```

azure/Guidelines.md
tomkerkhove: The current intro only mentions APIs that are doing breaking changes, however, I believe we should recommend/enforcing teams that have API deprecations in stable/preview APIs as well even if there is no breaking behavior.

For example - "This API version is going away by xyz", unless you see that as breaking of course, I figure "breaking" in this context means people change a given API version to act differently after a given date. So we might want to clarify this

azure/Guidelines.md
tomkerkhove: Kudos to @markweitzel for the idea on field spec

azure/Guidelines.md
JeffreyRichter: We don't deprecate APIs; we deprecate entire service versions.

azure/Guidelines.md
JeffreyRichter: We always deprecate an entire service version; never some of its operations. A new service version would omit the operations no longer supported. 

Also, it is unusual for an operation to become obsolete entirely. Instead, the operation remains but it breaks in some fashion: different request/response models, different behavior, different return codes, sync into async, non-pageable list becomes a pageable list, etc.

JeffreyRichter: Also, I get what you're saying about the structure but there may be some deprecations that don't quite fit the structure and since these strings are for human consumption only, the lack of structure may discourage someone from trying to parse them (which is goodness) and the service team can modify the string in a non-breaking way since there is no contract about their format.

azure/Guidelines.md
JeffreyRichter: Azure's policy is that a given API version can NEVER change its behavior. So, the definition of "breaking" is that the next version **requires** that old clients update their code. When the new/breaking version ships is when all the previous versions are deprecated. This is when the header would be returned from customer calls specifying any previous version.

azure/Guidelines.md
```diff
@@ -747,6 +749,33 @@ While removing a value from an enum is a breaking change, adding value to an enu
...
+```
+
+Where the following placeholders should be provided:
+- `description`: a human-readable description of what is being deprecated
+- `date`: the target date that this will be deprecated. This should be expressed following the format in [ISO 8601](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.1.1), e.g. "2022-10-31".
+- `url`: a fully qualified url that the user can follow to learn more about what is being deprecated, preferably to Azure Updates.
+
+For example: 
+```text
+azure-deprecating: API version 2009-27-07 will retire on 2022-12-01 (https://azure.microsoft.com/en-us/updates/video-analyzer-retirement);TLS 1.0 & 1.1 will retire on 2020-10-30 (https://azure.microsoft.com/en-us/updates/azure-active-directory-registration-service-is-ending-support-for-tls-10-and-11/)
```
heaths: Nit: per guidelines, locales should not be included in URIs to be viewed by humans behind browsers (not exact wording). Let users' browsers pick the best locale.
```suggestion
azure-deprecating: API version 2009-27-07 will retire on 2022-12-01 (https://azure.microsoft.com/updates/video-analyzer-retirement);TLS 1.0 & 1.1 will retire on 2020-10-30 (https://azure.microsoft.com/updates/azure-active-directory-registration-service-is-ending-support-for-tls-10-and-11/)
```
I get this is an example, but help show - maybe even call out - that locales should not be included in URIs.

azure/Guidelines.md
tomkerkhove: So what about `azure-api-version-deprecated`?

azure/Guidelines.md
tomkerkhove: I was going to suggest to at least provide advisory around what should be there but that was already added, thank you!

> We always deprecate an entire service version; never some of its operations. A new service version would omit the operations no longer supported.
> 
> Also, it is unusual for an operation to become obsolete entirely. Instead, the operation remains but it breaks in some fashion: different request/response models, different behavior, different return codes, sync into async, non-pageable list becomes a pageable list, etc.

This is majority of the cases, agreed. But what about things such as TLS version deprecation as you have in the example?

tomkerkhove: But let's resolve this comment, I'm OK with what we have now

azure/Guidelines.md
tomkerkhove: I think it's not 100% clear that a breaking change is only going to be noticed when using a newer API version given deprecations typically mean that at some point something will stop working unless they have migrated to another feature.

Here it is somewhat the same, but it's specific to API versions so I would highlight that and link to the breaking change definition:

```suggestion
When the [API Versioning](#API-Versioning) guidance above cannot be followed and the [Azure Breaking Change Reviewers](mailto:azbreakchangereview@microsoft.com) approve a [breaking change](#123-definition-of-a-breaking-change) to a specific API version it must be communicated to its callers. The API version that is being deprecated must add the `azure-deprecating` response header with a semicolon-delimited string notifying the caller what is being deprecated, when it will no longer function, and a URL linking to more information such as what new operation they should use instead.

The purpose is to inform customers (when debugging/logging responses) that they must take action to modify their call to the service's operation and use a newer API version or their call will soon stop working entirely. It is not expected that client code will examine/parse this header's value in any way; it is purely informational to a human being.
```

azure/Guidelines.md
```diff
@@ -747,6 +749,33 @@ While removing a value from an enum is a breaking change, adding value to an enu
...
+```
+```
```
tomkerkhove, line 776: Double code block end is breaking formatting of the doc
```suggestion
```

tomkerkhove (approved):

azure/Guidelines.md
```diff
@@ -747,6 +749,34 @@ While removing a value from an enum is a breaking change, adding value to an enu
...
+
+:white_check_mark: **DO** add the 'azure-deprecating' header with a string value to all service operations in your service's contract file (cadl/swagger).
+
+:white_check_mark: **DO** include this header in the operation's response _only if_ the operation will stop working in the future and the client _must take_ action in order for it to keep working. NOTE: We do not want to scare customers with this header.
+
+:white_check_mark: **DO** make the header's value a semicolon-delimited string indicating a set of deprecations where each one indicates what is deprecating, when it is deprecating, and a URL to more information.
+
+Deprecations should use the following pattern:
+```text
+<description> will retire on <date> (`url`)
```
```suggestion
<description> will retire on <date> (`url`);
```

azure/Guidelines.md
```diff
@@ -747,6 +749,34 @@ While removing a value from an enum is a breaking change, adding value to an enu
 
 > :ballot_box_with_check: **You SHOULD** use extensible enums unless you are positive that the symbol set will **NEVER** change over time.
 
+### Deprecating Behavior Notification
+
+When the [API Versioning](#API-Versioning) guidance above cannot be followed and the [Azure Breaking Change Reviewers](mailto:azbreakchangereview@microsoft.com) approve a [breaking change](#123-definition-of-a-breaking-change) to a specific API version it must be communicated to its callers. The API version that is being deprecated must add the `azure-deprecating` response header with a semicolon-delimited string notifying the caller what is being deprecated, when it will no longer function, and a URL linking to more information such as what new operation they should use instead.
+
+The purpose is to inform customers (when debugging/logging responses) that they must take action to modify their call to the service's operation and use a newer API version or their call will soon stop working entirely. It is not expected that client code will examine/parse this header's value in any way; it is purely informational to a human being.
```
tg-msft: Do we want to say the format of the message is not part of the API contract?

azure/Guidelines.md
```diff
@@ -747,6 +749,34 @@ While removing a value from an enum is a breaking change, adding value to an enu
...
+- `description`: a human-readable description of what is being deprecated
+- `date`: the target date that this will be deprecated. This should be expressed following the format in [ISO 8601](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.1.1), e.g. "2022-10-31".
+- `url`: a fully qualified url that the user can follow to learn more about what is being deprecated, preferably to Azure Updates.
+
+For example: 
+```text
+azure-deprecating: API version 2009-27-07 will retire on 2022-12-01 (https://azure.microsoft.com/updates/video-analyzer-retirement);TLS 1.0 & 1.1 will retire on 2020-10-30 (https://azure.microsoft.com/updates/azure-active-directory-registration-service-is-ending-support-for-tls-10-and-11/)
+```
+
+:no_entry: **DO NOT** introduce this header without this pattern without approval from [Azure Breaking Change Reviewers](mailto:azbreakchangereview@microsoft.com) and an official deprecation notice on [Azure Updates](https://azure.microsoft.com/updates/).
```
tg-msft: This reads a little odd.  Did we repeat something or forget a word?

```suggestion
:no_entry: **DO NOT** introduce this header without approval from [Azure Breaking Change Reviewers](mailto:azbreakchangereview@microsoft.com) and an official deprecation notice on [Azure Updates](https://azure.microsoft.com/updates/).
```

or 

```suggestion
:no_entry: **DO NOT** introduce this header without this pattern or without approval from [Azure Breaking Change Reviewers](mailto:azbreakchangereview@microsoft.com) and an official deprecation notice on [Azure Updates](https://azure.microsoft.com/updates/).
```

azure/Guidelines.md
```diff
@@ -747,6 +749,34 @@ While removing a value from an enum is a breaking change, adding value to an enu
...
+
+The purpose is to inform customers (when debugging/logging responses) that they must take action to modify their call to the service's operation and use a newer API version or their call will soon stop working entirely. It is not expected that client code will examine/parse this header's value in any way; it is purely informational to a human being.
+
+:white_check_mark: **DO** add the 'azure-deprecating' header with a string value to all service operations in your service's contract file (cadl/swagger).
+
+:white_check_mark: **DO** include this header in the operation's response _only if_ the operation will stop working in the future and the client _must take_ action in order for it to keep working. NOTE: We do not want to scare customers with this header.
```
tg-msft: Is it worth putting the NOTE on its own line so it stands out a little more?

tg-msft (commented):
Looks good - added a few small questions.

azure/Guidelines.md
```suggestion
The purpose is to inform customers (when debugging/logging responses) that they must take action to modify their call to the service's operation and use a newer API version or their call will soon stop working entirely. It is not expected that client code will examine/parse this header's value in any way; it is purely informational to a human being. The string is _not_ part of an API contract (except for the semi-colon delimiters) and may be changed/improved at any time without incurring a breaking change.
```

azure/Guidelines.md
JeffreyRichter: I modified the text

JeffreyRichter: I tried to play with this to make it stand out more. If my attempts fail, we can improve later.

azure/Guidelines.md
```diff
@@ -747,6 +749,34 @@ While removing a value from an enum is a breaking change, adding value to an enu
...
+
+The purpose is to inform customers (when debugging/logging responses) that they must take action to modify their call to the service's operation and use a newer API version or their call will soon stop working entirely. It is not expected that client code will examine/parse this header's value in any way; it is purely informational to a human being. The string is _not_ part of an API contract (except for the semi-colon delimiters) and may be changed/improved at any time without incurring a breaking change.
+
+:white_check_mark: **DO** add the 'azure-deprecating' header with a string value to all service operations in your service's contract file (cadl/swagger).
+
+:white_check_mark: **DO** include this header in the operation's response _only if_ the operation will stop working in the future and the client _must take_ action in order for it to keep working. NOTE: We do not want to scare customers with this header.
```
```suggestion
:white_check_mark: **DO** include this header in the operation's response _only if_ the operation will stop working in the future and the client _must take_ action in order for it to keep working. 
> NOTE: We do not want to scare customers with this header.
```

tomkerkhove (approved):

azure/Guidelines.md
tomkerkhove: LGTM

azure/Guidelines.md
```diff
@@ -747,6 +749,34 @@ While removing a value from an enum is a breaking change, adding value to an enu
...
 
+### Deprecating Behavior Notification
+
+When the [API Versioning](#API-Versioning) guidance above cannot be followed and the [Azure Breaking Change Reviewers](mailto:azbreakchangereview@microsoft.com) approve a [breaking change](#123-definition-of-a-breaking-change) to a specific API version it must be communicated to its callers. The API version that is being deprecated must add the `azure-deprecating` response header with a semicolon-delimited string notifying the caller what is being deprecated, when it will no longer function, and a URL linking to more information such as what new operation they should use instead.
+
+The purpose is to inform customers (when debugging/logging responses) that they must take action to modify their call to the service's operation and use a newer API version or their call will soon stop working entirely. It is not expected that client code will examine/parse this header's value in any way; it is purely informational to a human being. The string is _not_ part of an API contract (except for the semi-colon delimiters) and may be changed/improved at any time without incurring a breaking change.
```
tombuildsstuff, line 756: Whilst deprecation times may vary on a service-by-service basis, it'd be good to have some centralised guidelines for the minimum deprecation notice period here.

We've had a number of awkward breaking changes over the years (e.g. Soft Delete for Key Vault being enforced on December 31st) and deprecations ([service teams yanking API versions with (AFAICT) no heads up](https://github.com/hashicorp/terraform-provider-azurerm/issues/18185)) - that having a centralised minimum deprecation period (e.g. 6 months?) would be useful/worth having here?

Specifically around the Key Vault Soft Delete changes, would it also be worth recommending avoiding deprecation periods in the times of year where users are likely to be less available (e.g. don't schedule them in christmas/new year/the summer)?

JeffreyRichter: Officially, we do: When Azure announces a deprecation, the currently-shipping versions should work as-is for at least 3 years. This shrinks to 1 year [possibly less] if the old versions have a security or compliance issue and we need to get customers to move away from them more aggressively). We do batch the deprecation announcements so that they come out just a few times per year. We should be able to announce anytime (including x-mas/new year) because the old api-versions should continue running as-is for 3/1 years).

Occassionally, teams introduce breaking behaviors by mistake as our engineers are human, but we correct these as soon as we discover/are notified of them.

manicminer: That's good to know! Is this published anywhere that we can link to if we need to refer to that guideline?

azure/Guidelines.md
```diff
@@ -747,6 +749,34 @@ While removing a value from an enum is a breaking change, adding value to an enu
...
+```
+
+Where the following placeholders should be provided:
+- `description`: a human-readable description of what is being deprecated
+- `date`: the target date that this will be deprecated. This should be expressed following the format in [ISO 8601](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.1.1), e.g. "2022-10-31".
+- `url`: a fully qualified url that the user can follow to learn more about what is being deprecated, preferably to Azure Updates.
+
+For example: 
+```text
+azure-deprecating: API version 2009-27-07 will retire on 2022-12-01 (https://azure.microsoft.com/updates/video-analyzer-retirement);TLS 1.0 & 1.1 will retire on 2020-10-30 (https://azure.microsoft.com/updates/azure-active-directory-registration-service-is-ending-support-for-tls-10-and-11/)
```
manicminer, line 775: Suggest recommending that links are surrounded by angle brackets rather than parentheses, as is often customary when URIs are mixed with human readable text.

```suggestion
azure-deprecating: API version 2009-27-07 will retire on 2022-12-01 <https://azure.microsoft.com/updates/video-analyzer-retirement>;TLS 1.0 & 1.1 will retire on 2020-10-30 <https://azure.microsoft.com/updates/azure-active-directory-registration-service-is-ending-support-for-tls-10-and-11/>
```

azure/Guidelines.md
```diff
@@ -747,6 +749,34 @@ While removing a value from an enum is a breaking change, adding value to an enu
...
+
+The purpose is to inform customers (when debugging/logging responses) that they must take action to modify their call to the service's operation and use a newer API version or their call will soon stop working entirely. It is not expected that client code will examine/parse this header's value in any way; it is purely informational to a human being. The string is _not_ part of an API contract (except for the semi-colon delimiters) and may be changed/improved at any time without incurring a breaking change.
+
+:white_check_mark: **DO** add the 'azure-deprecating' header with a string value to all service operations in your service's contract file (cadl/swagger).
```
mikekistler: Recommend either dropping this or moving it to the Azure API Style Guide to maintain our separation of guidelines for the REST API implementation vs REST API description.
```suggestion
```

azure/Guidelines.md
JeffreyRichter: I'm not opposed to your suggestion, but my experience does not indicate that this is _customary_. Can you site something indicating that it is? Markdown surrounds link text in [], not <>

mikekistler (approved):
Looks good! 👍

azure/Guidelines.md
manicminer: The closest I know of is the [addr-spec for email addresses](https://www.rfc-editor.org/rfc/rfc822#section-6). I'm not sure I've actually seen a mixed-use HTTP header or similar field like this before, but angle brackets definitely feel more relevant than parentheses or square brackets - unless you're saying the value of this header _is_ Markdown?

tomkerkhove:
🎉

PR #390, opened 2022-08-26
Add note about URL-encoding for nextLink

mikekistler:
This is a tiny PR to add a note saying that the service is responsible for any URL-encoding needed on the value of `nextLink`.


JeffreyRichter (approved):

markweitzel (approved):
LGTM! THANK YOU!!

PR #391, opened 2022-08-29
Add naming guidelines for Azure services

mikekistler:
This PR adds basic guidance for naming in Azure REST APIs. Some of this is borrowed from the Microsoft and Graph API guidelines and some on naming discussions we've had in prior Azure REST API reviews.

### DIFF
--- a/azure/ConsiderationsForServiceDesign.md
+++ b/azure/ConsiderationsForServiceDesign.md
@@ -116,8 +116,6 @@ For example, `collectedItems` not `itemsCollected`
 
 For example, `nextUrl` not `nextURL`.
 
-:ballot_box_with_check: **YOU SHOULD** use "DateTime" suffix in names of `date-time` values.
-
 :ballot_box_with_check: **YOU SHOULD** use a suffix of the unit of measurement for values with a clear unit of measurement (such as bytes, miles, and so on). Use a generally accepted abbreviation for the units (e.g. "Km" rather than "Kilometers") when appropriate.
 
 :ballot_box_with_check: **YOU SHOULD** use an int for time durations and include the time units in the name.
--- a/azure/ConsiderationsForServiceDesign.md
+++ b/azure/ConsiderationsForServiceDesign.md
@@ -88,46 +88,53 @@ As you build out your service and API, there are a number of decisions that can 
 
 Good names for resources, properties, operations, and parameters are essential to a good developer experience.
 
-Resources are described by nouns. Resource and property names must be descriptive and easy for customers to understand. Ease of understanding comes from familiarity and recognition; you should favor consistency with other Azure services, names in the product's portal/user interface, and industry standards.
+Resources are described by nouns. Resource and property names must be descriptive and easy for customers to understand.
+Avoid names that refer to service implementation details, e.g. "TreeLeafNode", or simply describe the structure of value without
+describing its purpose, e.g. "KeyValuePair".
+Ease of understanding comes from familiarity and recognition; you should favor consistency with other Azure services, names in the product's portal/user interface, and industry standards.
 
 Names should aid developers in discovering functionality without having to constantly refer to documentation.
 Use common patterns and standard conventions to aid developers in correctly guessing common property names and meanings.
 Use verbose naming patterns and avoid abbreviations other than
 well-known acronyms in your service domain.
 
+:heavy_check_mark: **DO** use the same name for the same concept and different names for different concepts wherever possible.
+
 ### Recommended Naming Conventions
 
 The following are recommended naming conventions for Azure services:
 
-:heavy_check_mark: **DO** use singular nouns for resource (schema) names.
+:heavy_check_mark: **DO** name collections as plural nouns or plural noun phrases using correct English.
 
-:heavy_check_mark: **DO** use plural nouns for collections (for listing type or collection properties).
+:heavy_check_mark: **DO** name values that are not collections as singular nouns or singular noun phrases.
 
-:ballot_box_with_check: **YOU SHOULD** case all acronyms the same as a normal word.
+:ballot_box_with_check: **YOU SHOULD** should place the adjective before the noun in names that contain both a noun and an adjective.
 
-For example, `nextUrl` and not `nextURL`.
+For example, `collectedItems` not `itemsCollected`
 
-☑️ **YOU SHOULD** use "At" suffix in names of `date-time` values.
+:ballot_box_with_check: **YOU SHOULD** case all acronyms as though they were regular words (i.e. lower camelCase).
 
-:ballot_box_with_check: **YOU SHOULD** use "is" prefix in names of `boolean` values.
+For example, `nextUrl` not `nextURL`.
 
-☑️ **YOU SHOULD** use a suffix of the unit of measurement for values with a clear unit of measurement (such as bytes, miles, and so on). Use a generally accepted abbreviation for the units (e.g. "Km" rather than "Kilometers") when appropriate.
+:ballot_box_with_check: **YOU SHOULD** use "DateTime" suffix in names of `date-time` values.
+
+:ballot_box_with_check: **YOU SHOULD** use a suffix of the unit of measurement for values with a clear unit of measurement (such as bytes, miles, and so on). Use a generally accepted abbreviation for the units (e.g. "Km" rather than "Kilometers") when appropriate.
 
 :ballot_box_with_check: **YOU SHOULD** use an int for time durations and include the time units in the name.
 
 For example, `expirationDays` as `int` and not `expiration` as `date-time`.
 
-:ballot_box_with_check: **YOU SHOULD** prefix property names for properties concerning a different resource, particular for resource identifiers.
-
 :warning: **YOU SHOULD NOT** use brand names in resource or property names.
 
-⚠️ **YOU SHOULD NOT** use acronyms or abbreviations unless they are broadly understood for example, "ID" or "URL", but not "Num" for "number".
+:warning: **YOU SHOULD NOT** use acronyms or abbreviations unless they are broadly understood for example, "ID" or "URL", but not "Num" for "number".
+
+:warning: **YOU SHOULD NOT** use names that are reserved words in widely used programming languages.
+
+:no_entry: **DO NOT** use "is" prefix in names of `boolean` values.
 
 :no_entry: **DO NOT** use redundant words in names.
 
 For example, `/phones/number` and not `phone/phoneNumber`.
-
-:no_entry: **DO NOT** use a "request", "response", or "collection" suffix on resource names.
 
 ### Common names
 
@@ -135,13 +142,15 @@ The following are recommended names for properties that match the associated des
 
 | Name | Description |
 |------------- | --- |
-| createdAt | The date-time that the resource was created |
-| updatedAt | The date-time that the resource was last updated/modified |
+| createdDateTime | The date-time that the resource was created |
+| updatedDateTime | The date-time that the resource was last updated/modified |
 | kind   | The discriminator value for a polymorphic resource |
 
 ### `name` vs `id`
 
-The identifier of a resource should be suffixed with `Id`. This holds even in the case where the identifier is assigned by the user with a PUT/PATCH method.
+:heavy_check_mark: **DO** use "Id" suffix for the name of rhe identifier of a resource.
+
+This holds even in the case where the identifier is assigned by the user with a PUT/PATCH method.
 
 ## Use Previews to Iterate
 Before releasing your API plan to invest significant design effort, get customer feedback, & iterate through multiple preview releases. This is especially important for V1 as it establishes the abstractions and patterns that developers will use to interact with your service.
--- a/azure/ConsiderationsForServiceDesign.md
+++ b/azure/ConsiderationsForServiceDesign.md
@@ -89,8 +89,8 @@ As you build out your service and API, there are a number of decisions that can 
 Good names for resources, properties, operations, and parameters are essential to a great developer experience.
 
 Resources are described by nouns. Resource and property names must be descriptive and easy for customers to understand.
-Avoid names that refer to service implementation details, e.g. "TreeLeafNode", or simply describe the structure of value without
-describing its purpose, e.g. "KeyValuePair".
+Use names that correspond to user scenarios rather than service implementation details, e.g. "Diagnosis" and not "TreeLeafNode".
+Names should convey the value's purpose and not simply describe its structure, e.g. "ConfigurationSetting" and not "KeyValuePair".
 Ease of understanding comes from familiarity and recognition; you should favor consistency with other Azure services, names in the product's portal/user interface, and industry standards.
 
 Names should aid developers in discovering functionality without having to constantly refer to documentation.
@@ -128,7 +128,7 @@ For example, `expirationDays` as `int` and not `expiration` as `date-time`.
 
 :warning: **YOU SHOULD NOT** use names that are reserved words in widely used programming languages (including C#, Java, JavaScript/TypeScript, Python, C++, and Go).
 
-:no_entry: **DO NOT** use "is" prefix in names of `boolean` values.
+:no_entry: **DO NOT** use "is" prefix in names of `boolean` values, e.g. "enabled" not "isEnabled".
 
 :no_entry: **DO NOT** use redundant words in names.
 
@@ -167,7 +167,7 @@ Before releasing your API plan to invest significant design effort, get customer
 As your service evolves over time, it will be natural that you want to remove operations that are no longer needed. For example, additional requirements or new capability in your service, may have resulted in a new operation that, effectively, replaces an old one.
 Azure has a well established breaking changes policy that describes how to approach these kinds of changes. As part of this policy, the service team is required to clearly communicate to customers when their API is changing, e.g. deprecating operations. Often, this is done via an email to the address that is attached to the Azure subscription.
 
-However, given how many organizations are structured, it's common that this email address is different from the actual people writing code against your API. To address this, the service API should declare that it may return the `azure-deprecating` header, to indicate that this operation will be removed in the future. There is a simple string convention, specied in the [Azure REST API Guidelines](https://aka.ms/azapi/guidelines) that provides more information about the forthcoming deprecation.
+However, given how many organizations are structured, it's common that this email address is different from the actual people writing code against your API. To address this, the service API should declare that it may return the `azure-deprecating` header, to indicate that this operation will be removed in the future. There is a simple string convention, specifed in the [Azure REST API Guidelines](https://aka.ms/azapi/guidelines) that provides more information about the forthcoming deprecation.
 This header is targeted at developers or operation professionals, and it is intended to give them enough information and lead time to properly adapt to this change. Your documentation should reference this header and encourage logging and alerting practices based on its presence.
 
 ## Avoid Surprises
--- a/azure/ConsiderationsForServiceDesign.md
+++ b/azure/ConsiderationsForServiceDesign.md
@@ -84,6 +84,65 @@ As you build out your service and API, there are a number of decisions that can 
 
 :ballot_box_with_check: **YOU SHOULD** implement [conditional requests](https://tools.ietf.org/html/rfc7232) early. This allows you to support concurrency, which tends to be a concern later on.
 
+## Use Good Names
+
+Good names for resources, properties, operations, and parameters are essential to a good developer experience.
+
+Resources are typically described by nouns. Resource and property names must be descriptive and easy to understand for end-users. Ease of understanding comes from familiarity and recognition; you should favor consistency with other Azure services, names in the product user interface, and industry standards.
+
+Names should aid developers in discovering functionality without having to constantly refer to documentation.
+Use common patterns and standard conventions to aid developers in correctly guessing common property names and meanings.
+Use verbose naming patterns and avoid abbreviations other than
+well-known acronyms in your service domain.
+
+### Recommended Naming Conventions
+
+The following are recommended naming conventions for Azure services:
+
+:heavy_check_mark: **DO** use singular nouns for resource (schema) names.
+
+:heavy_check_mark: **DO** use plural nouns for collections (for listing type or collection properties).
+
+:ballot_box_with_check: **YOU SHOULD** case all acronyms the same as a normal word.
+
+For example, `nextUrl` and not `nextURL`.
+
+:ballot_box_with_check: **YOU SHOULD** use "at" suffix in names of `date-time` values.
+
+:ballot_box_with_check: **YOU SHOULD** use "is" prefix in names of `boolean` values.
+
+:ballot_box_with_check: **YOU SHOULD** use a suffix of the unit of measurement for values with a clear unit of measurement (such as bytes, miles, and so on). Use a generally accepted abbreviation for the units (e.g. "km" rather than "kilometers") when appropriate.
+
+:ballot_box_with_check: **YOU SHOULD** use an int for time durations and include the time units in the name.
+
+For example, `expirationDays` as `int` and not `expiration` as `date-time`.
+
+:ballot_box_with_check: **YOU SHOULD** prefix property names for properties concerning a different resource, particular for resource identifiers.
+
+:warning: **YOU SHOULD NOT** use brand names in resource or property names.
+
+:warning: **YOU SHOULD NOT** use acronyms or abbreviations unless they are broadly understood.
+
+:no_entry: **DO NOT** use redundant words in names.
+
+For example, `/phones/number` and not `phone/phoneNumber`.
+
+:no_entry: **DO NOT** use a "request", "response", or "collection" suffix on resource names.
+
+### Common names
+
+The following are recommended names for properties that match the associated description:
+
+| Name | Description |
+|------------- | --- |
+| createdAt | The date-time that the resource was created |
+| updatedAt | The date-time that the resource was last updated/modified |
+| kind   | The discriminator value for a polymorphic resource |
+
+### `name` vs `id`
+
+The identifier of a resource should be named `id`. This holds even in the case where the identifier is assigned by the user with a PUT method.
+
 ## Use Previews to Iterate
 Before releasing your API plan to invest significant design effort, get customer feedback, & iterate through multiple preview releases. This is especially important for V1 as it establishes the abstractions and patterns that developers will use to interact with your service.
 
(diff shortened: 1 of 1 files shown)


MushMal (approved):

azure/ConsiderationsForServiceDesign.md
```diff
@@ -77,6 +77,65 @@ Understanding how your service is used and defining its model and interaction pa
...
+
+### Recommended Naming Conventions
+
+The following are recommended naming conventions for Azure services:
+
+:heavy_check_mark: **DO** use singular nouns for resource (schema) names.
+
+:heavy_check_mark: **DO** use plural nouns for collections (for listing type or collection properties).
+
+:ballot_box_with_check: **YOU SHOULD** case all acronyms the same as a normal word.
```
johanste, line 99: I don't quite understand what this means...

azure/ConsiderationsForServiceDesign.md
```diff
@@ -77,6 +77,65 @@ Understanding how your service is used and defining its model and interaction pa
...
+Use verbose naming patterns and avoid abbreviations other than
+well-known acronyms in your service domain.
+
+### Recommended Naming Conventions
+
+The following are recommended naming conventions for Azure services:
+
+:heavy_check_mark: **DO** use singular nouns for resource (schema) names.
+
+:heavy_check_mark: **DO** use plural nouns for collections (for listing type or collection properties).
```
johanste: I would suggest that we use a page suffix or similar rather than pluralizing the name. If nothing else, it will make it possible to provide a reusable pattern in Cadl that doesn't require the user to provide both a singular and plural name for a given resource type.

azure/ConsiderationsForServiceDesign.md
johanste: I think we can be more specific and say that acronyms should be Pascal cased.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -77,6 +77,65 @@ Understanding how your service is used and defining its model and interaction pa
...
+
+:ballot_box_with_check: **YOU SHOULD** case all acronyms the same as a normal word.
+
+For example, `nextUrl` and not `nextURL`.
+
+:ballot_box_with_check: **YOU SHOULD** use "at" suffix in names of `date-time` values.
```
johanste: The guidance may read better if we separate model from property name guidelines.

johanste: This is in direct conflict with earlier naming guidance in the Microsoft REST API guidelines which stated a type suffix for date time typed properties. Just so we are in agreement that value of the the new/changed guideline outweigh the cost of the change.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -77,6 +77,65 @@ Understanding how your service is used and defining its model and interaction pa
...
+
+:ballot_box_with_check: **YOU SHOULD** use "is" prefix in names of `boolean` values.
```
johanste: This is not something I've seen in other REST APIs. I would expect this in guidelines for a more verbose programming language, not for REST APIs. Why are we making the names verbose like this?

azure/ConsiderationsForServiceDesign.md
```diff
@@ -77,6 +77,65 @@ Understanding how your service is used and defining its model and interaction pa
...
+
+:ballot_box_with_check: **YOU SHOULD** use a suffix of the unit of measurement for values with a clear unit of measurement (such as bytes, miles, and so on). Use a generally accepted abbreviation for the units (e.g. "km" rather than "kilometers") when appropriate.
+
+:ballot_box_with_check: **YOU SHOULD** use an int for time durations and include the time units in the name.
+
+For example, `expirationDays` as `int` and not `expiration` as `date-time`.
+
+:ballot_box_with_check: **YOU SHOULD** prefix property names for properties concerning a different resource, particular for resource identifiers.
```
johanste: What does this mean?

azure/ConsiderationsForServiceDesign.md
```diff
@@ -77,6 +77,65 @@ Understanding how your service is used and defining its model and interaction pa
...
+
+| Name | Description |
+|------------- | --- |
+| createdAt | The date-time that the resource was created |
+| updatedAt | The date-time that the resource was last updated/modified |
+| kind   | The discriminator value for a polymorphic resource |
+
+### `name` vs `id`
+
+The identifier of a resource should be named `id`. This holds even in the case where the identifier is assigned by the user with a PUT method.
```
johanste: Is this a full identifier or just the last segment?

azure/ConsiderationsForServiceDesign.md
mikekistler: A resource will only have a singular name. Only a collection of resources will have a plural name.

azure/ConsiderationsForServiceDesign.md
mikekistler: It's not correct to say that acronyms are Pascal-cased because acronyms at the beginning of a lower-camel-case name should be all lower case. But I admit that the current wording or at least a better example is needed here.

azure/ConsiderationsForServiceDesign.md
mikekistler: That's true and I meant to make this same exact comment on my own PR. I wrote it this way because this is the guidance I've heard us give in past API reviews. But we should change to the Microsoft (and Graph) convention here unless there is good cause to "go our own way".

azure/ConsiderationsForServiceDesign.md
mikekistler: I think we have given this advice in past API reviews and it also included in [Graph's naming guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/graph/GuidelinesGraph.md#:~:text=SHOULD%20prefix%20Boolean%20properties%20with%20is%2C%20unless%20this%20leads%20to%20awkward%20or%20unnatural%20sounding%20names%20for%20Boolean%20properties.)

But I would be fine to change it if others find it objectionable.

azure/ConsiderationsForServiceDesign.md
mikekistler: I'm referring to just the last segment here. How can I make this clearer?

azure/ConsiderationsForServiceDesign.md
weidongxu-microsoft: probably means `blobId` or `blobName` outside context of storage blob?

azure/ConsiderationsForServiceDesign.md
```diff
@@ -77,6 +77,65 @@ Understanding how your service is used and defining its model and interaction pa
...
+
+:ballot_box_with_check: **YOU SHOULD** use "at" suffix in names of `date-time` values.
+
+:ballot_box_with_check: **YOU SHOULD** use "is" prefix in names of `boolean` values.
+
+:ballot_box_with_check: **YOU SHOULD** use a suffix of the unit of measurement for values with a clear unit of measurement (such as bytes, miles, and so on). Use a generally accepted abbreviation for the units (e.g. "km" rather than "kilometers") when appropriate.
+
+:ballot_box_with_check: **YOU SHOULD** use an int for time durations and include the time units in the name.
+
+For example, `expirationDays` as `int` and not `expiration` as `date-time`.
```
weidongxu-microsoft, line 123: Would `expirationInDays` be more popular?

azure/ConsiderationsForServiceDesign.md
johanste: Are you referring to path segments or model names here? Because I'm referring to the model name for the response model for a list operation? If we require model names to use plural, the effect is that we'll require Cadl authors to specify the plural name explicitly for a given resource. I would prefer that we don't and instead use `PageOfThing` rather than `Things`.

mikekistler: I can see that this is unclear.  I think this mainly refers to property names of arrays or maps. But I don't love `PageOfThing` as a schema name. A good item for discussion.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -77,6 +77,65 @@ Understanding how your service is used and defining its model and interaction pa
 
 :white_check_mark: **DO** create an [OpenAPI Definition](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md) (with [autorest extensions](https://github.com/Azure/autorest/blob/master/docs/extensions/readme.md)) describing the service. The OpenAPI definition is a key element of the Azure SDK plan and is essential for documentation, usability and discoverability of services.
 
+## Use Good Names
+
+Good names for resources, properties, operations, and parameters are essential to a good developer experience.
+
+Resources are typically described by nouns. Resource and property names must be descriptive and easy to understand for end-users. Ease of understanding comes from familiarity and recognition; you should favor consistency with other Azure services, names in the product user interface, and industry standards.
```
heaths: Nit: "end users" (no hyphen). It's not being used as a complex adjective. It's rarely a hyphenate.

azure/ConsiderationsForServiceDesign.md
heaths: Given this discussion frequently comes up even further right in the SDK archboard reviews, maybe we just need a comprehensive doc - language agnostic except for exceptions are needed, like with Go - to use for guidance and examples for both REST/Cadl and SDKs.

azure/ConsiderationsForServiceDesign.md
heaths: Based on different conventions, the language generators already handle this correctly if no prefix like "is" exists. I agree that it's not common in REST (rare, though I've seen it) and doesn't seem to add much value anyway. Why waste the 2 bytes per bool property?

If we did add it - and it would presumably only exist for newer services that actually conform to guidelines - now code generators have to take the effectively optional prefix into account and handle even more cases to make sure the SDK is idiomatic.

azure/ConsiderationsForServiceDesign.md
heaths: IMO, it's such a subtle difference the addition of 2 bytes doesn't seem to add enough value. It's also makes many of the property names more wordy.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -77,6 +77,65 @@ Understanding how your service is used and defining its model and interaction pa
...
+
+:ballot_box_with_check: **YOU SHOULD** use an int for time durations and include the time units in the name.
+
+For example, `expirationDays` as `int` and not `expiration` as `date-time`.
+
+:ballot_box_with_check: **YOU SHOULD** prefix property names for properties concerning a different resource, particular for resource identifiers.
+
+:warning: **YOU SHOULD NOT** use brand names in resource or property names.
+
+:warning: **YOU SHOULD NOT** use acronyms or abbreviations unless they are broadly understood.
```
heaths: Maybe add examples?
```suggestion
:warning: **YOU SHOULD NOT** use acronyms or abbreviations unless they are broadly understood for example, "ID" or "URL", but not "Num" for "number".
```

azure/ConsiderationsForServiceDesign.md
```suggestion
Resources are described by nouns. Resource and property names must be descriptive and easy for customers to understand. Ease of understanding comes from familiarity and recognition; you should favor consistency with other Azure services, names in the product's portal/user interface, and industry standards.
```

azure/ConsiderationsForServiceDesign.md
JeffreyRichter: We have typically made a distinction between REST and swagger/cadl. The design guidelines document barely mentions swagger/cadl - it is very rest focus. Perhaps we should now have a swagger/cadl guidelines document and there we can talk about model naming guidelines. Or, we could just have a section in our main guidelines document that focuses on swagger/cadl.

azure/ConsiderationsForServiceDesign.md
JeffreyRichter: Here's .NET's guidance on this: Acronyms are Pascal cased unless 2 letters (like IO). Go's guidance is acronyms are all uppercase.

So, in order to adhere to language guidelines, the code generator will have to recognize an acronym and apply the right casing to it. For this reason, I think first letter upper or lower (as appropriate for query parameter/property) and all remaining letters lowercase (to detect end of word) so language can detect the word, match it against known words/acronyms (not sure how to best to this - dictionary lookup?) and adjust casing if word is an acronym.

azure/ConsiderationsForServiceDesign.md
```suggestion
:ballot_box_with_check: **YOU SHOULD** use "At" suffix in names of `date-time` values.
```

azure/ConsiderationsForServiceDesign.md
JeffreyRichter: I'm don't think we want this. Booleans should be easy to read in an if statement: if (r.Modifiable) is better than if (r.IsModifiable). Also, sometimes prefixing with "has" (hasAddress) or "can" (canCopy) is better. I thought I found a website once where someone had a great discussion on this. Does anyone remember this? I think I shared the link.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -77,6 +77,65 @@ Understanding how your service is used and defining its model and interaction pa
...
+
+:ballot_box_with_check: **YOU SHOULD** case all acronyms the same as a normal word.
+
+For example, `nextUrl` and not `nextURL`.
+
+:ballot_box_with_check: **YOU SHOULD** use "at" suffix in names of `date-time` values.
+
+:ballot_box_with_check: **YOU SHOULD** use "is" prefix in names of `boolean` values.
+
+:ballot_box_with_check: **YOU SHOULD** use a suffix of the unit of measurement for values with a clear unit of measurement (such as bytes, miles, and so on). Use a generally accepted abbreviation for the units (e.g. "km" rather than "kilometers") when appropriate.
```
```suggestion
:ballot_box_with_check: **YOU SHOULD** use a suffix of the unit of measurement for values with a clear unit of measurement (such as bytes, miles, and so on). Use a generally accepted abbreviation for the units (e.g. "Km" rather than "Kilometers") when appropriate.
```

azure/ConsiderationsForServiceDesign.md
JeffreyRichter: I don't know what this mean either. But we do want some guideline that resource Ids should always end with Id. I'm proposing Id and not "Name" because Fielding calls them "Resource Identifiers" in section 5.2.1.1 of his REST dissertation.

azure/ConsiderationsForServiceDesign.md
```suggestion
The identifier of a resource should be suffixed with `Id`. This holds even in the case where the identifier is assigned by the user with a PUT/PATCH method.
```

azure/ConsiderationsForServiceDesign.md
weidongxu-microsoft: Preference of naming for boolean could be different. Like Golang prefer it without "is" prefix, and Java might actually like the "is" prefix in getter method (but not in the variable itself).

So we might not want it to be in REST itself.

azure/ConsiderationsForServiceDesign.md
weidongxu-microsoft: I don't have strong opinion on it. Feel that `expirationDays` might be interpreted as "what day or days of the expriation", but saying "timeoutInSeconds" appear not better than "timeoutSeconds". Hence we probably just stick with `expirationDays` pattern.

azure/ConsiderationsForServiceDesign.md
mikekistler: I believe the Go generator already does this.

azure/ConsiderationsForServiceDesign.md
mikekistler: I'll take this one out. For languages where this is a convention it is easy to add, so leaving it out of the REST API seems better.

azure/ConsiderationsForServiceDesign.md
mikekistler: I'll just remove it for now.

azure/ConsiderationsForServiceDesign.md
mikekistler: I'll remove guidance on schema names. We can put that in the Azure API Style Guide.

mikekistler:
Thank you for all the comments! I think I have addressed nearly all and made some other minor improvements. Please have another look.

One issue remaining is the naming convention for date-times. Do we want to recommend the "At" suffix or align with the MS and MS Graph guidelines to use a "DateTime" suffix?

azure/ConsiderationsForServiceDesign.md
```diff
@@ -77,6 +77,71 @@ Understanding how your service is used and defining its model and interaction pa
...
+
+Good names for resources, properties, operations, and parameters are essential to a good developer experience.
+
+Resources are described by nouns. Resource and property names must be descriptive and easy for customers to understand. Ease of understanding comes from familiarity and recognition; you should favor consistency with other Azure services, names in the product's portal/user interface, and industry standards.
+
+Names should aid developers in discovering functionality without having to constantly refer to documentation.
+Use common patterns and standard conventions to aid developers in correctly guessing common property names and meanings.
+Use verbose naming patterns and avoid abbreviations other than
+well-known acronyms in your service domain.
+
```
KrzysztofCwalina, line 100: I think it would be useful to add something about a common naming mistake I see: API designers often name concepts based on how the concepts work (architecture and design), not what the concepts are for (the user scenario), i.e. they name them from the perspective of the concept designer, not the user. For example, AppConfig service has a concept KeyValuePair. A better name for the same concept is ConfigurationSetting, as what thew user has in mind when they work with configation is settings. The kev-value pair is one way to design a setting system

KrzysztofCwalina, line 100: Another common problem I see is that too many concept names are included in APIs. When a service/feature is being designed, the designers need many concept names to be able to debate and reason about detail of the design. Once the feature design is finished, concepts that are not absolutely required to be understood by users should not appear in the finished product, but they often do, yet users are not interested in knowing how the sausage is made.

KrzysztofCwalina: When I see a concept heavy name, I ask the designer "what is blah-foo-bar-baz method doing?". They often describe the semantics using plain English, and when they do I tell them to name the API just what they said.

azure/ConsiderationsForServiceDesign.md
mikekistler: I think I have removed all the guidance on model/schema names.

mikekistler: I did an analysis of the API definitions currently in the repo and found that the `DateTime` suffix is more common than `At` -- and there were several other flavors but all less common than `DateTime`. So I will switch the guidance to that.

azure/ConsiderationsForServiceDesign.md
mikekistler: Good points Krzysztof. I have tried to incorporate these but happy to take additional feedback.

mikekistler:
Thanks for all the comments and I have attempted to incorporate all these. Will be happy for any additional comments or a thumbs up if folks think this is ready to merge.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -84,6 +84,74 @@ As you build out your service and API, there are a number of decisions that can
 
 :ballot_box_with_check: **YOU SHOULD** implement [conditional requests](https://tools.ietf.org/html/rfc7232) early. This allows you to support concurrency, which tends to be a concern later on.
 
+## Use Good Names
+
+Good names for resources, properties, operations, and parameters are essential to a good developer experience.
```
tg-msft: Let's swing for the fences.  :smile:

```suggestion
Good names for resources, properties, operations, and parameters are essential to a great developer experience.
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -84,6 +84,74 @@ As you build out your service and API, there are a number of decisions that can
...
+
+:ballot_box_with_check: **YOU SHOULD** should place the adjective before the noun in names that contain both a noun and an adjective.
+
+For example, `collectedItems` not `itemsCollected`
+
+:ballot_box_with_check: **YOU SHOULD** case all acronyms as though they were regular words (i.e. lower camelCase).
+
+For example, `nextUrl` not `nextURL`.
+
+:ballot_box_with_check: **YOU SHOULD** use "DateTime" suffix in names of `date-time` values.
```
tg-msft: This was done historically, but I really dislike it for Track 2 generated clients. I'd personally rather remove this now, but we could also approach it via code generator heuristics if folks want to see this on the wire.

azure/ConsiderationsForServiceDesign.md
tg-msft: I'd probably want `InDays` for generated client APIs in .NET.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -84,6 +84,74 @@ As you build out your service and API, there are a number of decisions that can
...
+
+:ballot_box_with_check: **YOU SHOULD** use an int for time durations and include the time units in the name.
+
+For example, `expirationDays` as `int` and not `expiration` as `date-time`.
+
+:warning: **YOU SHOULD NOT** use brand names in resource or property names.
+
+:warning: **YOU SHOULD NOT** use acronyms or abbreviations unless they are broadly understood for example, "ID" or "URL", but not "Num" for "number".
+
+:warning: **YOU SHOULD NOT** use names that are reserved words in widely used programming languages.
```
tg-msft:
```suggestion
:warning: **YOU SHOULD NOT** use names that are reserved words in widely used programming languages (including C#, Java, JavaScript/TypeScript, Python, C++, and Go).
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -84,6 +84,74 @@ As you build out your service and API, there are a number of decisions that can
...
+
+| Name | Description |
+|------------- | --- |
+| createdDateTime | The date-time that the resource was created |
+| updatedDateTime | The date-time that the resource was last updated/modified |
+| kind   | The discriminator value for a polymorphic resource |
+
+### `name` vs `id`
+
+:heavy_check_mark: **DO** use "Id" suffix for the name of rhe identifier of a resource.
```
tg-msft:
```suggestion
:heavy_check_mark: **DO** use "Id" suffix for the name of the identifier of a resource.
```

tg-msft: Do we want to add a recommendation for `description` when the value needs something human readable?

tg-msft (approved):
This looks like a great start

azure/ConsiderationsForServiceDesign.md
```diff
@@ -84,6 +84,74 @@ As you build out your service and API, there are a number of decisions that can
 
 :ballot_box_with_check: **YOU SHOULD** implement [conditional requests](https://tools.ietf.org/html/rfc7232) early. This allows you to support concurrency, which tends to be a concern later on.
 
+## Use Good Names
+
+Good names for resources, properties, operations, and parameters are essential to a great developer experience.
+
+Resources are described by nouns. Resource and property names must be descriptive and easy for customers to understand.
+Avoid names that refer to service implementation details, e.g. "TreeLeafNode", or simply describe the structure of value without
+describing its purpose, e.g. "KeyValuePair".
```
KrzysztofCwalina: It would be good to say what names should be used instead and give examples, e.g. use names that correspond to user scenarios (what the user is trying to acomplish), e.g. ConfigurationSetting instead of KeyValuePair.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -84,6 +84,74 @@ As you build out your service and API, there are a number of decisions that can
...
+
+For example, `expirationDays` as `int` and not `expiration` as `date-time`.
+
+:warning: **YOU SHOULD NOT** use brand names in resource or property names.
+
+:warning: **YOU SHOULD NOT** use acronyms or abbreviations unless they are broadly understood for example, "ID" or "URL", but not "Num" for "number".
+
+:warning: **YOU SHOULD NOT** use names that are reserved words in widely used programming languages (including C#, Java, JavaScript/TypeScript, Python, C++, and Go).
+
+:no_entry: **DO NOT** use "is" prefix in names of `boolean` values.
```
tjprescott: What should they use instead?

azure/ConsiderationsForServiceDesign.md
```diff
@@ -84,6 +84,74 @@ As you build out your service and API, there are a number of decisions that can
 
 :ballot_box_with_check: **YOU SHOULD** implement [conditional requests](https://tools.ietf.org/html/rfc7232) early. This allows you to support concurrency, which tends to be a concern later on.
 
+## Use Good Names
+
+Good names for resources, properties, operations, and parameters are essential to a great developer experience.
+
+Resources are described by nouns. Resource and property names must be descriptive and easy for customers to understand.
```
tjprescott, line 91: I think it would be good to call out that endpoints should be described by verbs. I've had at least once instance where a service team used a noun as an endpoint name and it would be good to point them to these guidelines.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -84,6 +84,74 @@ As you build out your service and API, there are a number of decisions that can
...
+Ease of understanding comes from familiarity and recognition; you should favor consistency with other Azure services, names in the product's portal/user interface, and industry standards.
+
+Names should aid developers in discovering functionality without having to constantly refer to documentation.
+Use common patterns and standard conventions to aid developers in correctly guessing common property names and meanings.
+Use verbose naming patterns and avoid abbreviations other than
+well-known acronyms in your service domain.
+
+:heavy_check_mark: **DO** use the same name for the same concept and different names for different concepts wherever possible.
+
+### Recommended Naming Conventions
```
tjprescott, line 103: How about the following:
"DO NOT use HTTP constructs in names (FooResult, FooRequest)".

azure/ConsiderationsForServiceDesign.md
heaths: It would be hard to list all possible alternatives given the number of distinct keywords, but maybe we should include at least some examples like how we prefer "kind" over "type" for discriminators.

azure/ConsiderationsForServiceDesign.md
mikekistler: Deferred to a future PR as agreed in API Stewardship meeting.

azure/ConsiderationsForServiceDesign.md
mikekistler: They should use NO prefix. I'll elaborate.

azure/ConsiderationsForServiceDesign.md
mikekistler: OperationIds are not part of the formal API so they are covered in the [OpenAPI Style Guide](https://github.com/Azure/azure-api-style-guide/blob/main/openapi-style-guide.md#operationid).

For Post actions where the action is the last segment of the path, this is covered in the [API Guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#user-content-actions-action-name-is-verb).

azure/ConsiderationsForServiceDesign.md
mikekistler: Good suggestion. I have fixed.

azure/ConsiderationsForServiceDesign.md
mikekistler: We originally included this guidance but removed it as it mainly pertains to schema names which are not part of the formal API contract.

But we should include this in a update to the OpenAPI Style Guide.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -116,8 +116,6 @@ For example, `collectedItems` not `itemsCollected`
 
 For example, `nextUrl` not `nextURL`.
 
-:ballot_box_with_check: **YOU SHOULD** use "DateTime" suffix in names of `date-time` values.
```
heaths: What should they use then? We see such a mix of different names, it's a bit of a quagmire from library to library of "Date", "DateTime", "Timestamp", "TimeStamp" (that's just wrong), and nothing. Surely, we could strive for some consistency.

mikekistler: Opened #409 to track.

PR #397, opened 2022-09-04
Add guidelines for graph conventions around read-only properties

corranrogue9:
I am adding a new file to document some of the graph conventions we have around properties. This PR is documenting the convention that read-only properties can be "updated" so long as they are being set to the same value that they currently have. 

Future changes to this file will contain conventions around default property sets, and conventions for the requiring `$select`, which I will go to the council meeting with proposals for. The current change is documenting something that is already established and has come up a few times in the last several months.

### DIFF
--- a/graph/propertyconventions.md
+++ b/graph/propertyconventions.md
@@ -1,0 +1,70 @@
+# Guidelines for the conventions of properties on graph types
+
+## Updating read-only properties
+
+### Convention
+
+It is a common call pattern for Microsoft Graph clients to `GET` an entity, make changes to a few properties, and then `PATCH` the same entity with the updated values.
+When clients do this, they often ignore that some properties are read-only or that a property may have a value that is allowed in a response, but not in a request.
+Because of this, workloads should support the practice that a request that updates a property should be considered a no-op if the value in the request matches the current value of the property.
+
+### Example
+
+For the following schema:
+```xml
+<EntityType Name="entity">
+  <Key>
+    <PropertyRef Name="id" />
+  </Key>
+  <Property Name="id" Type="Edm.String" Nullable="false" />
+  <Property Name="foo" Type="Edm.String" Nullable="false" /> <!--this property is read-only-->
+  <Property Name="bar" Type="Edm.String" Nullable="false" />
+</EntityType>
+```
+A client retrieves an `entity`:
+```http
+GET /entities/{id}
+
+HTTP/1.1 200 OK
+{
+  "id": "...",
+  "foo": "somevalue",
+  "bar": "someothervalue"
+}
+```
+If the client attempts to update `foo`, they will receive an error:
+```http
+PATCH /entities/{id}
+{
+  "foo": "somenewvalue"
+}
+
+HTTP/1.1 400 Bad Request
+{
+    "error": {
+        "code": "ReadOnlyProperty",
+        "message": "The property 'foo' is read-only and cannot be updated."
+    }
+}
+```
+However, if the client were to update `bar` while including `foo` but leaving it the same value, the request will be successful:
+```http
+PATCH /entities/{id}
+{
+  "foo": "somevalue",
+  "bar": "somenewvalue"
+}
+
+HTTP/1.1 204 No Content
+```
+A subsequent `GET` call would reflect the change in `bar`:
+```http
+GET /entities/{id}
+
+HTTP/1.1 200 OK
+{
+  "id": "...",
+  "foo": "somevalue",
+  "bar": "somenewvalue"
+}
+```


graph/propertyconventions.md
```diff
@@ -0,0 +1,70 @@
...
+### Example
+
+For the following schema:
+```xml
+<EntityType Name="entity">
+  <Key>
+    <PropertyRef Name="id" />
+  </Key>
+  <Property Name="id" Type="Edm.String" Nullable="false" />
+  <Property Name="foo" Type="Edm.String" Nullable="false" /> <!--this property is read-only-->
```
tombuildsstuff, line 20: Q: are there plans to add a `ReadOnly="true"` to the `Property` Node where applicable (or else how can clients know this is a readonly field)?

corranrogue9: Hello! There are three ways to indicate this in the CSDL:
1. The `Computed` [annotation](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.xml#L384)
2. The `Immutable` [annotation](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.xml#L380)
3. The `NonUpdatableProperties` [property](https://github.com/oasis-tcs/odata-vocabularies/blob/1fd9313b8c747db3707562f3a56c955aaf15434f/vocabularies/Org.OData.Capabilities.V1.xml#L713) of `UpdateRestrictionsType`

It's also worth noting what the [standard](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#_Toc31358979) says about this:
> Key and other properties marked as read-only in metadata [...] can be omitted from the request. If the request contains a value for one of these properties, the service MUST ignore that value when applying the update.

Our behavior in graph doesn't actually match any of these. In graph, if a read-only property is provided, we check that it has the same value as the current property; if it does, we ignore it, and if it doesn't we produce an error. As a result of investigating all of this (thanks for making me look into it!), we have created [this](https://issues.oasis-open.org/browse/ODATA-1543) issue with the OASIS technical committee to discuss how to represent this behavior in OData. Most likely, we will update the `Immutable` annotation to be more flexible, changing it from always ignoring the property when provided to allowing it to be validated if provided.

PetRich-MSFT, line 20: We should include one (or more) of the annotations to show how to correctly mark a property as read-only, otherwise we give the impression that workload authors don't need to mark their properties

graph/propertyconventions.md
```diff
@@ -0,0 +1,70 @@
+# Guidelines for the conventions of properties on graph types
+
+## Updating read-only properties
+
+### Convention
+
+It is a common call pattern for Microsoft Graph clients to `GET` an entity, make changes to a few properties, and then `PATCH` the same entity with the updated values.
+When clients do this, they often ignore that some properties are read-only or that a property may have a value that is allowed in a response, but not in a request.
+Because of this, workloads should support the practice that a request that updates a property should be considered a no-op if the value in the request matches the current value of the property.
+
```
OlgaPodo, line 10: Can we please use our standard template https://github.com/microsoft/api-guidelines/blob/vNext/graph/patterns/PatternDescriptionTemplate.md to show what design problem we are trying to solve and examples of solving it.

graph/propertyconventions.md
```diff
@@ -0,0 +1,70 @@
+# Guidelines for the conventions of properties on graph types
+
+## Updating read-only properties
```
PetRich-MSFT, line 3: We should also describe the behavior when creating an entity with read-only properties

graph/propertyconventions.md
corranrogue9: I don't think workload authors currently do need to mark their properties. I did a spot check on some read-only properties and don't see them marked or a suppression indicating that they were "grandfathered in". From the discussion [elsewhere](https://github.com/microsoft/api-guidelines/pull/397#discussion_r970051574), OData doesn't currently have a one-size-fits-all annotation, and graph doesn't conform to any of the existing annotations. I'd rather not use an annotation that graph isn't conforming to when writing the guidance to explain in what ways graph doesn't conform.

graph/propertyconventions.md
corranrogue9: I was thinking to add this in a future PR. I don't know what we expect for creates, but I got a complete answer for updates, so I wanted to get it written down. 

If this is a blocker for you, I will work to understand the create behavior for this PR, otherwise I would like to update the document in the future for that use-case.

corranrogue9:
Since this is less "guidance" and more documentation of how a specific set of graph workloads function/are expected to function, I am going to close this PR and add this documentation in a different location. I intend to do further investigation to have unifying guidance on graph (for creates in addition to updates), but I don't have the time to do that write now.

PR #405, opened 2022-10-11
Remove guidance for two-letter acronyms

### LINKED ISSUE #404: Two letter acronym guidance
In our [naming guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/graph/GuidelinesGraph.md#naming), we say this about two-letter acronyms:

> ☑️ SHOULD case two-letter acronyms with the same case.

We also don't say anything specific about two-letter abbreviations. I think for ID (which is arguable if it's an acronym or an abbreviation, but I don't think it matters here) we have the general pattern of something like `myId` rather than `myID`. I also think that we do the same for other two-letter tokens. I suggest that we either remove this naming convention, or we specifically call out that ID should not follow this pattern.

corranrogue9:
To resolve [this](https://github.com/microsoft/api-guidelines/issues/404) issue, I am removing the guidance in our naming conventions that two-letter acronyms should have both characters capitalized. We don't seem to follow this guidance, and it causes confusion about "ID" being in a name (i.e. should it be `myID` or `myId`).

### DIFF
--- a/graph/GuidelinesGraph.md
+++ b/graph/GuidelinesGraph.md
@@ -114,7 +114,6 @@ Following is a short summary of the most often used conventions.
 | :ballot_box_with_check: **SHOULD** pluralize the noun even when followed by an adjective (a *postpositive*).| - **Right:** passersby or mothersInLaw    <BR> -  **Wrong:** notaryPublics or motherInLaws |
 | **CASING** | |
 | :heavy_check_mark: **MUST** use lower camel case for *all* names and namespaces.   | - **Right:** automaticRepliesStatus <BR> - **Wrong:** kebab-case or snake_case |
-| :ballot_box_with_check: **SHOULD** case two-letter acronyms with the same case.   | - **Right:** ioLimit or totalIOAmount <BR> - **Wrong:** iOLimit or totalIoAmount |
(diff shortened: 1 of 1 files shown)


graph/GuidelinesGraph.md
```diff
@@ -114,7 +114,6 @@ Following is a short summary of the most often used conventions.
 | :ballot_box_with_check: **SHOULD** pluralize the noun even when followed by an adjective (a *postpositive*).| - **Right:** passersby or mothersInLaw    <BR> -  **Wrong:** notaryPublics or motherInLaws |
 | **CASING** | |
 | :heavy_check_mark: **MUST** use lower camel case for *all* names and namespaces.   | - **Right:** automaticRepliesStatus <BR> - **Wrong:** kebab-case or snake_case |
-| :ballot_box_with_check: **SHOULD** case two-letter acronyms with the same case.   | - **Right:** ioLimit or totalIOAmount <BR> - **Wrong:** iOLimit or totalIoAmount |
 | :ballot_box_with_check: **SHOULD** case three+ letter acronyms the same as a normal word.  | - **Right:** fidoKey or oauthUrl <BR> - **Wrong:** webHTML |
```
dkershaw10, line 117: If we remove the line about two-letter acronyms, then do we need the one about three+?  Should we change it to two+?

OlgaPodo:
@corranrogue9  I think that the auxiliary verb SHOULD indicates a guideline that API designers should generally follow, but there are exceptions when the rule does not apply. I think that removing this rule would create uncertainty. I would like to close this PR without merging.

PR #412, opened 2022-10-30
Add naming guidance for date-time values

mikekistler:
This is a follow-on PR to #391 to tackle the issue of naming guidelines for date-time values.

The text in this initial commit is consistent with the original Microsoft API guidelines and with the majority of current Azure services.

### DIFF
--- a/azure/ConsiderationsForServiceDesign.md
+++ b/azure/ConsiderationsForServiceDesign.md
@@ -116,6 +116,8 @@ For example, `collectedItems` not `itemsCollected`
 
 For example, `nextUrl` not `nextURL`.
 
+:ballot_box_with_check: **YOU SHOULD** use "DateTime" suffix in names of `date-time` values.
+
 :ballot_box_with_check: **YOU SHOULD** use a suffix of the unit of measurement for values with a clear unit of measurement (such as bytes, miles, and so on). Use a generally accepted abbreviation for the units (e.g. "Km" rather than "Kilometers") when appropriate.
 
 :ballot_box_with_check: **YOU SHOULD** use an int for time durations and include the time units in the name.
--- a/azure/ConsiderationsForServiceDesign.md
+++ b/azure/ConsiderationsForServiceDesign.md
@@ -118,6 +118,16 @@ For example, `nextUrl` not `nextURL`.
 
 :ballot_box_with_check: **YOU SHOULD** use "DateTime" suffix in names of `date-time` values.
 
+For example, `createdDateTime` not `createdAt` or `createdTime`. Common date-time properties and their recommended names are:
+
+| Property Name | Description |
+| --- | --- |
+| createdDateTime | The date and time the resource was created. |
+| lastModifiedDateTime | The date and time the resource was last modified. |
+| deletedDateTime | The date and time the resource was deleted. |
+| startDateTime | The starting date and time of an interval. |
+| endDateTime | The ending date and time of an interval. |
+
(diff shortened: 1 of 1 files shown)


DominikMe:
Can we add an example for a well-named property? I know in ACS we have a couple of `expiresAt`, `validFrom`, `createdAt` and I'm not sure if `expiresAtDateTime`, `validFromDateTime`, `createdAtDateTime` would add value or add more noise that makes them harder to understand at a glance.

azure/ConsiderationsForServiceDesign.md
```diff
@@ -116,6 +116,8 @@ For example, `collectedItems` not `itemsCollected`
 
 For example, `nextUrl` not `nextURL`.
 
+:ballot_box_with_check: **YOU SHOULD** use "DateTime" suffix in names of `date-time` values.
+
```
JeffreyRichter, line 122: Personally, I prefer At suffix. But I can live with DateTime

mikekistler: If we have a consistent naming convention at the REST API level, we can choose to impose a different naming convention at the SDK level. So if you want to use the "At" suffix in Go SDKs -- if you think that would be idiomatic for Go -- then the generator can just make that substitution.

heaths: I doubt a "DateTime" suffix would pass muster in any SDKs, right? With a lot of models in a response, using "At" or "On" will decrease the payload size.

Personally, I like "On" better. It works in more places than "At" e.g., "createdOn", "modifiedOn" or even "lastModifiedOn", "deletedOn", etc. This is what @KrzysztofCwalina had us use in track 2 wave 1 clients like Key Vault e.g., `CreatedOn`.

tg-msft: I'd also prefer `On` or `At` to `DateTime`.  We're not putting any other type as a naming suffix so I don't see a benefit here.  Even if we wanted to make the argument it's because these are sent as strings over the wire, we don't use `ageInt` for formatting the values of numeric types, etc.

stewartadam: +1 for `On` or `At` suffix. I think we should do a cursory check on existing datetime fields in the API specs to confirm it makes semantic sense for a broad variety of cases, but besides the payload advantages I like it better than `DateTime` type suffix which would be one-off compared to other types.

mikekistler: This page has an interesting take on "on vs at":

https://www.differencebetween.com/difference-between-on-and-at/

Based on this, I think I would prefer the "At" suffix for `format: date-time` values.

Should we recommend "On" for `format: date` values?

heaths: But our timestamps do include time, hence - even by that site's definition - "on" makes more sense. When, for example, have you heard someone say "modified at" or "deleted at"? According to Websters,

"On" _used as a function word to indicate a time frame during which something takes place_
"At" _used as a function word to indicate age or position in time_

I think switching between "on" vs. "at", though, is confusing. The nuance of one vs the other I imagine would be lost on many developers.

tg-msft (approved):

azure/ConsiderationsForServiceDesign.md
tellarin: I agree "At" is probably the best choice for the typical DateTime case. But, as mentioned by @DominikMe ([here](https://github.com/microsoft/api-guidelines/pull/412#issuecomment-1386262617)), there are cases where the semantics benefits from terms like "From" or "To". Are these cases covered elsewhere? If not, there should be a note here.

mikekistler: We don't cover cases that might benefit from "From" or "To".  The guidance for "At" is "YOU SHOULD" rather than "DO", so that allows other choices when appropriate. If you think we need something more specific, what would you propose?

PR #417, opened 2022-11-16
Casing and expand

OlgaPodo:
updates for error and expand in guidelines, change tracking pattern update

### DIFF
--- a/graph/GuidelinesGraph.md
+++ b/graph/GuidelinesGraph.md
@@ -166,7 +166,7 @@ Microsoft Graph APIs should support basic query options in conformance with ODat
 |Requirements                                                                                        |
 |----------------------------------------------------------------------------------------------------|
 | :heavy_check_mark: **MUST** support `$select on resource` to enable properties projection. |
-| :ballot_box_with_check: **SHOULD** support `$expand=navProp1` option for navigation properties of entities. |
+| :ballot_box_with_check: **SHOULD** support `/entityTypeCollection/{id}?$expand=navProp1` option for navigation properties of entities. |
 | :ballot_box_with_check: **SHOULD** support `$filter` with `eq` and `ne` operations on properties of entity collections. |
 | :heavy_check_mark: **MUST** support [server-driven pagination](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#981-server-driven-paging) of collections using a [nextLink](http://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#sec_ControlInformationnextLinkodatanextL).  |
 | :ballot_box_with_check: **SHOULD** support [client-driven pagination](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#982-client-driven-paging) of collections using `$top` and `$skip` (or `$skipToken`). |
--- a/graph/GuidelinesGraph.md
+++ b/graph/GuidelinesGraph.md
@@ -166,7 +166,8 @@ Microsoft Graph APIs should support basic query options in conformance with ODat
 |Requirements                                                                                        |
 |----------------------------------------------------------------------------------------------------|
 | :heavy_check_mark: **MUST** support `$select on resource` to enable properties projection. |
-| :ballot_box_with_check: **SHOULD** support `\$filter with eq`, `ne` operations on properties of entities for collections. |
+| :ballot_box_with_check: **SHOULD** support `$expand=navProp1` option for navigation properties of entities. |
+| :ballot_box_with_check: **SHOULD** support `$filter with eq`, `ne` operations on properties of entities for collections. |
 | :heavy_check_mark: **MUST** support [server-driven pagination](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#981-server-driven-paging) of collections using a [nextLink](http://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#sec_ControlInformationnextLinkodatanextL).  |
 | :ballot_box_with_check: **SHOULD** support [client-driven pagination](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#982-client-driven-paging) of collections using `$top` and `$skip` (or `$skipToken`). |
 | :ballot_box_with_check: **SHOULD** support `$count` for collections. |
@@ -184,7 +185,7 @@ You can model structured resources for your APIs by using the OData entity type 
 Because objects of complex types in Microsoft Graph don’t have unique identifiers, they are not directly addressable via URIs. Therefore, you must not use complex type to model addressable resources such as individually addressable items within a collection. For more information, see the [Microsoft REST API Guidelines collection URL patterns](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#93-collection-url-patterns). Complex types are better suited to represent composite properties of API entities.
 
 ```xml
- <EntityType Name="Author">
+ <EntityType Name="author">
     <Key>
         <PropertyRef Name="id" />
     </Key>
@@ -192,7 +193,7 @@ Because objects of complex types in Microsoft Graph don’t have unique identifi
     <Property Name="name" Type="Edm.String" />
     <Property Name="address" Type="microsoft.graph.Address" />
 </EntityType>
-<ComplexType Name="Address">
+<ComplexType Name="address">
     <Property Name="city" Type="Edm.String" />
     <Property Name="street" Type="Edm.String" />
     <Property Name="stateOrProvince" Type="Edm.String" />
@@ -273,11 +274,11 @@ Microsoft REST API Guidelines provide guidelines that Microsoft Graph APIs shoul
 ```http
 {
 "error": {
-    "code": "BadRequest",
+    "code": "badRequest",
     "message": "Cannot process the request because a required field is missing.",
     "target": "query",    
     "innererror": {
-                "code": "RequiredFieldMissing",
+                "code": "requiredFieldMissing",
                            
                 }
     }
@@ -291,9 +292,9 @@ The top-level error code must be aligned with HTTP response status codes accordi
    ```http
     {
       "error": {
-        "code": "BadRequest",
+        "code": "badRequest",
         "message": "Cannot process the request because it is malformed or incorrect.",
-    	"target": "Resource X (Optional)"
+    	"target": "resource X (Optional)"
       }
     }
    ```
@@ -303,7 +304,7 @@ The top-level error code must be aligned with HTTP response status codes accordi
    ```http
     {
       "error": {
-        "code": "BadRequest",
+        "code": "badRequest",
         "message": "Cannot process the request because it is malformed or incorrect.",
         "innererror": {
           "code": "requiredFieldOrParameterMissing",
--- a/graph/GuidelinesGraph.md
+++ b/graph/GuidelinesGraph.md
@@ -182,7 +182,7 @@ Another way to avoid this is to use JSON batch as described in the [Microsoft Gr
 
 You can model structured resources for your APIs by using the OData entity type or complex type. The main difference between these types is that an entity type declares a key property to uniquely identify its objects, and a complex type does not. In Microsoft Graph, this key property is called `id` for server-created key values. If there is a natural name for the key property, then the workload can use that.
 
-Because objects of complex types in Microsoft Graph don’t have unique identifiers, they are not directly addressable via URIs. Therefore, you must not use complex type to model addressable resources such as individually addressable items within a collection. For more information, see the [Microsoft REST API Guidelines collection URL patterns](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#93-collection-url-patterns). Complex types are better suited to represent composite properties of API entities.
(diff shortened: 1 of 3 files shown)


graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,161 @@
...
+*The change tracking pattern provides the ability for API consumers to request changes in data from Microsoft Graph without having to re-read data that has not changed.*
+
+
+## Problem
+
+API consumers require an efficient way to acquire changes to data in the Microsoft Graph, for example to synchronize an external store or to drive a change-centric business process.
+
+## Solution
+
+API designers can enable the change tracking (delta) capability on a resource in the Microsoft Graph (typically on an entity collection or a parent resource) by declaring a delta function on that resource.
```
OlgaPodo: @garethj-msft   IMO we need to clarify that change tracking is enabled by declaring a delta function and applying ChangeTracking annotation to the targeted resource.

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,161 @@
...
+  
+
+## Alternatives
+
+- Change notifications pattern with rich payloads – for use cases where API consumers would find calling back into Microsoft Graph onerous and absolute integrity guarantees are less critical.
+
+
+## Examples
+
+### Change tracking on entity set
```
OlgaPodo, line 82: @garethj-msft I think we need to add an example of ChangeTracking on an Entities as it is one of most used patterns

graph/GuidelinesGraph.md
```diff
@@ -166,7 +166,8 @@ Microsoft Graph APIs should support basic query options in conformance with ODat
 |Requirements                                                                                        |
 |----------------------------------------------------------------------------------------------------|
 | :heavy_check_mark: **MUST** support `$select on resource` to enable properties projection. |
-| :ballot_box_with_check: **SHOULD** support `\$filter with eq`, `ne` operations on properties of entities for collections. |
+| :ballot_box_with_check: **SHOULD** support `$expand=navProp1` option for navigation properties of entities. |
+| :ballot_box_with_check: **SHOULD** support `$filter with eq`, `ne` operations on properties of entities for collections. |
```
```suggestion
| :ballot_box_with_check: **SHOULD** support `$filter` with `eq` and `ne` operations on properties of entity collections. |
```
Question - do we think `ne` is an important operation?  For me I wonder whether `lt` and `gt` are more important for DateTime properties.

graph/GuidelinesGraph.md
```diff
@@ -166,7 +166,8 @@ Microsoft Graph APIs should support basic query options in conformance with ODat
 |Requirements                                                                                        |
 |----------------------------------------------------------------------------------------------------|
 | :heavy_check_mark: **MUST** support `$select on resource` to enable properties projection. |
-| :ballot_box_with_check: **SHOULD** support `\$filter with eq`, `ne` operations on properties of entities for collections. |
+| :ballot_box_with_check: **SHOULD** support `$expand=navProp1` option for navigation properties of entities. |
```
dkershaw10: Is this saying should have for:

- `GET /entityTypeCollection/{id}?$expand=navProp1` and/or
- `GET /entityTypeCollection?$expand=navProp1`

The latter is more expensive for some resources, so not sure if that's a should or a nice-to-have

graph/GuidelinesGraph.md
```diff
@@ -184,15 +185,15 @@ You can model structured resources for your APIs by using the OData entity type
 Because objects of complex types in Microsoft Graph don’t have unique identifiers, they are not directly addressable via URIs. Therefore, you must not use complex type to model addressable resources such as individually addressable items within a collection. For more information, see the [Microsoft REST API Guidelines collection URL patterns](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#93-collection-url-patterns). Complex types are better suited to represent composite properties of API entities.
```
dkershaw10: Should we be recommending use of entity types here in this case?
```suggestion
Because objects of complex types in Microsoft Graph don’t have unique identifiers, they are not directly addressable via URIs. Therefore, you must use entity types to model addressable resources such as individually addressable items within a collection. For more information, see the [Microsoft REST API Guidelines collection URL patterns](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#93-collection-url-patterns). Complex types are better suited to represent composite properties of API entities.
```

graph/patterns/change-tracking.md
dkershaw10: We're also about to release support for change tracking on a function - but I guess we can update the doc once that's officially released.

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Solution
+
+API designers can enable the change tracking (delta) capability on a resource in the Microsoft Graph (typically on an entity collection or a parent resource) by declaring a delta function on that resource.
+
+This function returns a delta payload. A delta payload consists of a collection of annotated full or partial Microsoft Graph entities plus either a `nextLink` to further pages of original or change data that are immediately available OR a `deltaLink` to poll to get the next set of changes as they occur.
+
+Annotations allow the delta payload to indicate resources or links which have been deleted. API callers are expected to differentiate resource adds from updates by interpreting the id property of the change records against the existence of resources in whatever external system is doing the processing.
+
+The pattern requires a sequence of requests.
```
dkershaw10: The pattern bellow doesn't seem correct.  Please see the details here: https://learn.microsoft.com/en-us/graph/delta-query-overview?tabs=http#use-delta-query-to-track-changes-in-a-resource-collection

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,161 @@
...
+API consumers want a pull mechanism to request and process change to Microsoft Graph data, either via proactive polling or by responding to Microsoft Graph notifications.
+
+API consumers need guaranteed data integrity over the set of changes to Microsoft Graph data.
+
+## Issues and considerations
+
+ - API service MAY be able to respond to standard OData query parameters with the initial call to the delta function:
+
+    - `$select` to enforce the set of properties on which change is reported.
+    - `$filter` to influence the range of changes returned.
```
dkershaw10: "range" or "scope"?  Range makes it sound like a duration range, but we have other supported filters that are not duration based.

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,161 @@
...
+        </NavigationProperty>
+</EntityType>
+```
+
+### Delta payload
+
+ Here, a user resource is updated, and there is one user added to and one removed from that user’s directReports collection. Additionally, a second user is deleted. In this case, there are no further pages of change records currently available.
+
+```
+GET https://graph.microsoft.com/v1.0/users/delta
```
dkershaw10: Shouldn't this request have a $deltaToken (based on the $odata.deltaLink property returned previously) as you are showing deltas in the response?

dkershaw10 (changes requested):
Added some initial comments and suggested changes

graph/patterns/change-tracking.md
OlgaPodo: Add ChangeTracking annotation

graph/patterns/change-tracking.md
OlgaPodo: Maybe I'm missing something, but I believe it's very close: 1. delta no param, 2. optional if there is nextLink 3. update on delta with delta token 4. optional if there is nextLink. added a link to pub docs

OlgaPodo: @dkershaw10 could you please elaborate on what is the main deviation besides ability to use latest?

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Problem
+
+API consumers require an efficient way to acquire changes to data in the Microsoft Graph, for example to synchronize an external store or to drive a change-centric business process.
+
+## Solution
+
+API designers can enable the change tracking (delta) capability on a resource in the Microsoft Graph (typically on an entity collection or a parent resource) by declaring a delta function on that resource and applying `Org.OData.Capabilities.V1.ChangeTracking` annotation.
+
+This function returns a delta payload. A delta payload consists of a collection of annotated full or partial Microsoft Graph entities plus either a `nextLink` to further pages of original or change data that are immediately available OR a `deltaLink` to poll to get the next set of changes as they occur.
```
dkershaw10:
```suggestion
This function returns a delta payload. A delta payload consists of a collection of annotated full or partial Microsoft Graph entities plus either a `nextLink` to further pages of original or change data that are immediately available OR a `deltaLink` to get the next set of changes at some later date.
```

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,163 @@
...
+Annotations allow the delta payload to indicate resources or links which have been deleted. API callers are expected to differentiate resource adds from updates by interpreting the id property of the change records against the existence of resources in whatever external system is doing the processing.
+
+The pattern requires a sequence of requests on the delta function, for additional details see [Change Tracking](https://learn.microsoft.com/en-us/graph/delta-query-overview?tabs=http#use-delta-query-to-track-changes-in-a-resource-collection):
+
+  1. GET request which returns the first page of the current state of the resources that delta applies to.  
+  2. [Optionally] Further GET requests to retrieve more pages of the current state via the `@odata.nextLink` URL.
+  3. After some time, a GET request to see if there are new changes via the `@odata.deltaLink` URL.
+  4. [Optionally] GET requests to retrieve more pages of changes via the `@odata.nextLink` URL.
+
+The `nextLink` provides a mechanism to do server-driven paging through the change data that is currently available.  When there are no further pages of changes immediately available, a `deltaLink` is returned instead.
```
dkershaw10: I think you should move lines 27 and 29 up before line 18.  That provide better context about nextLink and deltaLink that makes the sequence described in the enumerated list above clearer.

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,163 @@
...
+
+The `deltaLink` provides a mechanism for the API consumer to catch up on changes since their last request to the delta function, `deltaLink` or last page’s `nextLink`. If no changes have happened since the last request, then the deltaLink MUST return an empty collection.
```
dkershaw10:
```suggestion
The `deltaLink` provides a mechanism for the API consumer to catch up on changes since their last request to the delta function. If no changes have happened since the last request, then the deltaLink MUST return an empty collection.
```

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,163 @@
...
+- The delta function can be bound to a collection, as with
+`/users/delta` that returns the changes to the users' collection
+  or to some logical parent resource, where the change records are implied to be relative to all collections contained within the parent, for example
+  `/me/planner/all/delta` – this returns changes to any resource within planner that a user is subscribed to as a heterogenous collection.
+- API producers might use `$skipToken` and `$deltaToken` within their implementations of `nextLink` and `deltaLink`, however the URLs are defined as being opaque and the existence of the tokens MUST NOT be documented.   It is not a breaking change to modify the structure of `nextLinks` or `deltaLinks`.-
```
dkershaw10: I'm not sure if this is true.  AGS looks for and encodes/decodes the skip and delta tokens. I think this is a *should/must* rather than *might*.

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,163 @@
...
+
+- Change notifications pattern with rich payloads – for use cases where API consumers would find calling back into Microsoft Graph onerous and absolute integrity guarantees are less critical.
+
+
+## Examples
+
+### Change tracking on entity set
+
+```
+<Function Name="delta" IsBound="true" ags:OwnerService="Microsoft.DirectoryServices">
```
dkershaw10: Should these be the **public** CSDLs rather than workload CSDLs?
If so remove any ags annotations, and coerce to the microsoft.graph namespace for type names.

graph/patterns/change-tracking.md
```diff
@@ -0,0 +1,163 @@
...
+
+    - `$select` to enforce the set of properties on which change is reported.
+    - `$filter` to influence the scope of changes returned.
+    - `$expand` to include linked resources with the set of changes.
+    - `$top` parameter to influence the size of the set of change records.
+  
+    These query parameters MUST be encoded into subsequent `@odata.nextLink` or `@odata.deltaLink`, such that the same options are preserved through the call sequence without callers respecifying them, which MUST NOT be allowed. OData query parameters must be honored in full, or a 400-error returned.
+- Making a sequence of calls to a delta function followed by the opaque URLs in the `nextLink` and `deltaLink` MUST guarantee that the data at the start time of the call sequence and all changes to the data thereafter will be returned at least once. It is not necessary to avoid duplicates in the sequence. When the delta function is returning changes, they MUST be sequenced chronologically refer to [public documentation](https://learn.microsoft.com/en-us/graph/delta-query-overview?view=graph-rest-1.0) for more details.
+- The delta function can be bound to a collection, as with
```
dkershaw10: I can also be bound to a function that returns an entity collection.  See [Pull Request 7242963](https://dev.azure.com/msazure/One/_git/AD-AggregatorService-Wiki/pullrequest/7242963): Add info on change tracking for functions, for an example, including the schema updates.

dkershaw10 (changes requested):
The guideline changes LGTM
I added comments for the change tracking topic.  I like how this is laid out, way more than the wiki.

graph/patterns/change-tracking.md
OlgaPodo: updated

graph/patterns/change-tracking.md
OlgaPodo: Actually the "some logical parent resource" phrase allows for a function but i also added an example and formatting to highlight options.

PR #418, opened 2022-11-16
Clarify allowed/expected client modification of LRO api-version

mikekistler:
This PR just adds a note to explicitly state that clients may modify the value of `api-version` that is present in the `Operation-Location` URI.

### DIFF
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -1043,6 +1043,8 @@ For all long-running operations, the client will issue a GET on a status monitor
 <a name="lro-status-monitor-accepts-any-api-version"></a>
 :ballot_box_with_check: **YOU SHOULD** allow any valid value of the `api-version` query parameter to be used in the get operation on the status monitor.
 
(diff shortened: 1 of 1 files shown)


azure/Guidelines.md
```diff
@@ -1033,6 +1033,8 @@ For a PUT (create or replace) with additional long-running processing:
 <a name="lro-put-operation-location-includes-api-version"></a>
 :ballot_box_with_check: **YOU SHOULD** include the `api-version` query parameter in the `Operation-Location` header with the same version passed on the initial request if it is required by the get operation on the status monitor.
 
+Note: Clients may replace the value of `api-version` in the `Operation-Location` URI with a value appropriate for their application.
```
```suggestion
Note: Clients should replace the value of `api-version` in the `Operation-Location` URI with a value appropriate for their application when calling back to the service.
```

mikekistler: Chatted with Jeffrey offline and he (begrudingly) agreed to leave the sentence as-is.

markweitzel (approved):
LGTM. Thanks!

PR #419, opened 2022-11-23
Add resourceLocation to operation resource

johanste:
It seems like we are missing the `resourceLocation` from the operations resource example. Added.


mikekistler (commented):
My recollection is that we intentionally omitted `resourceLocation` because we do not want to appear to condone the use of long-running POST for resource creation. In fact we explicitly forbid that pattern:

> ⛔ DO NOT use a long-running POST to create a resource -- use PUT as described below.

Let's not send mixed signals. We do explicitly allow additional properties in the status monitor, so for the _hopefully rare_ cases where `resourceLocation` is needed it can be added.

markweitzel:
Reviewed in API Stewardship LT meeting on 21-Mar-23. We do not accept this change.

PR #422, opened 2022-12-21
Return 200 from POST Actions

mikekistler:
This PR add guidance to use a 200 response from POST Actions even if there is currently no content to return. In this case, the operation should return an empty body.


JeffreyRichter (approved):

weidongxu-microsoft (approved):

markweitzel (approved):
LGTM

PR #433, opened 2023-02-13
Add guidelines for polymorphism

### LINKED ISSUE #331: Add guidance on patching polymorphic types
Add guidance on how to patch a polymorphic type (it is a bit tricky).

Clipped from email stream:

@bexxx says:

We discussed to change the ScaleSettings type of a deployment from a flat type to one using polymorphism to be future proof (in case we add a second type like “TrafficBased”…) and use the ScaleType as the type discriminator.
I did the changes and for GET and POST it looks ok.
 
Now we also have a PATCH with content-type Application/merge-patch+json. Here all properties used to be optional to allow the request to only submit the properties whose values change.
To enable the use of polymorphism here as well I need to make the scale_type (discriminator) required to know what type the user wants to patch, is this the accepted way?
Or should we do smarter validation and deserialization on the service to keep everything optional as well.
 
@johanste response

This is the prevailing guidance in this scenario has been:
 
Discriminator/kind is not required – and in fact, usually not allowed to change. This is the default answer.
If you do allow for change of kind, it is indeed service specific, and only properties that map 1:1 between the old and new kind/type are migrated. You may fail the request with a 409 response if one or more properties of the existing resource cannot be migrated to the new kind. The principle of least surprise is applied on a case-by-case basis. If you are changing to a kind that is effectively a superset of the existing kind, this can make sense.
If you do not allow for kind, and you get a request to modify properties not available in the current kind, you return 409 Conflict indicating that the resource cannot be updated like this.
 
… but as always, feedback is welcome…
 
On a side note, this problem exists whenever you have interdependency between property values in a resource. Discriminated polymorphic types is one instance where this happens. A common other case is that only certain SKUs of a resource allow for specific configurations to be set.

mikekistler:
This PR adds guidelines for polymorphism and specifically how polymorphism should work in versioned APIs.

### DIFF
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -509,9 +509,9 @@ This indicates to client libraries and customers that values of the enumeration 
 If you can't avoid them, then follow the guideline below.
 
 <a name="json-use-discriminator-for-polymorphism"></a>
-:white_check_mark: **DO** define a `kind` field indicating the kind of the resource and include any kind-specific fields in the body.
+:white_check_mark: **DO** define a discriminator field indicating the kind of the resource and include any kind-specific fields in the body.
 
-Below is an example of JSON for a Rectangle and Circle:
+Below is an example of JSON for a Rectangle and Circle with a discriminator field named `kind`:
 **Rectangle**
 ```json
 {
@@ -546,11 +546,13 @@ Below is an example of JSON for a Rectangle and Circle:
 ```
 Both Rectangle and Circle have common fields: `kind`, `fillColor`, `lineColor`, and `subscription`. A Rectangle also has `x`, `y`, `width`, and `length` while a Circle has `x`, `y`, and `radius`. The `subscription` is a nested polymorphic type. A `free` subscription has no additional fields and a `paid` subscription has `expiration` and `invoice` fields.
 
+The [Azure Naming Guidelines](./ConsiderationsForServiceDesign.md#common-names) recommend that the discriminator field be named `kind`.
+
 <a name="json-polymorphism-kind-extensible"></a>
-:ballot_box_with_check: **YOU SHOULD** define the kind field of a polymorphic type to be an extensible enum.
+:ballot_box_with_check: **YOU SHOULD** define the discriminator field of a polymorphic type to be an extensible enum.
 
 <a name="json-polymorphism-kind-immutable"></a>
-:warning: **YOU SHOULD NOT** allow an update (patch) to change the kind field of a polymorphic type.
+:warning: **YOU SHOULD NOT** allow an update (patch) to change the discriminator field of a polymorphic type.
 
 <a name="json-polymorphism-versioning"></a>
 :warning: **YOU SHOULD NOT** return properties of a polymorphic type that are not defined for the api-version specified in the request.
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -556,7 +556,7 @@ Both Rectangle and Circle have common fields: `kind`, `fillColor`, `lineColor`, 
 :warning: **YOU SHOULD NOT** return properties of a polymorphic type that are not defined for the api-version specified in the request.
 
 <a name="json-polymorphism-arrays"></a>
-:warning: **YOU SHOULD NOT** specify a polymorphic type as the type of an array property.
+:warning: **YOU SHOULD NOT** have a property whose value is an array of polymorphic objects.
 
 ## Common API Patterns
 
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -551,15 +551,8 @@ The [Azure Naming Guidelines](./ConsiderationsForServiceDesign.md#common-names) 
 <a name="json-polymorphism-kind-extensible"></a>
 :ballot_box_with_check: **YOU SHOULD** define the discriminator field of a polymorphic type to be an extensible enum.
 
-**WE SHOULD CHOOSE ONE OF THE FOLLOWING TWO OPTIONS**
-
 <a name="json-polymorphism-kind-immutable"></a>
 :warning: **YOU SHOULD NOT** allow an update (patch) to change the discriminator field of a polymorphic type.
-
-**OR**
-
-<a name="json-polymorphism-mutable-kind"></a>
-:ballot_box_with_check: **YOU SHOULD** remove all properties specific to the old discriminator value from the resource when an update (patch) changes the discriminator field.
 
 <a name="json-polymorphism-versioning"></a>
 :warning: **YOU SHOULD NOT** return properties of a polymorphic type that are not defined for the api-version specified in the request.
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -492,6 +492,12 @@ This indicates to client libraries and customers that values of the enumeration 
 <a name="json-document-extensible-enums"></a>
 :white_check_mark: **DO** document to customers that new values may appear in the future so that customers write their code today expecting these new values tomorrow.
 
+<a name="json-return-extensible-enum-value"></a>
+:heavy_check_mark: **YOU MAY** return a value for an extensible enum that is not one of the values defined for the api-version specified in the request.
+
+<a name="json-accept-extensible-enum-value"></a>
+:warning: **YOU SHOULD NOT** accept a value for an extensible enum that is not one of the values defined for the api-version specified in the request.
+
 <a name="json-removing-enum-value-is-breaking"></a>
 :no_entry: **DO NOT** remove values from your enumeration list as this breaks customer code.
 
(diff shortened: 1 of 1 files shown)


azure/Guidelines.md
```diff
@@ -492,6 +492,12 @@ This indicates to client libraries and customers that values of the enumeration
...
 :white_check_mark: **DO** document to customers that new values may appear in the future so that customers write their code today expecting these new values tomorrow.
 
+<a name="json-return-extensible-enum-value"></a>
+:heavy_check_mark: **YOU MAY** return a value for an extensible enum that is not one of the values defined for the api-version specified in the request.
+
+<a name="json-accept-extensible-enum-value"></a>
+:warning: **YOU SHOULD NOT** accept a value for an extensible enum that is not one of the values defined for the api-version specified in the request.
+
 <a name="json-removing-enum-value-is-breaking"></a>
 :no_entry: **DO NOT** remove values from your enumeration list as this breaks customer code.
```
JeffreyRichter, line 508: Just so we're all clear: A service no longer supporting a value is a breaking change and this is allowed following the breaking change policy. In this case, all bets are off though and so the enum value CAN be removed during a formal breaking change.

azure/Guidelines.md
```diff
@@ -540,6 +546,18 @@ Below is an example of JSON for a Rectangle and Circle:
 ```
 Both Rectangle and Circle have common fields: `kind`, `fillColor`, `lineColor`, and `subscription`. A Rectangle also has `x`, `y`, `width`, and `length` while a Circle has `x`, `y`, and `radius`. The `subscription` is a nested polymorphic type. A `free` subscription has no additional fields and a `paid` subscription has `expiration` and `invoice` fields.
 
+<a name="json-polymorphism-kind-extensible"></a>
+:ballot_box_with_check: **YOU SHOULD** define the kind field of a polymorphic type to be an extensible enum.
```
JeffreyRichter: We also recommend the field be called 'kind' unless there is a great reason not to - perhaps a guideline on this (maybe in the naming section).

azure/Guidelines.md
```diff
@@ -540,6 +546,18 @@ Below is an example of JSON for a Rectangle and Circle:
...
+
+<a name="json-polymorphism-kind-immutable"></a>
+:warning: **YOU SHOULD NOT** allow an update (patch) to change the kind field of a polymorphic type.
```
JeffreyRichter: Wait, I'm not sure about this. A client might want to change a circule to a rectangle or vice versa. I think this _has to be_ allowed. So PATCH CAN change the kind field's value as long as it also specifies any properties required by the new kind value.

azure/Guidelines.md
```diff
@@ -540,6 +546,18 @@ Below is an example of JSON for a Rectangle and Circle:
...
+
+<a name="json-polymorphism-versioning"></a>
+:warning: **YOU SHOULD NOT** return properties of a polymorphic type that are not defined for the api-version specified in the request.
+
+<a name="json-polymorphism-arrays"></a>
+:warning: **YOU SHOULD NOT** specify a polymorphic type as the type of an array property.
```
JeffreyRichter: I'm not sure what this means. An array of polymorphic types is OK? Or just problematic?
If problematic, then why? 
I do remember we said something about this, but I can't remember the full context.

azure/Guidelines.md
mikekistler: Yes, this guideline can be violated by declaring a formal breaking change. But I don't think we want to soften the guidance.

We have allowed some breaking change guidance to seep into this doc. In some ways it is unavoidable given that at least parts of our guidance are motivated by avoiding breaking changes later.

azure/Guidelines.md
mikekistler: I think that is covered in our [naming guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/ConsiderationsForServiceDesign.md#common-names).

azure/Guidelines.md
mikekistler: This is a special case of the known problem that arrays are not "merge-patch-friendly" -- to add or remove an element of an array with merge-patch you must specify the entire array contents, modified accordingly. This is problematic for polymorphic types because it acts like a "replace" on individual elements, so properties of the element that were not returned due to older api-version get lost.

azure/Guidelines.md
mikekistler: Are there actual cases in Azure where this is needed?

JeffreyRichter: @johanste would know better than I. 
But, before I joined MS, I worked on a subscription web service customers could change their subscription type from Free to Invoice to Monthly, etc. as there were all polymorphic types. Free had an expiration time period property; Invoice had invoice # and contact email; Subscription had credit card info. So, while I can't recall any specific Azure example, my service would have required this ability.

azure/Guidelines.md
JeffreyRichter: Oh yea, thanks for the refresher. Maybe rephrase this bullet:
**YOU SHOULD NOT** have a property whose value is an array of polymorphic types.

mikekistler:
```suggestion
:warning: **YOU SHOULD NOT** have a property whose value is an array of polymorphic objects.
```

johanste: So what should you do if you return a sequence of values? For example, Form Recognizer returns a set of recognized fields. They are of different type.

The json-merge-patch gotcha only applies to updatable resources.

JeffreyRichter: Good point. This rule should only apply to updatable resources.

azure/Guidelines.md
johanste: The question is what you do for fields that were previously set on the resource, but that are not valid on the new kind. This is a sometimes a reason to use PUT rather than PATCH.

The scenarios where this has come up is when changing a data source type etc. (where data source types are pluggable)

JeffreyRichter:
If I'm following this conversation, then changing a kind to a new value requires the service to delete properties associated with the old kind value

-- Jeffrey
________________________________

azure/Guidelines.md
heaths: Does it hurt to call out here plain as day? We have a lot of guidelines across 2 or 3 files.

azure/Guidelines.md
```diff
@@ -540,6 +546,18 @@ Below is an example of JSON for a Rectangle and Circle:
...
+:ballot_box_with_check: **YOU SHOULD** define the kind field of a polymorphic type to be an extensible enum.
+
+<a name="json-polymorphism-kind-immutable"></a>
+:warning: **YOU SHOULD NOT** allow an update (patch) to change the kind field of a polymorphic type.
+
+<a name="json-polymorphism-versioning"></a>
+:warning: **YOU SHOULD NOT** return properties of a polymorphic type that are not defined for the api-version specified in the request.
+
+<a name="json-polymorphism-arrays"></a>
+:warning: **YOU SHOULD NOT** have a property whose value is an array of polymorphic objects.
```
heaths: Should the "considerations" doc suggest a pattern so we don't have numerous different ways of doing it across services?

azure/Guidelines.md
heaths: To clarify, I don't think we should be so explicit everywhere and duplicate information; however, a brief mention of using "kind" is recommended seems trivial in this case.

azure/Guidelines.md
```diff
@@ -503,9 +509,9 @@ This indicates to client libraries and customers that values of the enumeration
 If you can't avoid them, then follow the guideline below.
 
 <a name="json-use-discriminator-for-polymorphism"></a>
-:white_check_mark: **DO** define a `kind` field indicating the kind of the resource and include any kind-specific fields in the body.
+:white_check_mark: **DO** define a discriminator field indicating the kind of the resource and include any kind-specific fields in the body.
```
weidongxu-microsoft, line 518: nit, it is a bit hard for me to understand phrase `kind-specific fields`. Would it be better if we use `"kind"-specific fields` or a bit more detailed like `fields relate to this particular "kind" of the resource`?

stewartadam: What do we think of:

```suggestion
:white_check_mark: **DO** define a discriminator field indicating the resource kind in the body, alongside fields that are specific to that kind of resource.
```

JeffreyRichter: I'm OK with this wording as it is close enough and most resources have only 1 polymorphic type in them But, in reality, a resource could have multiple polymorphic types in it, for example:
{
   "shapeKind" : "circle",
   "circle" : { ... circle-specific properties... },
   
   "subscriptionKind": "CreditCard",
   "creditCard" : { ... creditCard-specific properties ... },
}

We just need to be aware of this should it come up. I did build a service once many years ago where this was a real issue.

mikekistler: I don't believe it is possible to describe a schema with multiple polymorphic types in OpenAPI. A schema can have at most one `discriminator`.

JeffreyRichter:
Another reason to like our polymorphic type schema proposal.

-- Jeffrey
________________________________

heaths (approved):

weidongxu-microsoft (approved):

PR #442, opened 2023-03-23
Server Sent Events guidance

### LINKED ISSUE #445: Provide guidance for SSE (Server Sent Events)
We've had a few teams ask about server sent events and the proper patterns. We need to add guidance for these teams.

johanste:
Adding simple guidance on when and how service sent events may be used (and what constraints we put on their usage)

### DIFF
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -1279,6 +1279,31 @@ In addition to distributed tracing, Azure also uses a set of common correlation 
 |x-ms-client-request-id       |Both      |Optional. Caller-specified value identifying the request, in the form of a GUID with no decoration such as curly braces (e.g. `x-ms-client-request-id: 9C4D50EE-2D56-4CD3-8152-34347DC9F2B0`). If the caller provides this header the service **must** include this in their log entries to facilitate correlation of log entries for a single request. Because this header can be client-generated, it should not be assumed to be unique by the service implementation.
 |x-ms-request-id              |Response  |Required. Service generated correlation id identifying the request, in the form of a GUID with no decoration such as curly braces. In contrast to the the `x-ms-client-request-id`, the service **must** ensure that this value is globally unique. Services should log this value with their traces to facilitate correlation of log entries for a single request.
 
+<a name="serversentevents"></a>
+### Server Sent Events (SSE)
+For services that generate partial results over time, it is often beneficial to return the data as soon as it is available to reduce latency. A specific example of this is the generate method for Open AI. One option for this is [Server Sent Events (SSE)](https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events). SSE may be used under the following conditions:
+
+<a name="serversentevents-partial-results"></a>
+:heavy_check_mark: **YOU MAY** use SSE to incrementally return partial results.
+
+<a name="serversentevents-derived-from-request"></a>
+:white_check_mark: **DO** Ensure that the response stream is only derived/computed from information in the initial request.
+
+<a name="serversentevents-no-reconnect"></a>
+:no_entry: **DO NOT** require clients to reconnect/re-establish the stream.
+
+<a name="serversentevents-short-lived-streams"></a>
+:white_check_mark: **DO** only use SSE when streams complete (end the stream) within five (5) minutes P99 processing time.
+
+<a name="serversentevents-structured-data-only"></a>
+:white_check_mark: **DO** only use SSE for sequences of structured data/events. For binary/unstructured data, using chunked encoding is a better alternative.
+
+<a name="serversentevents-structured-single-line"></a>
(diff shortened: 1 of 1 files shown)


azure/Guidelines.md
```diff
@@ -1279,6 +1279,31 @@ In addition to distributed tracing, Azure also uses a set of common correlation
...
 |x-ms-request-id              |Response  |Required. Service generated correlation id identifying the request, in the form of a GUID with no decoration such as curly braces. In contrast to the the `x-ms-client-request-id`, the service **must** ensure that this value is globally unique. Services should log this value with their traces to facilitate correlation of log entries for a single request.
 
+<a name="serversentevents"></a>
+### Server Sent Events (SSE)
+For services that generate partial results over time, it is often beneficial to return the data as soon as it is available to reduce latency. A specific example of this is the generate method for Open AI. One option for this is [Server Sent Events (SSE)](https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events). SSE may be used under the following conditions:
```
vicancy, line 1284: * Do we want to mention some limitations of SSE for the service designer to be aware of? For example, the connection limit per browser, or some proxies do not support SSE. 
* Usually there might be some keep-alive mechanism for the SSE connection to detect if the underlying connection is dropped. For example when the network cable is unplugged, the TCP connection is still considered as connected for minutes.
*  Is it better for the service to always provide a fallback plan (e.g. longpolling) so that the client can switch to it when SSE does not work?

johanste: I assume that implementors are already familiar with SSE and its inherent limitations, so I wasn't intending to repeat that here. If we want to expand on the guidance, we can do so. 

Regarding keep-alive, the intent of limiting usage to relatively short sessions is to minimize the need. The standard keep-alive (e.g. occasionally sending comments) is something that I also assumed that service implementors are aware of. But, again, we can add this.

I would not recommend always having a fallback plan like long polling. Instead I'd recommend that there is a non-streaming way to get the same data once it is all available.

mikekistler, line 1284: Is there a description of how SSE works at the HTTP level?  Does it use "chunked responses"?

johanste: Chunked responses is a Transport-Encoding. SSE is a content/MIME type. See [this](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream) for more info.

azure/Guidelines.md
```diff
@@ -1279,6 +1279,31 @@ In addition to distributed tracing, Azure also uses a set of common correlation
...
+:no_entry: **DO NOT** require clients to reconnect/re-establish the stream.
+
+<a name="serversentevents-short-lived-streams"></a>
+:white_check_mark: **DO** only use SSE when streams complete (end the stream) within five (5) minutes P99 processing time.
+
+<a name="serversentevents-structured-data-only"></a>
+:white_check_mark: **DO** only use SSE for sequences of structured data/events. For binary/unstructured data, using chunked encoding is a better alternative.
+
+<a name="serversentevents-structured-single-line"></a>
+:white_check_mark: **DO** send each event as a single JSON line.
```
JeffreyRichter, line 1302: According to https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream, the event data is NOT JSON.

azure/Guidelines.md
```diff
@@ -1279,6 +1279,31 @@ In addition to distributed tracing, Azure also uses a set of common correlation
...
+
+<a name="serversentevents-end-of-stream-message"></a>
+:white_check_mark: **DO** provide a clear, documented way for clients to know that a given message is the last message in the stream.
```
JeffreyRichter, line 1305: I think we need to be more prescriptive here. Doesn't the "spec" say how to know when the event stream is complete?
If not, we need to define specifically what this must be.

azure/Guidelines.md
JeffreyRichter, line 1284: I think better guidance is required to explain exactly when this technology is the right solution. "For services that generate partial results over time" is not crisp enough. Could this be used for an LRO result or incremental results? What happens if the connection fails while getting events? Is it OK to repeat or skip events when re-connecting? Is it even OK to reconnect? How does a client know when the event stream has competed? SSE seems to not have a great solution for handing this - it seems to be more for a continuously ongoing stream of events.

JeffreyRichter (commented):
I recently spent a lot more time learning/understanding SSE and I believe that we should NOT have any guidance related to this in our guidelines. ChatGPT does not use SSE as it is spec'd at all. They deviated from it very substantially. If an Azure service did what they did, we would not allow it. I understand that Azure is wrapping OpenAI's ChatGPT APIs and we have to live with this but this should not be normal/promoted for Azure services.

mikekistler:
Closing until we had a definite adopter.

PR #449, opened 2023-04-19
Fix Dictionary guidance to use PATCH to add/remove/update properties.

mikepizzo:
Fix Dictionary guidance to use PATCH rather than POST/DELETE to add properties to the dictionary.


syprieur (commented):
Unrelated to changes in this PR:

Should the guidance be updated to clarify that an open complex type `microsoft.graph.Dictionary` is used as base type in Production today, not `Org.OData.Core.V1.Dictionary`?

graph/patterns/dictionary.md
```diff
@@ -10,9 +10,9 @@ The API design requires a resource to include an unknown quantity of data elemen
 
 ## Solution
 
-API designers use a JSON object to represent a dictionary in an `application/json` response payload. When describing the model in CSDL, a new complex type can be created that derives from `Org.OData.Core.V1.Dictionary` and then uses the `Org.OData.Validation.V1.OpenPropertyTypeConstraint` to constrain the type that can be used for the values in the dictionary.
+API designers use a JSON object to represent a dictionary in an `application/json` response payload. When describing the model in CSDL, a new complex type can be created that derives from `Org.OData.Core.V1.Dictionary` and then uses the `Org.OData.Validation.V1.OpenPropertyTypeConstraint` to constrain the type that can be used for the values in the dictionary as appropriate.
 
-Dictionary entries can be added via `POST`, updated via `PATCH`, and removed by setting the entry value to `null`. Multiple entries can be updated at the same time by using `PATCH` on the dictionary property.
+Dictionary values can be added, removed, or modified via `PATCH` to the dictionary property. Values are removed by setting the property to `null`.
```
OlgaPodo: @mikepizzo In this context, an entry is a (key, value) pair. IMO If this sentence is related only to values, then we need to explain how to add a new entry with a new key. #Resolved

graph/patterns/dictionary.md
```diff
@@ -36,8 +36,9 @@ Because dictionary entries are removed by setting the value to `null`, dictionar
 Open questions:
 
 - Can/should PUT be supported on the dictionary property and/or the entry value?
-- What does OData say about being able to POST to a structured property? Will OData Web API allow that?
 - Must an implementer support PATCH at both the dictionary level and the entry level?
+- Should we also allow DELETE to a property to be equivalent to setting to null?
+- Why do we not allow mixed primitives or mixed primitive/complex typed values? what about collections?
```
dkershaw10: Do we have a constraint here for SDKs or AGS?  Otherwise, why wouldn't we allow this?
If we do, should we have an updated example to show this usage?

I also wonder whether we should have 2 examples - 1 simple example that replaces the standard KeyValue(String primitive) pair which is the canonical example, and something more complex with mixed primitives/complex types/collections. #Resolved

graph/patterns/dictionary.md
mikepizzo: I see now the distinction you're making between an entry and a value. I'm fine changing to entry here...

graph/patterns/dictionary.md
```diff
@@ -175,7 +176,10 @@ PATCH https://graph.microsoft.com/v1.0/users/10/roles/author
...
 
 ```HTTP
-DELETE https://graph.microsoft.com/v1.0/users/10/roles/author
+PATCH https://graph.microsoft.com/v1.0/users/10/roles/author
+{
+  "domain": null
+}
 ```
 
 ### CDSL example
```
OlgaPodo: Working with a developer today, I realized that this example is missing how `assignedRoleGroupDictionary` is modeled in the context of the `role` entity type.
 #Resolved

mikepizzo: Yes.

Originally in this PR I tried to make the minimal change to the existing guidance to use PATCH rather than POST.  However, comments have demonstrated that there are many other issues with the guidance, so I am in the process of rewriting it to address some of those more systemic issues.

graph/patterns/dictionary.md
```diff
@@ -31,41 +31,122 @@ Before using a dictionary type in your API definition, make sure that your scena
...
+<EntityType Name="principal">
+  ...
+  <Property Name="assignedRoles" Type="self.assignedRoleGroupDictionary">
+</EntityType>
+
+<ComplexType Name="roleSettings">
+  <Property Name ="domain" Type="Edm.String" Nullable="false" />
+</ComplexType>
+
+<ComplexType Name="assignedRoleGroupDictionary" BaseType="graph.Dictionary">
```
OlgaPodo, line 121: i believe it should be an open type #Resolved

dkershaw10: Dictionary is an open type, so I think this is OK.

graph/patterns/dictionary.md
```diff
@@ -2,24 +2,24 @@
...
 
 ## Problem
 
-The API design requires a resource to include an unknown quantity of data elements of the same type that must be named by using values provided by the API consumer.
+The API design requires a resource to include an unknown quantity of data values whose keys are defined by the API consumer.
 
 ## Solution
 
-API designers use a JSON object to represent a dictionary in an `application/json` response payload. When describing the model in CSDL, a new complex type can be created that derives from `Org.OData.Core.V1.Dictionary` and then uses the `Org.OData.Validation.V1.OpenPropertyTypeConstraint` to constrain the type that can be used for the values in the dictionary.
+API designers use a JSON object to represent a dictionary in an `application/json` response payload. When describing the model in CSDL, a new complex type can be created that derives from `graph.Dictionary` and optionally uses the `Org.OData.Validation.V1.OpenPropertyTypeConstraint` to constrain the type that can be used for the values in the dictionary as appropriate.
```
dkershaw10, line 13: graph.Dictionary doesn't exist in a workload's CSDL.  So what are they supposed to use here?

I think we still have them create the Core Dictionary type, and suppress the conflict.  See https://msazure.visualstudio.com/One/_git/AD-AggregatorService-Workloads?path=%2FWorkloads%2FMicrosoft.DataClassificationService%2Foverride%2Fschema-Prod-beta.csdl&_a=contents&version=GBmaster for example.

syprieur: > So what are they supposed to use here?

Just define it, like any other shared complex type.

> I think we still have them create the Core Dictionary type

What the CSDL you shared does is simply define the shared complex type `graph.Dictionary` under their own private workload namespace named `Org.OData.Core.V1` (so that it maps to the actual OData type on their side).

If we want to support the "true" `Org.OData.Core.V1.Dictionary` out-of-the-box, that'll require an AGS change (as this type doesn't exist to AGS and is unknown today).

PR #460, opened 2023-06-21
Add documentation explaining the different cases for required properties, nullable properties, and properties with default values

corranrogue9:


graph/GuidelinesGraph.md
```diff
@@ -248,6 +248,11 @@ Following are a few pros and cons to decide which pattern to use:
 > **Note:**
 > As can be seen in a few of the pros and cons, one of the important aspects discussed here is that the API design goes beyond the syntactical aspects of the API. Therefore, it is important to plan ahead how the API evolves, lay the foundation, and allow users to form a good understanding of the semantics of the API. **Changing the semantics is always a breaking change.** The different modeling patterns differ in how they express syntax and semantics and how they allow the API to evolve without breaking compatibility. For more information, see [API contract and non-backward compatible changes](#api-contract-and-non-backward-compatible-changes) later in this article.
 
+#### Nullable properties
+
+The facets and flat bag approaches often require nullable properties, so it is important make sure to still use non-nullable properties where appropriate, and since inheretence can often remove the use of nullable properties completely, knowing when nullable properties are necessary is also important.
+Please view the guidelines [here](./nullable.md) for more details. 
+
```
OlgaPodo, line 249:
```suggestion
The facet and flat bag approaches often require nullable properties, so it is important to still use non-nullable properties where appropriate. Since inheritance can often remove the use of nullable properties completely, it is also important to know when nullable properties are necessary. See [Nullable properties](./nullable.md) for more details.
```

graph/nullable.md
```diff
@@ -0,0 +1,312 @@
...
+  "appId": "00000000-0000-0000-0000-000000000001",
+  "displayName": "a non-generated display name",
+  "foo": "testval",
+  "bar": "differentvalue",
+  ...
+}
+```
+Notes:
+1. `displayName` can be set to any value other than `null`
+2. The response body here is provided for clarity; normally on Microsoft Graph, a `204 No Content` will be the response for a `PATCH` request
```
OlgaPodo: IMO instead of providing non-standard response it would be better to add GET request

graph/nullable.md
```diff
@@ -0,0 +1,312 @@
...
+```HTTP
+PATCH /servicePrincipals/00000000-0000-0000-0000-000000000001
+{
+  "displayName": null
+}
+
+400 Bad Request
+{
+  "error": {
+    "code": "BadRequest",
```
OlgaPodo: NIT: "BadRequest" should be in camelCase like "badRequest"

corranrogue9:
```suggestion
    "code": "badRequest",
```

graph/nullable.md
corranrogue9: Hmm...

The [OData standard](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_UpdateanEntity) says:
> Upon successful completion the service responds with either [200 OK](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_ResponseCode200OK) and a representation of the updated entity, or [204 No Content](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_ResponseCode204NoContent).

I don't see any guidance that actually says that we prefer `204` for `PATCH` on graph. I'm inclined to update the note to say:
> The response body here is provided for clarity, and is not part of the guidance itself. The [OData standard](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_UpdateanEntity) states that the workload can decide the behavior. 

Does that sound alright with you?

OlgaPodo: Sounds good! I would just replace 'OData standard' with 'Graph standard'

OlgaPodo (approved):

graph/nullable.md
```diff
@@ -0,0 +1,312 @@
...
+  "appId": "00000000-0000-0000-0000-000000000001",
+  "displayName": "a non-generated display name",
+  "foo": "testval",
+  "bar": "differentvalue",
+  ...
+}
+```
+Notes:
+1. `displayName` can be set to any value other than `null`
+2. The response body here is provided for clarity, and is not part of the guidance itself. The [Graph standard](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_UpdateanEntity) states that the workload can decide the behavior.
```
OlgaPodo: Sorry for being a stickler.  I meant to put a reference to the Graph Guidelines with an idea that the Graph guidelines can be more prescriptive. However, if we use OData as a reference, I think we should display the text as "OData v4.01".

corranrogue9: > Sorry for being a stickler

I definitely want it to be right, and to set good precedent, I appreciate the feedback

> I meant to put a reference to the Graph Guidelines with an idea that the Graph guidelines can be more prescriptive

Do you mean to create a new `.md` file with just some "TODO" text? Or would you like me to write out some guidance? I'm not sure that there is consistent guidance regarding response payloads for created entities, because I would think it's probably going to be pretty scenario dependent

OlgaPodo: Let's use an incremental approach. please change [Graph standard] back to [OData v4.01] and I'll add a backlog item for 'Graph specific HTTP Request/Response guidelines' to the Guidelines project and API Council agenda so that we can close the gap later.

graph/GuidelinesGraph.md
```diff
@@ -248,6 +248,11 @@ Following are a few pros and cons to decide which pattern to use:
 > **Note:**
 > As can be seen in a few of the pros and cons, one of the important aspects discussed here is that the API design goes beyond the syntactical aspects of the API. Therefore, it is important to plan ahead how the API evolves, lay the foundation, and allow users to form a good understanding of the semantics of the API. **Changing the semantics is always a breaking change.** The different modeling patterns differ in how they express syntax and semantics and how they allow the API to evolve without breaking compatibility. For more information, see [API contract and non-backward compatible changes](#api-contract-and-non-backward-compatible-changes) later in this article.
 
+#### Nullable properties
+
+The facets and flat bag approaches often require nullable properties, so it is important make sure to still use non-nullable properties where appropriate, and since inheretence can often remove the use of nullable properties completely, knowing when nullable properties are necessary is also important.
```
```suggestion
```
Looks like we cleaned up but didn't remove the original?

graph/nullable.md
```diff
@@ -0,0 +1,312 @@
+# Nullable Properties
```
dkershaw10, line 1: @OlgaPodo is this going to be standard practice now. to have these guidance expansion topics vs pattern topics?

graph/nullable.md
```diff
@@ -0,0 +1,312 @@
...
+
+A nullable property means *only* that the property may have `null` as a value; the "nullability" of a property does not say anything about how a value is set into a property.
+For example, a non-nullable property is *not* required to create a new instance of an entity.
```
dkershaw10, line 4: We need to fix the linting.  Pretty sure there's a rule that warns of a breaking change when adding a non-nullable property.

graph/nullable.md
```diff
@@ -0,0 +1,312 @@
...
+
+```xml
+<EntitySet Name="servicePrincipals" Type="self.servicePrincipal" />
+...
+<EntityType Name="servicePrincipal">
+  <Key>
+    <PropertyRef Name="id" />
+  </Key>
+  <Property Name="id" Type="Edm.String" Nullable="false" />
+
```
dkershaw10: Any reason for this extra line?

graph/nullable.md
dkershaw10: I thought that most PATCH requests respond with 204 No content, but using the Prefer representation HTTP request header, the service should respond with a 200 OK and the full response.  Can we just add the Prefer header and the preference applied response header?

graph/nullable.md
```diff
@@ -0,0 +1,312 @@
...
+  "foo": "testval",
+  "bar": "differentvalue",
+  ...
+}
+```
+Notes:
+1. `displayName` can be set to any value other than `null`
+2. The response body here is provided for clarity, and is not part of the guidance itself. The [Graph standard](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_UpdateanEntity) states that the workload can decide the behavior.
+
+### {5} Update the foo property of a service principal to null
```
dkershaw10, line 105: Can we combine some of these examples to make updates to multiple properties, for the successful cases?  That would shorten the document.

I also think we have a **lot** of examples here. Are we repeating any concepts that could be removed?  I'm all for being thorough, but I'm pretty sure we can shorten the doc, and trust that our readers are smart enough to figure things out based on fewer examples.

dkershaw10 (changes requested):
Added a few comments for you @corranrogue9

graph/nullable.md
corranrogue9: I did this as a bit of a "workaround" since explaining the nuances of nullability and required and such aren't really a "pattern". I don't think this should be standard practice per se, but I do think it's important to not call "guidelines" patterns. I'm open to restructuring wherever makes the most sense.

graph/nullable.md
corranrogue9: I created [this](https://dev.azure.com/IdentityDivision/Engineering/_workitems/edit/2634761) work item for that. Do you think that work item needs to be completed before the guidance is published?

graph/nullable.md
corranrogue9: I changed it back to "OData" for now. Dan, I'm happy to add a prefer header if you think that's best. What I'm trying to convey with this note is that the response body, a prefer header, etc are all just to allow the sample to fully exemplify the behavior; I don't want anyone to be confused that they must have a response body or require a header or anything like that. Let me know, I'll add the headers if you think that's the best way to show that to readers.

graph/nullable.md
corranrogue9: There are 3 dimensions to the properties:
1. Nullable
2. Default value
3. Required for creation

I did at least trim this down from the possible 8 to only 4. 

There are 3 cases per property for each request:
1. Not provided
2. Provided as null
3. Provided as not null

So, I wrote 2 samples for the creation, and then I did cover 100% of the cases per property. 

I see your point about folks filling in the blanks for themselves, but I'm not really sure who would read this particular document top to bottom in the first place. To me, this doc seems more for the exact purpose of covering 100% of cases so that if there's any confusion, we can immediately clear it up.

I'm pushing back on this feedback in particular. If there's something you would like to use this document for where brevity would be better, I can add some notes at the top or something like that to suit those needs.

corranrogue9:
@dkershaw10 I've replied to your feedback. I'm happy to discuss more offline if that's needed. Please take another look when you get a chance.

graph/nullable.md
dkershaw10: Nope.  If anything it's the reverse, because when a rule fails, it should point to guidance.

graph/nullable.md
```diff
@@ -0,0 +1,311 @@
...
+## CSDL
+
+```xml
+<EntitySet Name="servicePrincipals" Type="self.servicePrincipal" />
+...
+<EntityType Name="servicePrincipal">
+  <Key>
+    <PropertyRef Name="id" />
+  </Key>
+  <Property Name="id" Type="Edm.String" Nullable="false" />
```
dkershaw10, line 18: You can resolve when you see this.
I just had an API owner try to use a defaultValue for a primary key...

graph/nullable.md
dkershaw10: I'm not going to block on this.  This topic is very useful and closes a major gap in API producers understanding in this space, so I'm in favour of publishing and potentially refining and making more concise later.

dkershaw10 (approved):
LGTM

PR #463, opened 2023-06-27
Update wording of top level error codes to be more prescriptive.

gareth-a-jones:


graph/GuidelinesGraph.md
```diff
@@ -285,7 +285,7 @@ Microsoft REST API Guidelines provide guidelines that Microsoft Graph APIs shoul
 }
 ```
 
-The top-level error code must be aligned with HTTP response status codes according to [rfc7231 (ietf.org)](https://datatracker.ietf.org/doc/html/rfc7231#section-6). The following examples demonstrate error modeling for common use cases:
+The top-level error code must match the HTTP response status codes, converted to camelCase, according to [rfc7231 (ietf.org)](https://datatracker.ietf.org/doc/html/rfc7231#section-6). The following examples demonstrate error modeling for common use cases:
```
darrelmiller:
```suggestion
The top-level error code must match the HTTP response status code description, converted to camelCase, as listed in the [Status Code Registry (iana.org)](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml). The following examples demonstrate error modeling for common use cases:
```

PR #482, opened 2023-07-21
Add guidance for the use of collections of entity types

corranrogue9:
I'm adding guidance to suggest that collections should be entity types instead of complex types because they can be modified piecewise, while collections of complex types must always be replaced, which can result in data loss and/or race conditions.

Please ignore the commit comments, I will perform a squash merge to complete the PR.

### DIFF
--- a/graph/GuidelinesGraph.md
+++ b/graph/GuidelinesGraph.md
@@ -182,7 +182,9 @@ Another way to avoid this is to use JSON batch as described in the [Microsoft Gr
 
 You can model structured resources for your APIs by using the OData entity type or complex type. The main difference between these types is that an entity type declares a key property to uniquely identify its objects, and a complex type does not. In Microsoft Graph, this key property is called `id` for server-created key values. If there is a natural name for the key property, then the workload can use that.
 
-Because objects of complex types in Microsoft Graph don’t have unique identifiers, they are not directly addressable via URIs. Therefore, you must use entity types to model addressable resources such as individually addressable items within a collection. For more information, see the [Microsoft REST API Guidelines collection URL patterns](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#93-collection-url-patterns). Complex types are better suited to represent composite properties of API entities.
+Because objects of complex types in Microsoft Graph don’t have unique identifiers, they are not directly addressable via URIs. Therefore, you must use entity types to model addressable resources such as individually addressable items within a collection. For more information, see the [Microsoft REST API Guidelines collection URL patterns](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#93-collection-url-patterns).
+This is also why [collection properties](./collection-properties.md) should almost always be modeled using an entity type rather than a complex type. 
+Complex types are better suited to represent composite properties of API entities.
 
 ```xml
  <EntityType Name="author">
--- a/graph/collection-properties.md
+++ b/graph/collection-properties.md
@@ -97,5 +97,19 @@ GET  /interestingData/bars
 
 ## Retrieving individual elements from a collection
 
-TODOD
-https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_RequestingIndividualEntities
+The [OData standard](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_RequestingIndividualEntities) also specifies that clients can retrieve individual elements of a collection of entity types:
+> To retrieve an individual entity, the client makes a GET request to a URL that identifies the entity, e.g. its read URL.
+
+However, there is no way to do this for complex types because there is not way to identity a particular instance of a complex type. 
+
+### Entity Type
+
+```HTTP
+GET  /interestingData/bars/firstBarId
+
+200 OK
+{
+  "id": "firstBarId",
+  "differentProperty": 42
+}
+```
--- a/graph/collection-properties.md
+++ b/graph/collection-properties.md
@@ -1,0 +1,2 @@
+OData services treat collections of complex types differently than they treat collections of entity types. 
+This is due to the nature of entity types being "individually addressable" (they have some key which uniquely identifies them within the collection) while complex types are not individually addressable. 
--- a/graph/collection-properties.md
+++ b/graph/collection-properties.md
@@ -1,6 +1,6 @@
 OData services treat collections of complex types differently than they treat collections of entity types. 
 This is due to the nature of entity types being "individually addressable" (they have some key which uniquely identifies them within the collection) while complex types are not individually addressable. 
-The result of these differences is that it is almost always preferable to use entity types for collections rather than complex types; please see the exceptions [here](TODO).
+The result of these differences is that it is almost always preferable to use entity types for collections rather than complex types; please see the exceptions [here](#exceptions).
 Let's use the following CSDL as an example:
 
 ```xml
--- a/graph/collection-properties.md
+++ b/graph/collection-properties.md
@@ -136,7 +136,7 @@ There is no way to do this for complex types because there is no way to identity
 ### Entity Type
 
 ```HTTP
-DELETE  /interestingData/bars/firstBarId
+DELETE  /interestingData/bars/thirdBarId
 
 204 No Content
 ```
--- a/graph/collection-properties.md
+++ b/graph/collection-properties.md
@@ -1,6 +1,6 @@
 OData services treat collections of complex types differently than they treat collections of entity types. 
 This is due to the nature of entity types being "individually addressable" (they have some key which uniquely identifies them within the collection) while complex types are not individually addressable. 
-The result of this difference is that it is almost always preferable to use entity types for collections rather than complex types; please see the exceptions [here](#exceptions).
+As a result, it is almost always preferable to use entity types for collections rather than complex types. See [exceptions](#exceptions) for acceptable usage of complex types in collections.
 Let's use the following CSDL as an example:
 
 ```xml
--- a/graph/collection-properties.md
+++ b/graph/collection-properties.md
@@ -1,4 +1,4 @@
-OData services treat collections of complex types differently than they treat collections of entity types. 
+Microsoft Graph treats collections of complex types differently than collections of entity types. 
 This is due to the nature of entity types being "individually addressable" (they have some key which uniquely identifies them within the collection) while complex types are not individually addressable. 
 As a result, it is almost always preferable to use entity types for collections rather than complex types. See [exceptions](#exceptions) for acceptable usage of complex types in collections.
 Let's use the following CSDL as an example:
--- a/graph/collection-properties.md
+++ b/graph/collection-properties.md
@@ -336,7 +336,26 @@ GET /interestingData/bars
 
 #### Change the elements in a collection
 
-TODO
+```HTTP
+PATCH /interestingData/bars
+{
+  "@context": "#$delta",
+  "value": [
+    {
+      "id": "fifthBarId",
+      "differentProperty": 1024
+    },
+    {
+      "id": "sixthBarId",
+      "differentProperty": 9801
+    },
+    {
+      "id": "fourthBarId",
+      "@removed": {}
+    }
+  ]
+}
+```
 
 #### Check the new contents of the collection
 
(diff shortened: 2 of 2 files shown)


graph/collection-properties.md
```diff
@@ -0,0 +1,402 @@
...
+      "differentProperty": -6914
+    }
+  ]
+}
+```
+
+## Adding individual elements to a collection
+
+For both `foos` and `bars`, the OData standard specifies that elements can be added to the collection using a `POST` request
+1. [Complex Types](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_UpdateaCollectionProperty):
```
mikepizzo, line 72: The protocol supports this, but it's a bit asymmetric to support adding but not removing, so I'm not sure it's a pattern we want to encourage on ms graph. at least we should all out that it's an uncommon pattern due to this asymmetry.

graph/collection-properties.md
```diff
@@ -0,0 +1,402 @@
...
+### Complex Type
+
+```HTTP
+GET  /interestingData/foos
+
+200 OK
+{
+  "value": [
+    {
+      "someProperty": "an original value"
```
mikepizzo, line 43: is there a reason that your complex collection example starts with one value, and your entity collection sample starts with 2? it makes the sample payloads look more different than they are.

graph/collection-properties.md
```diff
@@ -0,0 +1,402 @@
...
+      "differentProperty": 15
+    },
+    {
+      "id": "secondBarId",
+      "differentProperty": -6914
+    }
+  ]
+}
+```
+
```
mikepizzo, line 243: also, you can have other objects reference an entity within a collection, which you cannot do with a complex type:

```xml
<EntityType Name="baz">
      <Property Name="Description" Type="Edm.String"/>
      <NaviationProperty Name="bar" Type="self.bar"/>
</EntityType>
```

```HTTP
GET interestingData/baz
200 OK
{
    "Description": "my favorite bar",
    "bar": 
    {
        "id": "firstBarId",
        "differentProperty": 15
     }
}
```

graph/collection-properties.md
```diff
@@ -0,0 +1,402 @@
...
+## Updating a collection
+
+The [OData standard](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_UpdateanEntity) specifies that clients can replace all elements of a collection of complex types or all elements of a collection of entity types using a `PATCH` request:
+> Collection properties...provided in the payload corresponding to updatable properties MUST replace the value of the corresponding property in the entity or complex type.
+
+The standard also specifies that collections of entities can be updated in a relative fashion using the [delta syntax](https://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#_Toc38457777) in a `PATCH` request:
```
mikepizzo, line 249: maybe move this to after the replace request and just before the delta example.  Also call out (importantly) that this is not possible with complex types.

mikepizzo, line 249: I'm not sure "relative fashion" will resonate as strongly with what the user is trying to do here. maybe something like "The standard allows updating a collection of entities, including adding, removing, or modifying individual existing entities in a single PATCH request. This is not possible with complex types because there is no way to identity them individually within the collection."

graph/collection-properties.md
```diff
@@ -0,0 +1,402 @@
...
+      "id": "sixthBarId",
+      "differentProperty": 9801
+    }
+  ]
+}
+```
+
+## Exceptions
+
+It might not make sense for a collection property to be a collection of entities in cases where the collection must be viewed holistically; that is, cases where one element in the collection impacts or is coupled with another element in the collection.
```
mikepizzo, line 385: OData does define the concept of [ordered collections](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_PositionalInserts), with the ability to insert within the ordered collection according to ordinal. We could choose to adopt that pattern in graph, which would help facilitate these types of scenarios on graph and not fall back to a collection of complex types.

graph/collection-properties.md
```diff
@@ -0,0 +1,402 @@
...
+
+## Conclusion
+
+Collections of entity types have several behaviors that are not available for collections of complex types:
+1. Individual elements can be retrieved
+2. Individual elements can be removed
+3. Individual elements can be updated
+4. Several elements within the collection may be added, removed, or updated in a single request
+
+Further, to remove or update elements in a collection of complex types, the entire collection must be replaced with a `PATCH` request.
```
mikepizzo, line 400: suggest: "within"

graph/collection-properties.md
```diff
@@ -0,0 +1,402 @@
...
+## Conclusion
+
+Collections of entity types have several behaviors that are not available for collections of complex types:
+1. Individual elements can be retrieved
+2. Individual elements can be removed
+3. Individual elements can be updated
+4. Several elements within the collection may be added, removed, or updated in a single request
+
+Further, to remove or update elements in a collection of complex types, the entire collection must be replaced with a `PATCH` request. 
+This has the potential to result in data loss for clients who accidentally don't include all of the current elements of the collection, or encounter a race condition where two clients are attempting to `PATCH` the same collection.
```
mikepizzo, line 401: here I would just use "update"; the key is that they are updating the collection, not what verb they are using.

graph/collection-properties.md
```diff
@@ -0,0 +1,402 @@
...
+In such a case, the workload should simply not support a delta `PATCH` request on this collection.
+
+## Conclusion
+
+Collections of entity types have several behaviors that are not available for collections of complex types:
+1. Individual elements can be retrieved
+2. Individual elements can be removed
+3. Individual elements can be updated
+4. Several elements within the collection may be added, removed, or updated in a single request
+
```
mikepizzo, line 399: and individual entities within a collection can be referenced, i.e., by other entities.

graph/GuidelinesGraph.md
```diff
@@ -182,7 +182,9 @@ Another way to avoid this is to use JSON batch as described in the [Microsoft Gr
...
 
-Because objects of complex types in Microsoft Graph don’t have unique identifiers, they are not directly addressable via URIs. Therefore, you must use entity types to model addressable resources such as individually addressable items within a collection. For more information, see the [Microsoft REST API Guidelines collection URL patterns](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#93-collection-url-patterns). Complex types are better suited to represent composite properties of API entities.
+Because objects of complex types in Microsoft Graph don’t have unique identifiers, they are not directly addressable via URIs. Therefore, you must use entity types to model addressable resources such as individually addressable items within a collection. For more information, see the [Microsoft REST API Guidelines collection URL patterns](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#93-collection-url-patterns).
```
mikepizzo, line 185: small nit, and i realize this is previous wording, but "must" immediately makes me think that you're making me do something that i wouldn't otherwise want to do.  i realize "you get to use entity types..." also doesn't quite hit the mark.  maybe something more neutral like "Entity types enable you to model addressable resources..."

graph/GuidelinesGraph.md
```diff
@@ -182,7 +182,9 @@ Another way to avoid this is to use JSON batch as described in the [Microsoft Gr
...
+This is also why [collection properties](./collection-properties.md) should almost always be modeled using an entity type rather than a complex type.
```
mikepizzo, line 186: Suggestion: "The inability to directly address an individual complex type within a collection means that..."

mikepizzo (commented):
It will surprise no-one to hear that I'm glad to get these guidelines published... ;-)

Looks good in general; a few comments below...

graph/GuidelinesGraph.md
OlgaPodo: This year we committed to merge MS REST guidelines with Graph guidelines, since Azure had already diverged completely from the baseline guidelines. Thus, I think we should delete the reference to the baseline guidelines and create a comprehensive pattern for when and how to use collections. We should also move all the details we need to this pattern and leave the standard OData query parameters to the OData specification, with a reference to it.

graph/collection-properties.md
```diff
@@ -0,0 +1,402 @@
...
+## Retrieving the elements in a collection
+
+The [OData](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#_Toc31358935) [standard](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#_Toc31358947) specifies `foos` and `bars` both can be retrieved using a `GET` request:
+> OData services support requests for data via HTTP GET requests.
+> 
+> ...
+> 
+> OData services support querying collections of entities, complex type instances, and primitive values.
+
+### Complex Type
```
OlgaPodo, line 34: As i understood your suggestion was to prefer collection of entity types when there is a need to update individual elements therefore i would remove references to complex types and show difference only when it's absolutely essential so that the pattern would be relatively short. IMO if you show differences for GET then the POST doesn't add much value. While PATCH is important.

graph/collection-properties.md
```diff
@@ -0,0 +1,402 @@
+OData services treat collections of complex types differently than they treat collections of entity types.
```
dkershaw10: Isn't this missing a title?

dkershaw10: Not sure we should start a topic with "Odata services treat"

```suggestion
Microsoft Graph treats collections of complex types differently than collections of entity types. 
```

graph/collection-properties.md
```diff
@@ -0,0 +1,402 @@
+OData services treat collections of complex types differently than they treat collections of entity types. 
+This is due to the nature of entity types being "individually addressable" (they have some key which uniquely identifies them within the collection) while complex types are not individually addressable. 
+The result of this difference is that it is almost always preferable to use entity types for collections rather than complex types; please see the exceptions [here](#exceptions).
```
dkershaw10: I think the style guide is normally about brevity rather than politeness...

```suggestion
As a result, it is almost always preferable to use entity types for collections rather than complex types. See [exceptions](#exceptions) for acceptable usage of complex types in collections.
```

graph/collection-properties.md
```diff
@@ -0,0 +1,402 @@
+OData services treat collections of complex types differently than they treat collections of entity types. 
+This is due to the nature of entity types being "individually addressable" (they have some key which uniquely identifies them within the collection) while complex types are not individually addressable.
```
dkershaw10, line 2: Is there anywhere in this topic were we want to say something about collections of primitive types?

graph/collection-properties.md
```diff
@@ -0,0 +1,402 @@
+OData services treat collections of complex types differently than they treat collections of entity types. 
+This is due to the nature of entity types being "individually addressable" (they have some key which uniquely identifies them within the collection) while complex types are not individually addressable. 
+The result of this difference is that it is almost always preferable to use entity types for collections rather than complex types; please see the exceptions [here](#exceptions).
+Let's use the following CSDL as an example:
+
+```xml
+<EntityType Name="interestingData">
```
dkershaw10, line 7: Can we have something a little more realistic?
I think we can do better than using foo and bar everywhere...

dkershaw10, line 7: Is this intended to be a singleton?  We might want to mention that, as earlier we point out that entity types have keys :)

graph/collection-properties.md
dkershaw10: Yes - this came up recently in a Defender API review - they wanted to reference an item in another collection, but that collection used complex types :(

graph/collection-properties.md
```diff
@@ -0,0 +1,402 @@
...
+The most common example of this would be a list of "tasks" that should be executed in order.
+Inserting a single task into a collection only makes sense because the client knows the preceeding and proceeding tasks.
+In cases like this, it might make sense for a collection to use a complex type instead of an entity type.
+However, even in this case, it might make sense to use a collection of entity types anyway so that clients can individually address the tasks in the list, having URLs to individual tasks that can be referred to. 
+In such a case, the workload should simply not support a delta `PATCH` request on this collection.
+
+## Conclusion
+
+Collections of entity types have several behaviors that are not available for collections of complex types:
+1. Individual elements can be retrieved
```
corranrogue9, line 395: Note that the corollary of this is that items in the collection can be referenced by other navigation properties

PR #483, opened 2023-07-24
Remove the requirement for an enum property in the collection subsets pattern

corranrogue9:
The collection subsets pattern currently has a requirement that the base type should have an enum property which is parallel to and analogous with the dervied types. It was added to allow clients to more "easily" filter (they don't have to use `isof`). During reviews, we are often advising workload owners to *not* include the enum property, so this change removes it as a requirement, and explains that it is optional for cases where filtering using `isof` may be a concern.

### DIFF
--- a/graph/patterns/subsets.md
+++ b/graph/patterns/subsets.md
@@ -182,3 +182,36 @@ _Note: Unrelated properties on entities are omitted for easier readability._
     }
 }
 ```
+
+### Filter when base type has the "kind" enum property
+
+```HTTP
+GET .../memberships?$filter=membershipKind eq 'all'
+
+200 OK
+{
+  "value": [
+    {
+      "@odata.type": "#microsoft.graph.allMembership",
+      "membershipKind": "all"
+    },
+    ...
+  ]
+}
+```
+
+### Filter when base type lacks the "kind" enum property
+
+```HTTP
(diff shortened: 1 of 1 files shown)


dkershaw10 (approved):
LGTM.
My major issue is when the enum is used in a singleton - here it provides no value whatsoever.  It only provides value in a collection where it can be filtered.

graph/patterns/subsets.md
```diff
@@ -16,12 +16,18 @@ Existing patterns for this either have special-cased strings or have tightly cou
 
 Have an abstract base class where all variants of the subset are derived types from the base subset. For more information, see the [general subtyping guidance](./subtypes.md).
 
-The abstract base class should also hold an enum for all possible variants. The purpose of including this is to allow for easier ways to do query and filter operations on variants like `all` and `none` without relying on `isof` functions.
+The abstract base class may also optionally hold an `enum` for the different variants. If it does, the `enum` must have a member for all possible variants. The purpose of including this is to allow for easier ways to do query and filter operations on variants like `all` and `none` without relying on `isof` functions.
```
OlgaPodo, line 20: Could you please include an example of a querie with  isof  and an example without?

OlgaPodo: As much as I prefer examples from real SCDL  i think that it's better if CSDL examples match to query examples.

PR #485, opened 2023-07-31
corranrogue9/snapshot

corranrogue9:


graph/patterns/snapshots.md
```diff
@@ -0,0 +1,164 @@
...
+  <!--
+  For this example, only these properties are being stored in the snapshot. Any subset of properties from `user` is valid to store in a snapshot type. The properties should have the same names, types, nullability, etc. as the entity that is referred to.
+  -->
+  <NavigationProperty Name="user" Type="self.user" Nullable="true" /> <!--This is nullable because the user may no longer exist.-->
+</ComplexType>
+```
+
+TODO example with nested snapshots?
+
+## When to use this pattern
```
corranrogue9, line 58: From discussion with Tyler, please make it clear that only the *relevant* properties need to be on the snapshot (maybe have a comparison to a "non-snapshot" version of the same API).

PR #487, opened 2023-08-09
Olgapo-viewpoint

### LINKED ISSUE #470: Viewpoint pattern

OlgaPodo:


baywet (approved):

graph/patterns/change-notifications.md
```diff
@@ -0,0 +1,30 @@
+# Pattern name
```
corranrogue9: Did you mean to add this file?

graph/patterns/viewpoint.md
```diff
@@ -0,0 +1,112 @@
...
+For example, the `viewpoint` property can indicate whether a message is read, deleted, or flagged for a given user.
+The consistent naming convention ensures that when a developer uses Graph APIs all ‘viewpoints’ structural properties represent type specific user context across different M365 services and features.
+
+This pattern simplifies the API client logic by hiding the state transition details and providing state persistency on the server side. The server can manage the different viewpoints for the shared resource without exposing them to the client. You should also make this property filterable.
+## Issues and considerations
+
+- Because the `viewpoint` property reflects an individual user's context, it is null when accessed with application permissions.
+- Sometimes, the viewpoint can be computed on the server. In this case, an API producer should add OData annotations to the property to provide more information for downstream tools, such as SDKs and documentation generation.
+```
+    <Annotations Target="microsoft.graph.approvalItem/viewPoint">
```
corranrogue9, line 22: I didn't take the time to fully investigate, but the workload pipeline seems to fail for this annotation: https://msazure.visualstudio.com/One/_git/AD-AggregatorService-Workloads/pullrequest/7613960?_a=files&path=/Workloads/Microsoft.DirectoryServices/override/schema-Review-beta.csdl

I'm not sure if it failed because my "repro" was wrong, but I think ODL might have a limitation regarding the `Target` attribute for external annotations. I can look further if you can confirm that I've not messed up my repro.

graph/patterns/viewpoint.md
```diff
@@ -0,0 +1,112 @@
...
+        <Property Name="isHidden" Type="Edm.Boolean" />
+        <Property Name="lastMessageReadDateTime" Type="Edm.DateTimeOffset" />
+  </ComplexType>
+
+  <EntityType Name="chat" BaseType="graph.entity" ags:OwnerService="Microsoft.Teams.GraphSvc">
+        <Property Name="chatType" Type="graph.chatType" Nullable="false" />
+        <Property Name="createdDateTime" Type="Edm.DateTimeOffset" />
+        <Property Name="lastUpdatedDateTime" Type="Edm.DateTimeOffset" />              
+        <Property Name="topic" Type="Edm.String" />
+        <Property Name="viewpoint" Type="graph.chatViewpoint" />
```
corranrogue9, line 45: As I understand this pattern, I have some concerns:
1. A URL locates a resource and the data in that resource should be agnostic to the client that is retrieving it.
2. I am also unclear how this sort of pattern would affect the 2-segment permissions scopes (e.g. if there were a `Chat.Read` segment, how would `viewpoint` be affected? Or the other way around, should `Chat.Read.All` give access to `viewpoint`, or should the client be required to have `Chat.Read.All` *and* `Chat.Read`.
3. This pattern also seems to encourage the use of delegate permissions with no corresponding app-only permission, which I don't think is a good idea. Let's say I want to write a script that let's a certain group of people know that I know they have messaged me but I've been busy. I want to read my unread chats, filter them by user, and then respond to the resulting chats. As a human, I don't want to be involved at all, but I can't get the necessary viewpoint with an app-only token. 

I would prefer a pattern that reverses the direction of the relationship. The `user` entity could have a property of `chatViews` which would be a collection of `chatViewpoint`s, which could have a navigation property to the `chat` entity that it is a view of. Doing this addresses all 3 of my concerns above because the data returned for a particular viewpoint is always relative to the user the viewpoint is on.

graph/patterns/antiPatternTemplate.md
```diff
@@ -0,0 +1,24 @@
+
+# Antipattern name
+
+*name with a negative connotation*
+
+*Example: Fleabag of properties*
```
dkershaw10: Freudian slip?
```suggestion
*Example: Flatbag of properties* 
```

graph/patterns/viewpoint.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Problem
+A shared resource, such as a website or a group message, may have different states for different users who access it at different times in an organizational context. For example, user1 may read and delete a message, while user2 may not have seen it yet. This usually happens when a shared item is presented in an individual context.
+## Solution
+
+The viewpoint pattern provides a solution to how to model an individual user context on a shared resource using a `viewpoint` structural property on an API entity type.
+For example, the `viewpoint` property can indicate whether a message is read, deleted, or flagged for a given user.
+The consistent naming convention ensures that when a developer uses Graph APIs all ‘viewpoints’ structural properties represent type specific user context across different M365 services and features.
+
+This pattern simplifies the API client logic by hiding the state transition details and providing state persistency on the server side. The server can manage the different viewpoints for the shared resource without exposing them to the client. You should also make this property filterable.
```
dkershaw10:
```suggestion
This pattern simplifies the API client logic by hiding the state transition details and providing state persistency on the server side. The server can manage the different viewpoints for the shared resource without exposing them to the client. The `viewpoint` property should also be filterable.
```

graph/patterns/viewpoint.md
```diff
@@ -0,0 +1,112 @@
...
+## Issues and considerations
+
+- Because the `viewpoint` property reflects an individual user's context, it is null when accessed with application permissions.
+- Sometimes, the viewpoint can be computed on the server. In this case, an API producer should add OData annotations to the property to provide more information for downstream tools, such as SDKs and documentation generation.
+```
+    <Annotations Target="microsoft.graph.approvalItem/viewPoint">
+        <Annotation Term="Org.OData.Core.V1.Computed" Bool="true" ags:OwnerService="Microsoft.Approvals" />
```
dkershaw10: Is the ags:OwnerService annotation really part of the Computed annotation?  Is that in case of shared types?

graph/patterns/viewpoint.md
```diff
@@ -0,0 +1,112 @@
...
+```
+- An alternative to this design would be to store the user state on the client side. However, this may be problematic in some cases, because of the many devices that a user may have and the need to synchronize the state across them.
+
+## Examples
+
+### Defining a viewpoint
+
+The following example demonstrates how to define the 'viewpoint' property for the `chat` entity, where a chat is a collection of chatMessages between one or more participants: 
+```
+  <ComplexType Name="chatViewpoint" ags:WorkloadIds="Microsoft.Teams.GraphSvc">
```
dkershaw10: Please remove the ags:WorkLoadIds and ags:OwnerService annotations.

graph/patterns/viewpoint.md
```diff
@@ -0,0 +1,112 @@
...
+            "viewpoint":{
+                "lastMessageReadDateTime": "0000-01-01T00:00:00.000Z" // User hasnt read anything since no message was posted
+            }
+        }
+    ]
+}
+```
+### Updating a viewpoint
+
+You can update the `viewpoint` property only if the server does not compute it automatically. Updating the `viewpoint` property usually has a side effect, so you should use an OData action to perform the update.
```
dkershaw10: Not sure why this requires an OData action.  Why can't updating the property (via PATCH), initiate whatever other changes need to happen?  I think an action or a PATCH here would be fine, but this reads like the API producer must use an action.

graph/patterns/change-notifications.md
OlgaPodo: good catch. no change -notifications should go to a different PR.

graph/patterns/viewpoint.md
OlgaPodo: Here is what i see in  /Workloads/Microsoft.Approvals/override/schema-Prod-beta.csdl  
`<Annotations Target="self.approvalItem/viewPoint">
        <Annotation Term="Org.OData.Core.V1.Computed" Bool="true" />
      </Annotations>`

graph/patterns/antiPatternTemplate.md
OlgaPodo: i think it was a demonstration of a negative connotation.

graph/patterns/viewpoint.md
OlgaPodo: Updating a viewpoint requires a change of a shared resource created for saving client-specific state and not exposed via Graph. Therefore, there is an unpredicted side-effect and use of a function. Using PATCH, I'm not sure how to manage concurrency and ETags. IMO an action is preferable.

graph/patterns/viewpoint.md
OlgaPodo: Viewpoints are specific to a user; they don't exist for app permissions. This pattern reflects the current state where we have modeled multiple resources, such as Approvals, chats, and others, using this approach.

OlgaPodo: it also intended to hide server side complexity when necessary..

graph/patterns/viewpoint.md
OlgaPodo: updated the language to make it sound optional, please check again. Based on reviewing all five examples of the current viewpoint implementation, I found that all updatable viewpoints are using actions.

dkershaw10: I still don't get this.  The viewpoint presented is in the scope of the signed-in user.  So why can't the viewpoint update also be in the context of the signed-in user.

I might buy that ETags could be an interesting challenge here.

graph/patterns/viewpoint.md
```diff
@@ -0,0 +1,112 @@
...
+            }
+        }
+    ]
+}
+```
+### Updating a viewpoint
+
+You can update the `viewpoint` property only if the server does not compute it automatically. Updating the `viewpoint` property usually has a side effect, so you may want to use an OData action to perform the update.
+
+The following example shows marking a chat as read for a user:
```
dkershaw10: We've said that a viewpoint is specific view for the signed-in user.  However, the action below seems to imply that a privileged user (like an admin) could change the state (chat read) for another user.  Is this something we really want to show as an example?

Do the guidelines permit user A updating user B's viewpoint?

graph/patterns/antiPatternTemplate.md
dkershaw10: So do you intend to leave it as Fleabag?

graph/patterns/viewpoint.md
corranrogue9: I can align with everyone else if this is an established pattern and we are just writing it down for the first time, but I don't think I've ever heard this sort of pattern discussed.

graph/patterns/viewpoint.md
```diff
@@ -0,0 +1,112 @@
+# Viewpoint
+
+Microsoft Graph API Design Pattern
+
+
+*The viewpoint pattern provides the ability to manage an individual status of a shared object for multiple independent actors.*
```
darrelmiller:
```suggestion
*The viewpoint pattern provides the ability to manage properties of a shared object that have different values for different users.*
```

graph/patterns/viewpoint.md
```diff
@@ -0,0 +1,112 @@
...
+A shared resource, such as a website or a group message, may have different states for different users who access it at different times in an organizational context. For example, user1 may read and delete a message, while user2 may not have seen it yet. This usually happens when a shared item is presented in an individual context.
+## Solution
+
+The viewpoint pattern provides a solution to how to model an individual user context on a shared resource using a `viewpoint` structural property on an API entity type.
+For example, the `viewpoint` property can indicate whether a message is read, deleted, or flagged for a given user. 
+The consistent naming convention ensures that when a developer uses Graph APIs all ‘viewpoints’ structural properties represent type specific user context across different M365 services and features.
+
+This pattern simplifies the API client logic by hiding the state transition details and providing state persistency on the server side. The server can manage the different viewpoints for the shared resource without exposing additional complexity to the client.To support queries for a user state the `viewpoint` property should support be filterable.
```
darrelmiller:
```suggestion
This pattern simplifies the API client logic by hiding the state transition details and providing state persistency on the server side. The server can manage the different viewpoints for the shared resource without exposing additional complexity to the client.To support queries for a user state the `viewpoint` property should support filtering.
```

graph/patterns/viewpoint.md
```diff
@@ -0,0 +1,112 @@
...
+## Issues and considerations
+
+- Because the `viewpoint` property reflects an individual user's context, it is null when accessed with application permissions.
+- Sometimes, the viewpoint can be computed on the server. In this case, an API producer should add OData annotations to the property to provide more information for downstream tools, such as SDKs and documentation generation.
```
darrelmiller, line 20: I'm not sure what this sentence is trying to convey. What annotations would be useful here that are specific to viewpoints?

graph/patterns/viewpoint.md
darrelmiller: Is this necessary to mention?  Are computed properties on viewpoints any different than computed properties that are not on viewpoints?

graph/patterns/viewpoint.md
```diff
@@ -0,0 +1,112 @@
...
+```
+    <Annotations Target="microsoft.graph.approvalItem/viewPoint">
+        <Annotation Term="Org.OData.Core.V1.Computed" Bool="true" />
+    </Annotations>
+```
+- An alternative to this design would be to store the user state on the client side. However, this may be problematic in some cases, because of the many devices that a user may have and the need to synchronize the state across them.
```
darrelmiller, line 26: I don't understand how storing API state on a client would be a viable solution. What am I missing?

OlgaPodo: like storing personal bookmarks on a client browser instead of managing it for all users on a server

graph/patterns/viewpoint.md
darrelmiller: @corranrogue9 An alternative to the viewpoint design is to attach all the user specific state to a user instance.  However, there are cases where this cure is worse than the disease.  Having `/me/serviceAnnouncementStates` so that we can track which service announcements an admin has read would be a fairly horrible design to work with.

I need to double check, but my understanding is that viewpoints are always based on the identity of the calling user.  Not on a user mentioned in the path.  For this reason viewPoints should only be used on tenant relative entities.  Now I have to go find out what happens when you access the viewpoint of a chat that belongs to a user.

graph/patterns/viewpoint.md
darrelmiller: I see, so if we removed the state from the API completely.

graph/patterns/viewpoint.md
OlgaPodo: Based on the current design, the Read operation cannot return another user's viewpoint, while Update seems possible; therefore, a privileged user can potentially update viewpoints of other users. so far i can't think of a good example.

darrelmiller: There is no reason that an API cannot exist to change the read state for a user other than the calling user.  This example doesn't use the viewPoint to make this change.  
I am going to argue that you could not PATCH an entity and update a viewPoint for any user other than the signed in user.  However, I need to go ask how chatViewPoint has been implemented, because I also don't think you should be able to read the viewPoint for a different user.  e.g. via /users/{notsignedinuser}/chats/{chatId}

OlgaPodo: I think it is very much possible for an admin user to update a viewpoint. We can imagine that a tenant admin needs to pin LT announcements to a user's pinned chat list on their behalf.

OlgaPodo: @dkershaw10 i moved the blurb about update operations to issues and considerations section. Could you please look at it again?

graph/patterns/viewpoint.md
```diff
@@ -0,0 +1,136 @@
...
+
+*The viewpoint pattern provides the ability to manage properties of a shared object that have different values for different users.*
+
+## Problem
+A shared resource, such as a website or a group message, may have different states for different users who access it at different times in an organizational context. For example, user1 may read and delete a message, while user2 may not have seen it yet. This usually happens when a shared item is presented in an individual context.
+## Solution
+
+The viewpoint pattern provides a solution to how to model an individual user context on a shared resource using a `viewpoint` structural property on an API entity type.
+For example, the `viewpoint` property can indicate whether a message is read, deleted, or flagged for a given user. 
+The consistent naming convention ensures that when a developer uses Graph APIs all ‘viewpoints’ structural properties represent type specific user context across different M365 services and features.
```
dkershaw10:
```suggestion
The consistent naming convention ensures that when a developer uses Graph APIs all `viewpoint` structural properties represent type specific user context across different M365 services and features.
```

dkershaw10 (commented):
LGTM - just the one minor change.

dkershaw10 (approved):
LGTM - just the one minor change

PR #490, opened 2023-08-17
First draft of upsert pattern

dkershaw10:


graph/patterns/infrastructure-as-code.md
```diff
@@ -0,0 +1,172 @@
+# Infrastructure as code
```
dkershaw10: Not sure whether to leave this as is, or change it to the "Upsert pattern".

eketo-msft: I think it should generally be known as upsert.  IaC is an application of that pattern.

graph/patterns/infrastructure-as-code.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Solution
+
+The solution is to use an `UPSERT` pattern, to solve for the non-idempotent creation and client-provided naming problems.
+
+* For IaC scenarios, resources must use `UPSERT` semantics with an [alternate key](./alternate-key.md):
+  * Use `PATCH` with a client-provided alternate key.
+  * For a non-existent resource (specified by the alternate key) the service must handle this as a "create". As part of creation, the service must still generate the primary key value.
+  * For an existing resource (specified by the alternate key) the service must handle this as an "update.
+  * Any new alternate key, used for IaC scenarios, be called `uniqueName`.
```
eketo-msft: I don't think this should be a requirement.  Imagine adding SP today, why would we want them to use uniqueName instead of leveraging the appId?  I think uniqueName is fine as a default in lieu of something that is scenario driven.

Is there value in having another property named uniqueName?

graph/patterns/infrastructure-as-code.md
```diff
@@ -0,0 +1,172 @@
...
+  * NOTE: the service must also support `GET` using the alternate key pattern.
+* For consistent CRUD Microsoft Graph behaviors, all resources, **including** resources used in IaC scenarios, should use `POST` and a service-generated primary key, per existing guidelines, and support `GET`, `PATCH` and `DELETE` using the primary key.
+* If a service does not support `UPSERT`, then a `PATCH` call against a non-existent resource must result in an HTTP "409 conflict" error.
```
eketo-msft: This should result in HTTP 404 Not Found.

graph/patterns/infrastructure-as-code.md
```diff
@@ -0,0 +1,172 @@
...
+{
+    "displayName": "My favorite group",
+    "description": "All my favorite people in the world"
+}
+```
+
+Response:
+
+```http
+409 conflict
```
eketo-msft: 404 Not Found.

graph/patterns/infrastructure-as-code.md
```diff
@@ -0,0 +1,172 @@
...
+As a consequence, API consumers that want to offer IaC solutions, must create compensation layers that can mimic idempotent behavior. 
+For example, when creating a resource, the compensation layer must check whether the resource first exists, before trying to create or update the resource.
+
+Additionally, IaC code scripts or templates usually employ client-provided names (or keys) to track resources in a predictable manner, whereas [Microsoft Graph guidelines](../GuidelinesGraph.md#behavior-modeling) suggests use of `POST` to create new entities with service-generated keys.
+
+## Solution
+
+The solution is to use an `UPSERT` pattern, to solve for the non-idempotent creation and client-provided naming problems.
+
+* For IaC scenarios, resources must use `UPSERT` semantics with an [alternate key](./alternate-key.md):
```
eketo-msft: I don't think alternate key calls this out, but we should state that uniqueName or another client provided alternate key is immutable after being set.  If the value is null then it should be updateable as a way to backfill existing objects to use for IaC scenarios.

graph/patterns/infrastructure-as-code.md
dkershaw10: Yeah - I realized the same once I'd pushed the PR.  Will make the updates

graph/patterns/infrastructure-as-code.md
dkershaw10: It's missing the word "should".  OK how about I qualify this as it should be called `uniqueName` if there isn't already an existing property that could be used as an alternate key.

The reason I want to make the naming recommendation is that it would make things more consistent.  

I had hoped all client provided keys here would have the same name to make the extension simpler - but I guess we'll need some declarative annotations to mark which alternate keys are enabled for upsert, to allow for autogeneration of the bicep types with variably named keys..

graph/patterns/infrastructure-as-code.md
dkershaw10: Not according the the Microsoft REST API guidelines :).

graph/patterns/infrastructure-as-code.md
dkershaw10: Thanks - forgot to mention that.  That was something I wanted to say.  I need to go back to the other doc I wrote, which I think is more exhaustive.

dkershaw10: Adding this to the Issues and considerations section.

graph/patterns/infrastructure-as-code.md
dkershaw10: Resolving and tracking with the other comment.

graph/patterns/infrastructure-as-code.md
eketo-msft: Ah, so there are two scenarios...

1. Client is request UPSERT but service does not support it.  This is a 409.
2. Client is not requesting UPSERT and is updating an object that doesn't exist.  This is a 404.

Is that accurate?

eketo-msft (approved):

graph/patterns/upsert.md
```diff
@@ -0,0 +1,173 @@
+# Upsert
+
+Microsoft Graph API Design Pattern
+
+*The `UPSERT` pattern is a non-destructive idempotent operation using a client-provided key, that ensures that system resources can be deployed in a reliable, repeatable, and controlled way, typically used in Infrastructure as Code (IaC) scenarios.*
+
+## Problem
+
+Infrastructure as code (IaC) defines system resources and topologies in a descriptive manner that allows teams to manage those resources as they would code.
```
OlgaPodo: Should we say that Infrastructure as code (IaC) defines system resources and topologies in a **declarative** manner. It's an industry term i believe.

graph/patterns/upsert.md
```diff
@@ -0,0 +1,173 @@
...
+        </PropertyValue>
+      </Record>
+    </Collection>
+  </Annotation>
+  </Property>
+</EntityType>
+```
+
+### Upserting a record (creation path)
+
```
OlgaPodo, line 98: If Upsert behavior is an opt-in based on the Prefer header, then we should probably demonstrate it with an example

graph/patterns/upsert.md
```diff
@@ -0,0 +1,173 @@
...
+Create a new group, with a `uniqueName` of "Group157", exactly like before. Except in this case, this group already exists. This is a common scenario in IaC, when a deployment template is re-run multiple times.
+
+```http
+PATCH /groups(uniqueName='Group157')
+```
+
+```json
+{
+    "displayName": "My favorite group",
```
OlgaPodo, line 140: May be it make sense to provide different property values to demonstrate that PATCH will work as a simple PATCH for an existing resource.

graph/patterns/upsert.md
```diff
@@ -0,0 +1,173 @@
...
+This pattern should be adopted for resources that are managed through infrastructure as code or desired state configuration.
+
+## Issues and considerations
+
+* The addition of this new pattern (with alternate key) does not represent a breaking change.
+However, some API producers may have concerns about accidental usages of this new pattern unwittingly creating many new resources when the intent was an update.
+As a result, API producers can use the `Prefer: idempotent` to require clients to opt-in to the UPSERT behavior.
+* The client-provided alternate key must be immutable after being set. If its value is null then it should be settable as a way to backfill existing resources for use in IaC scenarios.
+* API producers could use `PUT` operations to create or update, but generally this approach is not recommended due to the destructive nature of `PUT`'s replace semantics.
+* API producers could to use `UPSERT` with a primary (client-provided) key and this may be appropriate for some scenarios. However, the recommendation is for resources to support creation using `POST` and a service-generated primary key, for consistency reasons.
```
OlgaPodo:
```NIT:
* API producers could use `UPSERT` 
```

graph/patterns/upsert.md
```diff
@@ -0,0 +1,173 @@
...
+* API producers may annotate entity sets, singletons and collections to indicate that entities can be "upserted". The example below shows this annotation for the `groups` entity set.
```
OlgaPodo, line 55: I'm not sure how Upsert will work for a singleton on Graph.

graph/patterns/upsert.md
```diff
@@ -0,0 +1,173 @@
...
+  * For an existing resource (specified by the alternate key) the service must handle this as an "update.
+  * Any new alternate key, used for IaC scenarios, should be called `uniqueName`, if there isn't already an existing property that could be used as an alternate key.
+* NOTE: the service must also support `GET` using the alternate key pattern.
+* For consistent CRUD Microsoft Graph behaviors, all resources, **including** resources used in IaC scenarios, should use `POST` and a service-generated primary key, per existing guidelines, and support `GET`, `PATCH` and `DELETE` using the primary key.
+* If a service does not support `UPSERT`, then a `PATCH` call against a non-existent resource must result in an HTTP "409 conflict" error.
+
+This solution allows for existing resources that follow Microsoft Graph conventions for CRUD operations to add `UPSERT` without impacting existing apps or functionality.
+
+Ideally, all new entity types should support an `UPSERT` mechanism, especially if the resource is likely be used in IaC scenarios.
```
OlgaPodo: Should we use `Upsert` like in pascal case instead of `UPSERT` in the article because for a newbie just `UPSERT` looks like another HTTP method? It may be silly but i do expect a question if it's a new HTTP method.

OlgaPodo: I think that Upsert would be recommended for entities supporting control plane APIs, not necessarily for all new entities.

graph/patterns/upsert.md
```diff
@@ -0,0 +1,173 @@
...
+## Solution
+
+The solution is to use an `UPSERT` pattern, to solve for the non-idempotent creation and client-provided naming problems.
+
+* For IaC scenarios, resources must use `UPSERT` semantics with an [alternate key](./alternate-key.md):
+  * Use `PATCH` with a client-provided alternate key.
+  * For a non-existent resource (specified by the alternate key) the service must handle this as a "create". As part of creation, the service must still generate the primary key value.
+  * For an existing resource (specified by the alternate key) the service must handle this as an "update.
+  * Any new alternate key, used for IaC scenarios, should be called `uniqueName`, if there isn't already an existing property that could be used as an alternate key.
+* NOTE: the service must also support `GET` using the alternate key pattern.
```
OlgaPodo: and probably return alternate key by default?

graph/patterns/upsert.md
dkershaw10: Happy to make that change.  I took most of this text straight from Azure docs :)

graph/patterns/upsert.md
dkershaw10: I'm not a big fan of using the term control-plane with Graph. Graph is su0ch a mixed bag of resources/entities, and it's not like we mark them as "control-plane" or "data-plane".
However, I think control-plane probably resonates with folks - and if they think of their APIs as being either administrative or control plane APIs, then they should strongly consider supporting the Upsert mechanism.

graph/patterns/upsert.md
dkershaw10: Good catch.  I'll need to check this.  It's not like a singleton needs a client provided key.

graph/patterns/upsert.md
dkershaw10: Not planning to make this change.  I would expect that to be part of the alternate key pattern.
For IaC scenarios, we actually need full representation responses, so the alt-key would always be present in the response.

graph/patterns/upsert.md
dkershaw10: @mikepizzo what is the behaviour for upsert to a singleton?
My guess is that it's still preferable to have a client-provided key in these cases as this allows for referencing in the template, and in things like Azure Deployment Stacks.

However, there'll be some singletons (like a Policy) that are never created, but always exist in the system, and can only be PATCHed, that wouldn't have a client-provided key (really they're keyless singletons).

graph/patterns/upsert.md
dkershaw10: @OlgaPodo Did you want me to add a new scenario where I show a regular update, that changes properties?
This particular one is explicitly the same to demonstrate that the outcome is predictable, when the template is run multiple times.

graph/GuidelinesGraph.md
```diff
@@ -256,7 +256,8 @@ If possible, APIs SHOULD use resource-based designs with standard HTTP methods r
 
 |  Microsoft Graph rules for modeling behavior                     |
 |------------------------------------------------------------------|
-| :heavy_check_mark: **MUST** use POST to create new entities in insertable entity sets or collections.<BR>This approach requires the server to produce system generated identities. |
+| :heavy_check_mark: **MUST** use POST to create new entities in insertable entity sets or collections.<BR>This approach requires the server to produce a system generated key. |
+| :heavy_check_mark: **MAY** use PATCH to create new entities in insertable entity sets or collections.<BR>This [Upsert](./patterns/upsert.md) approach requires the caller to provide a key. |
```
dkershaw10: @OlgaPodo Not sure how to do this one.
It's not really a MUST or a SHOULD, but more of a MAY.  Except we don't have MAY in our legend...

OlgaPodo: two entries don't work together because the first one will always take precedence. Can we express criteria for PATCH better something like 'insertable entity sets or collections with customer provided unique keys'? and something corresponding for the first one? then we can use SHOULD for both use case IMO

dkershaw10: I don't think we should say both are "should".  I think the POST is a MUST, for consistency.
The Upsert is an additional thing that can be supported.

I've made an update that hopefully works for you?

graph/patterns/upsert.md
mikepizzo: Note that a PATCH to a singleton doesn't use a primary/alternate key in the URL; it's just a regular PATCH to the navigation property. Upsert should apply here as well; if I do a PATCH to users/1/calendar, and the calendar doesn't exist (it is null) it should be created with the provided values, ortherwise it should be updated with the provided values.  As opposed to PUT which replaces the existing entity with the provided values and sets unspecified values to default.

graph/patterns/upsert.md
```diff
@@ -0,0 +1,209 @@
...
+
+## When to use this pattern
+
+This pattern should be adopted for resources that are managed through infrastructure as code or desired state configuration.
+
+## Issues and considerations
+
+* The addition of this new pattern (with alternate key) does not represent a breaking change.
+However, some API producers may have concerns about accidental usages of this new pattern unwittingly creating many new resources when the intent was an update.
+As a result, API producers can use the `Prefer: idempotent` to require clients to opt-in to the Upsert behavior.
```
mikepizzo: >Prefer: idempotent` [](http://example.com/codeflow?start=40&length=19)

Not a huge fan of this new preference.
1) Is it really a preference? Can services ignore?
2) Is this something we would register w/IANA?
3) This is the same as providing If-Match="*" to force update-only, except has the opposite default, right?
4) How do we document where the client has to specify this versus where it's the default?

I would rather existing workloads that are worried about backward compatibility not support upsert for existing client-specified keys; I think there are few instances of this.

graph/patterns/upsert.md
```diff
@@ -0,0 +1,209 @@
...
+As a consequence, API consumers that want to offer IaC solutions, must create compensation layers that can mimic idempotent behavior. 
+For example, when creating a resource, the compensation layer must check whether the resource first exists, before trying to create or update the resource.
+
+Additionally, IaC code scripts or templates usually employ client-provided names (or keys) to track resources in a predictable manner, whereas [Microsoft Graph guidelines](../GuidelinesGraph.md#behavior-modeling) suggests use of `POST` to create new entities with service-generated keys.
+
+## Solution
+
+The solution is to use an `Upsert` pattern, to solve for the non-idempotent creation and client-provided naming problems.
+
+* For IaC scenarios, resources must use `Upsert` semantics with an [alternate key](./alternate-key.md):
```
mikepizzo: >[alternate key](./alternate-key.md) [](http://example.com/codeflow?start=67&length=35)

this pattern should apply to any client-provided key. if the primary key is server generated then the guidance should be to define an alternate key to support idempotent creates, but if there is a natural key that is meaningful to the client, then the service should use that as the key and support upserts against when doing a PATCH to the canonical URL.  We wont force that on existing APIs that use natural keys if there is a concern about backward compatibility, but new APIs with natural keys should support upsert using those natural keys.

graph/patterns/upsert.md
```diff
@@ -0,0 +1,209 @@
...
+  * Use `PATCH` with a client-provided alternate key.
+  * For a non-existent resource (specified by the alternate key) the service must handle this as a "create". As part of creation, the service must still generate the primary key value.
+  * For an existing resource (specified by the alternate key) the service must handle this as an "update.
+  * Any new alternate key, used for IaC scenarios, should be called `uniqueName`, if there isn't already an existing property that could be used as an alternate key.
+* NOTE: the service must also support `GET` using the alternate key pattern.
+* For consistent CRUD Microsoft Graph behaviors, all resources, **including** resources used in IaC scenarios, should use `POST` and a service-generated primary key, per existing guidelines, and support `GET`, `PATCH` and `DELETE` using the primary key.
```
mikepizzo: >and a service-generated primary key [](http://example.com/codeflow?start=129&length=35)

why should services use a service-generated primary key if there is a meaningful natural key?

graph/patterns/upsert.md
```diff
@@ -0,0 +1,209 @@
...
+* If a service does not support `Upsert`, then a `PATCH` call against a non-existent resource must result in an HTTP "409 conflict" error.
```
mikepizzo: >409 conflict [](http://example.com/codeflow?start=118&length=12)

I don't recall why we decided on this versus a 405 method not allowed.  Is this documented in an ADR? Can we reference that reasoning somehow/somewhere?

graph/patterns/upsert.md
```diff
@@ -0,0 +1,209 @@
...
+This pattern should be adopted for resources that are managed through infrastructure as code or desired state configuration.
+
+## Issues and considerations
+
+* The addition of this new pattern (with alternate key) does not represent a breaking change.
+However, some API producers may have concerns about accidental usages of this new pattern unwittingly creating many new resources when the intent was an update.
+As a result, API producers can use the `Prefer: idempotent` to require clients to opt-in to the Upsert behavior.
+* The client-provided alternate key must be immutable after being set. If its value is null then it should be settable as a way to backfill existing resources for use in IaC scenarios.
+* API producers could use `PUT` operations to create or update, but generally this approach is not recommended due to the destructive nature of `PUT`'s replace semantics.
+* API producers could use `Upsert` with a primary (client-provided) key and this may be appropriate for some scenarios. However, the recommendation is for resources to support creation using `POST` and a service-generated primary key, for consistency reasons.
```
mikepizzo: >support creation using `POST` and a service-generated primary key [](http://example.com/codeflow?start=168&length=65)

I would separate these. Services should always support POST to the collection URL.  For server-generated keys, this should return the server generated key.  For client-provided keys the client should provide the key as part of the payload.  This is in addition to supporting PATCH to the url of the resource being upserted (using the canonical url for client provided keys, or alternate key if the primary key is server generated)

graph/patterns/upsert.md
```diff
@@ -0,0 +1,209 @@
...
+```xml
+<EntityType Name="group">
+  <Key>
+    <PropertyRef Name="id"/>
+  </Key>
+  <Property Name="id" Type="Edm.String"/> 
+  <Property Name="uniqueName" Type="Edm.String"/>
+  <Property Name="displayName" Type="Edm.String"/>
+  <Property Name="description" Type="Edm.String"/> 
+  <Annotation Term="OData.Community.Keys.V1.AlternateKeys">
```
mikepizzo: >OData.Community.Keys.V1. [](http://example.com/codeflow?start=20&length=24)

can we move away from using OData.Community.Keys.V1, since this is now in Core? ([https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.xml#L479](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.xml#L479))?

graph/patterns/upsert.md
```diff
@@ -0,0 +1,209 @@
...
+
+This solution allows for existing resources that follow Microsoft Graph conventions for CRUD operations to add `Upsert` without impacting existing apps or functionality.
+
+Ideally, all new entity types should support an `Upsert` mechanism, especially where they support control-plane APIs, or are used in admin style or IaC scenarios.
+
+## When to use this pattern
+
+This pattern should be adopted for resources that are managed through infrastructure as code or desired state configuration.
+
+## Issues and considerations
```
mikepizzo, line 45: >Issues and considerations [](http://example.com/codeflow?start=3&length=25)

Do we want to mention forcing update behavior with If-Match="*" (fails if already exists)?  What about forcing create behavior with If-None-Match="*"?

I think the former is useful, the later probably not terribly useful.

mikepizzo (changes requested):
:clock1:

graph/patterns/upsert.md
dkershaw10: I believe it's in the Microsoft REST API guidelines.  Happy to change this.

dkershaw10, line 30: @mikepizzo - so want to say that API producers still need to offer POST to create for consistency.
If there is a natural client-provided primary key, then when using POST, I assume this should be supplied in the body?

Or are we thinking of saying, that in this case it's unnecessary to support POST?

graph/patterns/upsert.md
dkershaw10: I believe it's in the Microsoft REST API guidelines. Happy to change this.
I think @eketo-msft wants this to be a 404.

graph/patterns/upsert.md
dkershaw10: @mikepizzo we had this conversation multiple times.  Not sure where to take this.  It never got resolved with a suitable option, and Graph API council were comfortable with a Prefer header here.

We implemented the prefer header, because If-Match="*" didn't suit the purpose, and the Directory team insisted on some header to indicate this behaviour, over their previous default.

I think we may actually remove this prefer header before we go GA, as services are more comfortable with the approach. So I don't want to register this with IANA.

At the moment, the only client likely to use this is the Graph Bicep Extension.  Other client will likely be a controlled set.  We would recommend they all use the Prefer header.

graph/patterns/upsert.md
dkershaw10: Is supporting these headers required? Not sure either is useful, given the current behaviour.

For IaC scenarios, the point is that the template author doesn't know whether the resource is created or not - they just specify the resource, plus the deployment engine also doesn't know the state, so the if-Match and If-None_Match headers aren't particularly useful.

graph/patterns/upsert.md
dkershaw10: @OlgaPodo - do I need to provide another example for singletons, or should I just add a note to the considerations?  I'm leaning towards the latter.

graph/patterns/upsert.md
dkershaw10: Argh - not sure how that snuck in.  Good catch
I copied it from the Directory schema :(

graph/patterns/upsert.md
```diff
@@ -0,0 +1,214 @@
...
+For example, when creating a resource, the compensation layer must check whether the resource first exists, before trying to create or update the resource.
+
+Additionally, IaC code scripts or templates usually employ client-provided names (or keys) to track resources in a predictable manner, whereas [Microsoft Graph guidelines](../GuidelinesGraph.md#behavior-modeling) suggests use of `POST` to create new entities with service-generated keys.
+
+## Solution
+
+The solution is to use an `Upsert` pattern, to solve for the non-idempotent creation and client-provided naming problems.
+
+* For IaC scenarios, resources must use `Upsert` semantics with a client-provided key:
+  * Use `PATCH` with a client-provided key:
```
mikepizzo: I think we need to be clear here that it is a PATCH with the client-provided key *in the URL*.  Probably need an example as well. I've seen folks doing a PATCH to the url of the collection with the key in the body.

mikepizzo (dismissed):

graph/patterns/upsert.md
mikepizzo: This is not needed for IaC scenarios, and fail if exists/forcing create are not very interesting

However, if the service does support upsert, and the client (not IaC) wants to be able to update a resource only if it already exists (not create if it doesn't) then the If-Match=* is useful.

graph/patterns/upsert.md
mikepizzo: Yes; 404 makes the most sense in this case.

graph/patterns/upsert.md
mikepizzo: As per discussion in API Council; I don't think this should be called "Idempotent", but I am okay introducing a new preference for:
-existing services with client defined keys
-want to start supporting upsert
-are worried about backward compat

graph/patterns/upsert.md
dkershaw10: All the examples below have this. Adding "in the URL".

graph/patterns/upsert.md
dkershaw10: Updating...

graph/patterns/upsert.md
dkershaw10: Updated

graph/patterns/infrastructure-as-code.md
dkershaw10: Updated with 404

PR #493, opened 2023-08-23
Add enum guidance

### LINKED ISSUE #468: When to use Enum type guidance
Describe criteria when to use enum over string property

XVincentX:
Closes #468

### DIFF
--- a/graph/GuidelinesGraph.md
+++ b/graph/GuidelinesGraph.md
@@ -323,7 +323,6 @@ For a complete mapping of error codes to HTTP statuses, see
 <a name="api-contract-and-non-backward-compatible-changes"></a>
 
 
-
 ## API contract and non-backward compatible changes
 
 The Microsoft Graph definition of breaking changes is based on the
--- a/graph/GuidelinesGraph.md
+++ b/graph/GuidelinesGraph.md
@@ -339,7 +339,7 @@ In OData, enums represent a subset of the nominal type they rely on, and are esp
 
 #### Enum or Booleans
 
-Enumerations are a good alternative to booleans when one of the two values (`true`, `false`) conveyes other possible values not yet conceived. Let's assume we have an `Error` type and a property to communicate how to display it:
+Enumerations are a good alternative to Booleans when one of the two values (`true`, `false`) conveys other possible values not yet conceived. Let's assume we have an `Error` type and a property to communicate how to display it:
 
 ```xml
 <ComplexType Name="Error">
--- a/graph/GuidelinesGraph.md
+++ b/graph/GuidelinesGraph.md
@@ -318,7 +318,7 @@ For a complete mapping of error codes to HTTP statuses, see
 
 ### Enums
 
-In OData, enum are nominal types representing a subset of the nominal type they use, and are especially useful in cases where certain properties have predefined, limited options.
+In OData, enums represent a subset of the nominal type they rely on, and are especially useful in cases where certain properties have predefined, limited options.
 
 ```xml
 <EnumType Name="Color">
--- a/graph/GuidelinesGraph.md
+++ b/graph/GuidelinesGraph.md
@@ -398,6 +398,36 @@ By using an evolvable enum, instead, all we need to do is to add new members:
 </EnumType>
 ```
 
+Similarly speaking, if you find yourself using a `nullable` Enum, that is a indication that maybe what you are trying to model is something that has 3 states and an enum is more appropraite. For instance, let's assume we have a boolean property called `syncEnabled`, where `null` means that the value is undefined and inherited from the general tenant configuration. Instead of modelling like a boolean:
+
+```xml
+<Property Name="syncEnabled" Type="Edm.Boolean" Nullable="true"/>
+```
+
+An enum not only better conveys the message:
+
+```xml
+<EnumType Name="syncState">
+    <Member Name="enabled" Value="0" />
+    <Member Name="disabled" Value="1" />
+    <Member Name="tenantInherit" Value="2" />
+    <Member Name="unknownFutureValue" Value="3" />
+</EnumType>
+```
+
+but it is also open for future scenarios:
+
+```diff
+<EnumType Name="syncState">
+    <Member Name="enabled" Value="0" />
+    <Member Name="disabled" Value="1" />
+    <Member Name="tenantInherit" Value="2" />
+    <Member Name="unknownFutureValue" Value="3" />
++   <Member Name="groupInherit" Value="4" />
+</EnumType>
+```
+
+
 
 ## API contract and non-backward compatible changes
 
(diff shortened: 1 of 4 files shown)


graph/GuidelinesGraph.md
```diff
@@ -326,6 +327,78 @@ For a complete mapping of error codes to HTTP statuses, see
...
+#### Pros
+
+- Our SDK generators will translate the enum to the best representation of the target programming language, resulting in a better developer experience and free client side validation
+
+#### Cons
+
+- Adding a new value requires to go through a (generally fast) API Review
+- If the enum is not [evolvable](./patterns/evolvable-enums.md), adding a new value is a breaking change and will generally not be allowed
+
+#### Enum or Booleans
```
dkershaw10: Another thing we see a **lot** of, are that folks define collections of enum, rather than using flags enums.  Can we add something about that too please?

Also Garrett has previously pointed out that a nullable enum is an indicator that the enum should have an additional member for the null state.  Similarly, a nullable Boolean indicates tri-state, which should be modelled as an enum instead.

XVincentX: Adding a paragraph about this now.

graph/GuidelinesGraph.md
```diff
@@ -326,6 +316,133 @@ For a complete mapping of error codes to HTTP statuses, see
...
+```xml
+<Property Name="displayMethods" Type="Collection(displayMethod)"/>
+```
+
+However, [Flagged Enums](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#_Toc38530378) can model this use case scenario:
+
+```diff
+- <EnumType Name="displayMethod">
++ <EnumType Name="displayMethod" isFlag="true">
+     <Member Name="tip" Value="0" />
```
eketo-msft: Flags enum should start at value 1.  You could argue that 0 could be added as a special value for None but it wouldn't be useful for bitwise operations.

graph/GuidelinesGraph.md
```diff
@@ -326,6 +316,136 @@ For a complete mapping of error codes to HTTP statuses, see
 
 <a name="api-contract-and-non-backward-compatible-changes"></a>
 
+### Enums
```
dkershaw10: @OlgaPodo does this belong in the guidelines or as a pattern?  Works for me either way, but just double checking.

graph/GuidelinesGraph.md
```diff
@@ -326,6 +316,136 @@ For a complete mapping of error codes to HTTP statuses, see
...
+
+In OData, enum are nominal types representing a subset of the nominal type they use, and are especially useful in cases where certain properties have predefined, limited options.
```
dkershaw10: Can we rephrase?  There are a lot of nominal types here.
Also I think you meant enums (not enum).

graph/GuidelinesGraph.md
```diff
@@ -326,6 +316,136 @@ For a complete mapping of error codes to HTTP statuses, see
...
+- Our SDK generators will translate the enum to the best representation of the target programming language, resulting in a better developer experience and free client side validation
+
+#### Cons
+
+- Adding a new value requires to go through a (generally fast) API Review
+- If the enum is not [evolvable](./patterns/evolvable-enums.md), adding a new value is a breaking change and will generally not be allowed
+
+#### Enum or Booleans
+
+Enumerations are a good alternative to booleans when one of the two values (`true`, `false`) conveyes other possible values not yet conceived. Let's assume we have an `Error` type and a property to communicate how to display it:
```
```suggestion
Enumerations are a good alternative to Booleans when one of the two values (`true`, `false`) conveys other possible values not yet conceived. Let's assume we have an `Error` type and a property to communicate how to display it:
```

graph/GuidelinesGraph.md
XVincentX: Done

graph/GuidelinesGraph.md
```diff
@@ -330,6 +319,136 @@ For a complete mapping of error codes to HTTP statuses, see
 
 <a name="api-contract-and-non-backward-compatible-changes"></a>
 
+### Enums
+
+In OData, enums represent a subset of the nominal type they rely on, and are especially useful in cases where certain properties have predefined, limited options.
```
OlgaPodo: @XVincentX It seems that the topic of Enums is a bit out of context and sequence as it placed right now. We clearly need a description of when to use enums for modelling, but it falls under the 'Resource modelling' section in the main guidelines IMO. I would also make the Enum page a separate page and link to the 'Resource modelling' section.

graph/GuidelinesGraph.md
```diff
@@ -330,6 +319,136 @@ For a complete mapping of error codes to HTTP statuses, see
...
+
+- Adding a new value requires to go through a (generally fast) API Review
+- If the enum is not [evolvable](./patterns/evolvable-enums.md), adding a new value is a breaking change and will generally not be allowed
+
+#### Enum or Booleans
+
+Enumerations are a good alternative to Booleans when one of the two values (`true`, `false`) conveys other possible values not yet conceived. Let's assume we have an `Error` type and a property to communicate how to display it:
+
+```xml
+<ComplexType Name="Error">
```
OlgaPodo: NIT: all type names should be camel case. Also, we have a ComplexType "publicError" that has dependencies on a utility library. I would take a different real example to avoid creating confusion for our customers.

graph/GuidelinesGraph.md
XVincentX: Moved

graph/GuidelinesGraph.md
XVincentX: I can change it, but the example has a property that does **not** belong to `publicError`. I will let you decide what you want me to do with this

OlgaPodo: Yes. Please create a different example or, better yet, find a good existing one. Also, since we do not allow new non-evolvable enums, why don't you add these considerations to the existing pattern?

XVincentX: I have called this `publicNotification`

>  Also, since we do not allow new non-evolvable enums, why don't you add these considerations to the existing pattern?

If that is the case then evolvable enum is not a pattern anymore, it's a mandatory guideline. Should we move the section where appropriate?

eketo-msft (approved):

graph/GuidelinesGraph.md
OlgaPodo: I wouldn't say that all Enums are evolvable. For example, Day of Week wouldn't be. Also, it's up to the API producer to choose whether to use an Enum or another construct for their use cases.

XVincentX: > Also, since we do not allow new non-evolvable enums, why don't you add these considerations to the existing pattern?

> I wouldn't say that all Enums are evolvable. For example, Day of Week wouldn't be. Also, it's up to the API producer to choose whether to use an Enum or another construct for their use cases.

I am confused now. Do enum need to be evolvable or not?

PR #495, opened 2023-08-28
Merge enums

### LINKED ISSUE #494: Fix content table in the main Graph guidelines. merger enum and evolvable enum patterns

XVincentX:
Closes #494


graph/patterns/enums.md
```diff
@@ -1,4 +1,4 @@
-### Enums
+# Enums
```
OlgaPodo, line 1: @XVincentX could you please use the pattern template https://github.com/microsoft/api-guidelines/blob/vNext/graph/patterns/PatternDescriptionTemplate.md? Describing 1. what problem we are solving by using enums, 2. solution aka modeling enums, 3. when this solution is better than other patterns and 4. usage examples. The idea is to keep all articles in the same format. Google has a similar approach to AIPs https://google.aip.dev/8

graph/patterns/enums.md
```diff
@@ -100,7 +100,7 @@ but it is also open for future scenarios:
 
 Additionally speaking, depending on the situation, a nullable enum can very likely be avoided by adding a `none` member.
 
-#### Flag Enums or Collection of Enums
+## Flag Enums or Collection of Enums
 
 In case an enum can have multiple values at the same time the tentation is to model the property as a collection of Enums:
```
OlgaPodo, line 113: Did you observe this behavior in practice? may be we need to describe only preferred option like flagged enums?

graph/patterns/enums.md
```diff
@@ -1,25 +1,33 @@
...
-    <Member Name="Red" Value="0" />
-    <Member Name="Green" Value="1" />
-    <Member Name="Blue" Value="2" />
-</EnumType>
-```
+Enums represent a subset of the nominal type they rely on, and are especially useful in cases where certain properties have predefined, limited options.
+
+## Problem
+
+The API Design requires that a specific property of a data structure has a finite and limited set of values, and such limitation needs to be clearly communicated to the user and give the developer the same structure
```
OlgaPodo, line 9: Could you please ground the problem in a real design scenario. A good example is at https://github.com/microsoft/api-guidelines/blob/vNext/graph/patterns/alternate-key.md#problem or https://github.com/microsoft/api-guidelines/blob/vNext/graph/patterns/evolvable-enums.md#problem.

graph/patterns/enums.md
```diff
@@ -1,25 +1,33 @@
...
+
+## Solution
+
+In OData, Enums represent a subset of the nominal type they rely on, and are especially useful in cases where certain properties have predefined, limited options.
```
OlgaPodo, line 14: Enum solution must be based on the evolvable enum. Not evolvable enums are exceptions. Could you please merge all template sections together? it doesn't make sense to have the same section in one document twice.

graph/patterns/enums.md
```diff
@@ -1,25 +1,33 @@
...
+## Issues and considerations
+
+### Pros
 
 - Our SDK generators will translate the enum to the best representation of the target programming language, resulting in a better developer experience and free client side validation
 
-#### Cons
+### Cons
 
 - Adding a new value requires to go through a (generally fast) API Review
```
OlgaPodo, line 27: Our internal process shouldn't impact API design considerations especially as they published for external users.

graph/patterns/enums.md
```diff
@@ -1,25 +1,33 @@
...
-#### Pros
+## When to use this pattern
+
+Everytime a specific property of a data structure has a finite and limited set of values, and such limitation needs to be clearly communicated to the user and give the developer the same structure.
+
+## Issues and considerations
+
+### Pros
 
 - Our SDK generators will translate the enum to the best representation of the target programming language, resulting in a better developer experience and free client side validation
```
OlgaPodo, line 23: One of the pros for Enums is that clients can write a processing logic based on the enum values.

XVincentX: Isn't that true for any property? Clients write processing logic based on values - not too sure what is different here

graph/patterns/enums.md
XVincentX: Well, it is a clear side effect which is annoying for a lot of workload owners. I can remove it if you want, but we're missing a particular part that affect us. Your call

OlgaPodo: Please remove this. We might offer some process guidance in the future, but I don't think it's our area of expertise now or something we are working on.

PR #496, opened 2023-08-30
New default properties pattern

dkershaw10:
Introduce a new pattern for default properties that also solves for the discovery problem (or non-default properties) by using an instance annotation.

### DIFF
--- a/graph/patterns/default-properties.md
+++ b/graph/patterns/default-properties.md
@@ -1,0 +1,201 @@
+# Default properties
+
+Microsoft Graph API Design Pattern
+
+*The default properties pattern allows API producers to omit specific properties from the response unless they are explicitly requested using `$select`. The pattern also uses an instance annotation to inform callers that other properties are also available. The same annotation is also use to encourage callers to use `$select`.*
+
+## Problem
+
+API producers want to control the set of properties that their entities return by default, when the incoming request does not specify a `$select`. This can be desirable when an entity type has many properties or an API producer needs to add properties that are computationally expensive to return by default.
+
+## Solution
+
+For incoming requests targeting an entity type where the caller does not specify a `$select`` clause, API producers **may** return a subset of the entity type's properties, omitting computationally expensive properties.
+
+To get the non-default properties of an entity type, callers must explicitly request them using `$select`.
+
+## When to use this pattern
+
+API producers should use this pattern when adding expensive or non-performant properties to an existing entity type, or when adding properties to an entity type that has already grown too large (with more than 20 properties).
+
+## Issues and considerations
+
+- Do **not** rely on the `ags:Default` schema annotation for default properties functionality, as this is a legacy implementation. Returning default properties **must** be implemented by API producers.
+- Changing a default property to non-default is considered a breaking change.
+- One of the challenges with default properties is informing developers that the response does not contain the full set of properties. To solve for this discovery problem, if the response contains default properties only, then:
+  - the response **must** contain a `@microsoft.graph.tips` instance annotation.
+  - the `@microsoft.graph.tips` instance annotation **must** only be emitted if the client uses "developer mode" via the `Prefer: ms-graph-dev-mode` HTTP request header. It is expected that this header will only be used by client developer and scripting tools like Graph Explorer, the Microsoft Graph Postman collections, and Microsoft Graph PowerShell.
+  - the `@microsoft.graph.tips` instance annotation value **must** contain "This request only returns a subset of the resource's properties. Your app will need to use $select to return non-default properties. To find out what other properties are available for this resource see https://learn.microsoft.com/graph/api/resources/{entityTypeName}".
+- Callers must be able to use `$filter` with non-default properties, even though they won't show up by default in the response.
+
+Additionally, for incoming requests targeting an entity type where the caller does not specify a `$select` clause, the API Gateway Service will inject a `@microsoft.graph.tips` instance annotation, informing callers to use $select, when in "developer mode".
+API producers who use [response passthrough](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/391069/Enabling-response-passthrough) must also implement this behavior, supplying the same information as shown in the [examples section below](#calling-an-api-without-using-select).
+
+## Examples
+
+In this example we'll use the following `channel` entity type.
+
+```xml
+<EntityType Name="channel" BaseType="graph.entity">
+  <Property Name="createdDateTime" Type="Edm.DateTimeOffset"/>
+  <Property Name="description" Type="Edm.String"/>
+  <Property Name="displayName" Type="Edm.String" Nullable="false"/>
+  <Property Name="email" Type="Edm.String"/>
+  <Property Name="isFavoriteByDefault" Type="Edm.Boolean"/>
+  <Property Name="membershipType" Type="graph.channelMembershipType"/>
+  <!-- moderationSettings is a new computed property that is very expensive -->
+  <Property Name="moderationSettings" Type="graph.channelModerationSettings"/> 
+  <Property Name="webUrl" Type="Edm.String"/>
+  <Property Name="filesFolderWebUrl" Type="Edm.String"/>
+</EntityType>
+```
+
+In this scenario, the API producer wants to add the `moderationSettings` property to the `channel` entity type.
+But when paging through 1000 channels at a time, this additional property will introduce a considerable increase in the response times.
+The API producer will use the default properties pattern here, and **not** return `moderationSettings` by default.
+
+### Calling an API with default properties
+
+In this example, the caller, using Graph Explorer, does not use $select, and the API returns just the default properties.
+
+#### Request
+
+```http
+GET /teams/{id}/channels
+Prefer: ms-graph-dev-mode
+```
+
+#### Response
+
+```http
+200 ok
+Content-type: application/json
+```
+
+```json
+{
+    "@odata.context": "https://graph.microsoft.com/beta/$metadata#Collection(microsoft.graph.channel)",
+    "@microsoft.graph.tips": "This request only returns a subset of the resource properties. Your app will need to use $select to return non-default properties. To find out what other properties are supported for this resource, please see the Properties section in https://learn.microsoft.com/graph/api/resources/channel.",
+    "value": [
+        {
+            "displayName": "My First Shared Channel",
+            "description": "This is my first shared channels",
+            "id": "19:PZC_kAPAm12RPBMkEaJyXaY_d2PE6mJV6MzO1EiCbnk1@thread.tacv2",
+            "membershipType": "shared",
+            "email": "someemail@dot.com",
+            "webUrl": "webUrl-value",
+            "filesFolderWebUrl": "sharePointUrl-value",
+            "tenantId": "tenantId-value",
+            "isFavoriteByDefault": null,
+            "createdDateTime": "2019-08-07T19:00:00Z"
+        },
+        {
+            "displayName": "My Second Private Channel",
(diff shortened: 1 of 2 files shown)


graph/patterns/default-properties.md
```diff
@@ -0,0 +1,201 @@
...
+
+```http
+200 ok
+Content-type: application/json
+```
+
+```json
+{
+    "@odata.context": "https://graph.microsoft.com/beta/$metadata#Collection(microsoft.graph.channel)",
+    "@microsoft.graph.tips": "This request only returns a subset of the resource properties. Your app will need to use $select to return non-default properties. To find out what other properties are supported for this resource, please see the Properties section in https://learn.microsoft.com/graph/api/resources/channel.",
```
XVincentX, line 78: Why is this an human string and not something that a machine could read and understand what's happening?

dkershaw10: @XVincentX because it's absolutely intended for humans and not machines :)
It's absolutely not intended for machines, because we want to inform developers so that they can take corrective action in their code.  These messages are only provided in "developer mode" - so pointless when a developer doesn't see them.

For default properties we get a **ton** of support calls from developers who don't understand why certain properties are not showing up.

XVincentX: That's interesting. My point is that with a `code` it can be for **both** developers and machines.

If we structure the tip as a Graph Errror we could have a code, a message with as well an help link where people could learn more. Additionally, our SDKs find such specific code, they could emit a warning (on the standard output) that would be very much visible to the developer and go from there. I would argue it will reach way more people that way.

dkershaw10: I was hoping to keep the tip as a string, but I see your point. Not sure we need the full error object, but something like this?

```json
  "@microsoft.graph.tips": {
       "code": "Default_Properties",
       "message": "This request only returns a subset of the resource properties. Your app will need to use $select to return non-default properties. To find out what other properties are supported for this resource, please see the Properties section in https://learn.microsoft.com/graph/api/resources/channel."
  }
```

XVincentX: That is what I would suggest. We can run it by other people to see if other people think it's a good idea.

XVincentX: Also, if coming with codes for now is too much, let's at least leave it as object instead of a string, so that we can do that in the future at least

```json
{
  "@microsoft.graph.tips": [{
    "message": "This request only returns a subset of the resource properties. Your app will need to use $select to return non-default properties. To find out what other properties are supported for this resource, please see the Properties section in https://learn.microsoft.com/graph/api/resources/channel."
  }
]}
```

graph/patterns/default-properties.md
```diff
@@ -0,0 +1,201 @@
...
+
+*The default properties pattern allows API producers to omit specific properties from the response unless they are explicitly requested using `$select`. The pattern also uses an instance annotation to inform callers that other properties are also available. The same annotation is also use to encourage callers to use `$select`.*
+
+## Problem
+
+API producers want to control the set of properties that their entities return by default, when the incoming request does not specify a `$select`. This can be desirable when an entity type has many properties or an API producer needs to add properties that are computationally expensive to return by default.
+
+## Solution
+
+For incoming requests targeting an entity type where the caller does not specify a `$select`` clause, API producers **may** return a subset of the entity type's properties, omitting computationally expensive properties.
```
darrelmiller: Is this pattern limited to entity types?  Or can we use the same pattern for complex types?  e.g. A function that returns a collection of complex types?

graph/patterns/default-properties.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Issues and considerations
+
+- Do **not** rely on the `ags:Default` schema annotation for default properties functionality, as this is a legacy implementation. Returning default properties **must** be implemented by API producers.
+- Changing a default property to non-default is considered a breaking change.
+- One of the challenges with default properties is informing developers that the response does not contain the full set of properties. To solve for this discovery problem, if the response contains default properties only, then:
+  - the response **must** contain a `@microsoft.graph.tips` instance annotation.
+  - the `@microsoft.graph.tips` instance annotation **must** only be emitted if the client uses "developer mode" via the `Prefer: ms-graph-dev-mode` HTTP request header. It is expected that this header will only be used by client developer and scripting tools like Graph Explorer, the Microsoft Graph Postman collections, and Microsoft Graph PowerShell.
```
darrelmiller, line 27: PowerShell?

graph/patterns/default-properties.md
darrelmiller: If we want to convey Graph tips via standard output in the SDK, we can definitely do that. But I don't see a need for having a code to do that?  I don't think we are going to build a mechanism where developers can choose which tips they do and do not want to have emitted to standard out.

XVincentX: @darrelmiller how would an SDK understand what message to output? Shall I assume the SDK would just print out the message as it is?

graph/patterns/default-properties.md
```diff
@@ -0,0 +1,201 @@
+# Default properties
+
+Microsoft Graph API Design Pattern
+
+*The default properties pattern allows API producers to omit specific properties from the response unless they are explicitly requested using `$select`. The pattern also uses an instance annotation to inform callers that other properties are also available. The same annotation is also use to encourage callers to use `$select`.*
```
OlgaPodo:
```suggestion
*The default properties pattern allows API producers to omit specific properties from the response unless they are explicitly requested using `$select`. The pattern also uses an instance annotation to inform callers that other properties are also available. The same annotation is also use to encourage callers to use `$select`.*
```
can we move these two sentences to the Solution section?

graph/patterns/default-properties.md
OlgaPodo: Agree, may be we should refer to resources instead of entity types?

graph/patterns/default-properties.md
dkershaw10: Sorry - are you asking to add these to the Solution section as well?  If I move them, there'll be nothing here :)

I can't actually see a difference with your suggested change, but did notice I need to change use -> used.

graph/patterns/default-properties.md
dkershaw10: Sure, it could apply to complex types.  We only implemented it so far for entity types.

graph/patterns/default-properties.md
dkershaw10: Do my eye's deceive me, or do I not mention PowerShell at the end of the sentence?  If that's not what your comment is about, please add more detail...

graph/patterns/default-properties.md
OlgaPodo: I assume that all these suggestions are only for developer mode?

dkershaw10: @darrelmiller Not sure if you've seen Vincenzo's response.

graph/patterns/default-properties.md
```diff
@@ -0,0 +1,201 @@
...
+  - the `@microsoft.graph.tips` instance annotation value **must** contain "This request only returns a subset of the resource's properties. Your app will need to use $select to return non-default properties. To find out what other properties are available for this resource see https://learn.microsoft.com/graph/api/resources/{entityTypeName}".
+- Callers must be able to use `$filter` with non-default properties, even though they won't show up by default in the response.
+
+Additionally, for incoming requests targeting an entity type where the caller does not specify a `$select` clause, the API Gateway Service will inject a `@microsoft.graph.tips` instance annotation, informing callers to use $select, when in "developer mode".
+API producers who use [response passthrough](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/391069/Enabling-response-passthrough) must also implement this behavior, supplying the same information as shown in the [examples section below](#calling-an-api-without-using-select).
+
+## Examples
+
```
OlgaPodo, line 35: I wonder if there are any depencdencies between default properties and nullable and default value attributes?

graph/patterns/default-properties.md
OlgaPodo: I'm asking to move it to the Solutions section. I shouldn't have used the Suggestions feature at all; it should have just been a comment.

graph/patterns/default-properties.md
```diff
@@ -0,0 +1,201 @@
...
+## Problem
+
+API producers want to control the set of properties that their entities return by default, when the incoming request does not specify a `$select`. This can be desirable when an entity type has many properties or an API producer needs to add properties that are computationally expensive to return by default.
+
+## Solution
+
+For incoming requests targeting an entity type where the caller does not specify a `$select`` clause, API producers **may** return a subset of the entity type's properties, omitting computationally expensive properties.
+
+To get the non-default properties of an entity type, callers must explicitly request them using `$select`.
```
OlgaPodo: I want to capture my concern here that our current solution is not discoverable at the API design time, and therefore downstream tools like documentation, GE and SDKs can only implement dynamic runtime support. This basically excludes documentation from the picture. It also makes the API behavior less predictable upfront and complicates the developer experience. I suggest adding an additional annotation to specify a default set of properties at the design time.

graph/patterns/default-properties.md
```diff
@@ -0,0 +1,201 @@
+# Default properties
+
+Microsoft Graph API Design Pattern
+
+*The default properties pattern allows API producers to omit specific properties from the response unless they are explicitly requested using `$select`. The pattern also uses an instance annotation to inform callers that other properties are also available. The same annotation is also use to encourage callers to use `$select`.*
+
+## Problem
+
+API producers want to control the set of properties that their entities return by default, when the incoming request does not specify a `$select`. This can be desirable when an entity type has many properties or an API producer needs to add properties that are computationally expensive to return by default.
```
OlgaPodo:
```suggestion
API designers want to control the set of properties that their entities return by default, when the incoming request does not specify a `$select`. This can be desirable when an entity type has many properties or an API producer needs to add properties that are computationally expensive to return by default.
```

graph/patterns/default-properties.md
mikepizzo: To play devils advocate for a moment -- why are we telling developers which properties are returned by default and which they have to $select.  Shouldn't the guidance be to always select exactly the set of properties that you need?  Do we expect them to write code that normally omits $select, but conditionally adds it if they need a non-default property?

graph/patterns/default-properties.md
dkershaw10: Triage: Doc what is implemented and file a work item to change this to a collection of complex types (for improved messages and future extensibility).

graph/patterns/default-properties.md
dkershaw10: Triage: Doc what is implemented and file a work item to extend to complex types

graph/patterns/default-properties.md
dkershaw10: Triage - move to the solutions section

graph/patterns/default-properties.md
dkershaw10: Triage: will leave as is and add a new work item for additional annotations for default properties (CSDL and TypeSpec)

graph/patterns/default-properties.md
dkershaw10: Will take this change.

graph/patterns/default-properties.md
dkershaw10: Filing a work item for this.

graph/patterns/default-properties.md
dkershaw10: Filing a work item for this

graph/patterns/default-properties.md
dkershaw10: Added a work item for this.

PR #501, opened 2023-10-02
Clarified LRO pattern & added VLRO pattern

JeffreyRichter:


mpodwysocki (approved):

azure/Guidelines.md
```diff
@@ -840,32 +841,27 @@ The ability to retry failed requests for which a client never received a respons
...
 Considerations for Service Design for an introduction to the design of long-running operations.
 
-<a href="#lro-response-time" name="lro-response-time">:white_check_mark:</a> **DO** implement an operation as an LRO if the 99th percentile response time is greater than 1s.
+NOTE: Use the _Very Long-Running Operations (VLRO)_ guidelines for operations that should execute asynchronously where clients should not logically block waiting for the operation to complete.
+ 
+<a href="#lro-response-time" name="lro-response-time">:white_check_mark:</a> **DO** implement an operation as an LRO if the 99th percentile response time is greater than 1 second and when the client should poll the operation before making more progress.
```
mikekistler, line 851: Is this a reasonable restatement of the guideline?
```suggestion
<a href="#lro-response-time" name="lro-response-time">:white_check_mark:</a> **DO** implement an operation as an LRO if the 99th percentile response time is greater than 1 second and the client cannot make progress until it determines that the operation is complete.
```

azure/Guidelines.md
```diff
@@ -840,32 +841,27 @@ The ability to retry failed requests for which a client never received a respons
...
 The second pattern applies only in the case of a PUT operation to create a resource that also involves additional long-running processing.
 For guidance on when to use a specific pattern, please refer to [Considerations for Service Design, Long Running Operations](./ConsiderationsForServiceDesign.md#long-running-operations).
-These are described in the following two sections.
+These two patterns are described in the following two sections.
 
 #### POST or DELETE LRO pattern
 
 A POST or DELETE long-running operation accepts a request from the client to initiate the operation processing and returns
-a [status monitor](https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.3) that reports the operation's progress.
+a [status monitor](https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.3) reflecting the operation's progress.
```
mikekistler, line 864: Why is "reflecting" better than "that reports"?

azure/Guidelines.md
```diff
@@ -885,11 +881,11 @@ a [status monitor](https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.3)
 
 <a href="#lro-returns-operation-location" name="lro-returns-operation-location">:ballot_box_with_check:</a> **YOU SHOULD** include an `Operation-Location` header in the response with the absolute URL of the status monitor for the operation.
 
-<a href="#lro-operation-location-includes-api-version" name="lro-operation-location-includes-api-version">:ballot_box_with_check:</a> **YOU SHOULD** include the `api-version` query parameter in the `Operation-Location` header with the same version passed on the initial request if it is required by the get operation on the status monitor.
+<a href="#lro-operation-location-includes-api-version" name="lro-operation-location-includes-api-version">:ballot_box_with_check:</a> **YOU SHOULD** include the `api-version` query parameter in the `Operation-Location` header with the same version passed on the initial request but expect a client to change the api-version value to whatever a new/different client desires it to be.
 
 #### PUT operation with additional long-running processing
 
-For a PUT (create or replace) with additional long-running processing:
+For a PUT (create or replace) requiring long-running processing:
```
mikekistler, line 888: The original wording was intentional. The create or replace happens synchronously -- it does not "require long-running processing". Rather, there is some "additional long-running processing" that must be done _as a result of the create or replace_ that the client may need to track.

azure/Guidelines.md
```diff
@@ -907,45 +903,59 @@ For a PUT (create or replace) with additional long-running processing:
...
 
-<a href="#lro-put-operation-location-includes-api-version" name="lro-put-operation-location-includes-api-version">:ballot_box_with_check:</a> **YOU SHOULD** include the `api-version` query parameter in the `Operation-Location` header with the same version passed on the initial request if it is required by the get operation on the status monitor.
+<a href="#lro-put-operation-location-includes-api-version" name="lro-operation-location-includes-api-version">:ballot_box_with_check:</a> **YOU SHOULD** include the `api-version` query parameter in the `Operation-Location` header with the same version passed on the initial request but expect a client to change the api-version value to whatever a new/different client desires it to be.
 
 #### Obtaining status and results of long-running operations
 
-For all long-running operations, the client will issue a GET on a status monitor resource to obtain the current status of the operation.
+For all long-running operations, a client repeatedly polls the service to GET the operation's status monitor resource to obtain the current status of the operation.
```
mikekistler, line 910: The client does not need to "repeatedly poll the service" to "obtain the current status of the operation". One GET is sufficient to get the current status.

azure/Guidelines.md
```diff
@@ -907,45 +903,59 @@ For a PUT (create or replace) with additional long-running processing:
...
 
 <a href="#lro-status-monitor-retention" name="lro-status-monitor-retention">:white_check_mark:</a> **DO** retain the status monitor resource for some publicly documented period of time (at least 24 hours) after the operation completes.
 
+<a href="#vlro" name="vlro"></a>
+### Very Long-Running Operations (VLROs)
+
+A _very long-running operation (VLRO)_ is an operation that should execute asynchronously. Typically, a VLRO can be identified when 1 or more logical clients need to get the result. For example, if one client initiates the operation and a different browser/client needs the result. An example would be a dashboard or portal that shows all the operations along with their status. In fact, the VLRO pattern offers a way to list all the operations while the LRO pattern does not. Also, the VLRO pattern does not expect a client to make no forward progress due to polling for the long period of time it takes to complete of the operation.
```
mikekistler, line 943:
```suggestion
A _very long-running operation (VLRO)_ is an operation that executes asynchronously and may take an arbitrarily long time to complete. Because of the arbitrarily long execution time, there may be multiple logical clients accessing a VLRO result. For example, if one client initiates the operation and a different browser/client needs the result. An example would be a dashboard or portal that shows all the operations along with their status. In fact, the VLRO pattern offers a way to list all the operations while the LRO pattern does not. Also, the VLRO pattern does not expect a client to make forward progress due to polling for the long period of time it takes to complete of the operation.
```

azure/Guidelines.md
```diff
@@ -907,45 +903,59 @@ For a PUT (create or replace) with additional long-running processing:
...
+
+The VLRO pattern is very simple to implement.
+<a href="#vlro-create-update" name="vlro-create-update">:white_check_mark:</a> **DO** implement a PATCH operation to create/update a VLRO service resource. The request/response schemas are identical. The client's request must specify any properties required to create/update the VLRO operation. The response includes the operations's properties as well as the status monitor properties.
```
mikekistler, line 946: What is a "VLRO service resource". I think this is the first mention in this file. This should be clearly defined above before we refer to it here.

azure/Guidelines.md
```diff
@@ -907,45 +903,59 @@ For a PUT (create or replace) with additional long-running processing:
...
 
 <a href="#lro-status-monitor-retention" name="lro-status-monitor-retention">:white_check_mark:</a> **DO** retain the status monitor resource for some publicly documented period of time (at least 24 hours) after the operation completes.
 
+<a href="#vlro" name="vlro"></a>
+### Very Long-Running Operations (VLROs)
+
+A _very long-running operation (VLRO)_ is an operation that should execute asynchronously. Typically, a VLRO can be identified when 1 or more logical clients need to get the result. For example, if one client initiates the operation and a different browser/client needs the result. An example would be a dashboard or portal that shows all the operations along with their status. In fact, the VLRO pattern offers a way to list all the operations while the LRO pattern does not. Also, the VLRO pattern does not expect a client to make no forward progress due to polling for the long period of time it takes to complete of the operation.
+
+The VLRO pattern is very simple to implement.
```
mikekistler, line 945: I don't think we have a clear definition of the "VLRO pattern" at this point.

azure/Guidelines.md
```diff
@@ -885,11 +881,11 @@ a [status monitor](https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.3)
 
 <a href="#lro-returns-operation-location" name="lro-returns-operation-location">:ballot_box_with_check:</a> **YOU SHOULD** include an `Operation-Location` header in the response with the absolute URL of the status monitor for the operation.
 
-<a href="#lro-operation-location-includes-api-version" name="lro-operation-location-includes-api-version">:ballot_box_with_check:</a> **YOU SHOULD** include the `api-version` query parameter in the `Operation-Location` header with the same version passed on the initial request if it is required by the get operation on the status monitor.
+<a href="#lro-operation-location-includes-api-version" name="lro-operation-location-includes-api-version">:ballot_box_with_check:</a> **YOU SHOULD** include the `api-version` query parameter in the `Operation-Location` header with the same version passed on the initial request but expect a client to change the api-version value to whatever a new/different client desires it to be.
```
mikekistler, line 884: This change isn't necessary because lro-status-monitor-accepts-any-api-version already states that the GET on the status monitor must accept any api-version.

azure/Guidelines.md
```diff
@@ -907,45 +903,59 @@ For a PUT (create or replace) with additional long-running processing:
 
 <a href="#lro-put-returns-operation-location" name="lro-put-returns-operation-location">:ballot_box_with_check:</a> **YOU SHOULD** include an `Operation-Location` header in the response with the absolute URL of the status monitor for the operation.
 
-<a href="#lro-put-operation-location-includes-api-version" name="lro-put-operation-location-includes-api-version">:ballot_box_with_check:</a> **YOU SHOULD** include the `api-version` query parameter in the `Operation-Location` header with the same version passed on the initial request if it is required by the get operation on the status monitor.
+<a href="#lro-put-operation-location-includes-api-version" name="lro-operation-location-includes-api-version">:ballot_box_with_check:</a> **YOU SHOULD** include the `api-version` query parameter in the `Operation-Location` header with the same version passed on the initial request but expect a client to change the api-version value to whatever a new/different client desires it to be.
```
mikekistler, line 906: This change isn't necessary because lro-status-monitor-accepts-any-api-version already states that the GET on the status monitor must accept any api-version.

mikekistler (changes requested):
I like the basic ideas here but I think there are some details we need to work out.

JeffreyRichter:
Replaced by new PR opening shortly

PR #512, opened 2023-12-18
Add graph pattern for common query functions

corranrogue9:
Adding the guidance pattern according to [this](https://microsoft.sharepoint.com/teams/APICouncil/_layouts/15/Doc.aspx?sourcedoc=%7B3569deb7-5543-404f-99b0-70554b7f8778%7D&action=edit&wd=target(ADRs.one%7C0edc1435-5069-4f42-8318-bf31c71c4e37%2FADR-024%20Using%20functions%20as%20query%20parameter%20aliases%7Ce157a2d9-7960-4f08-b0da-8e22ac5248c6%2F)&wdorigin=NavigationUrl&share=IgG33mk1Q1VPQJmwcFVLf4d4AQ7bKVE-D30Of6JCTvaTqbc) ADR. The pattern is to use OData functions as "aliases" for the more complicated or common OData query options for robust APIs that support a wide range of query options.


graph/patterns/query-option-aliasing.md
```diff
@@ -0,0 +1,119 @@
...
+
+This pattern can be employed in almost any circumstance.
+It is also able to be used once an API as shipped and in any order.
+It is ok to ship a data-focused API with robust `$filter`ing support, and *later* ship functions that alias certain specific filters.
+It is likewise fine to ship a handful of functions and then later ship an API that is a collection of data with a more fully-featured set of filters.
+
+## Issues and considerations
+
+The tradeoff with this pattern is ensuring that the functions don't become so numerous that they remove the aliasing benefit.
+It is important to remember that the functions exist to lower dicsoverability, decrease onboarding costs, and prevent client mistakes writing complicated OData queries.
```
mikepizzo: >dicsoverability [](http://example.com/codeflow?start=62&length=15)

```suggestion
discoverability
```

mikepizzo: >lower [](http://example.com/codeflow?start=56&length=5)

```suggestion
increase
```

graph/patterns/query-option-aliasing.md
```diff
@@ -0,0 +1,119 @@
+# Query option aliasing
```
mikepizzo: >Query option aliasing [](http://example.com/codeflow?start=2&length=21)

I'm not sure that the name "Query option aliasing" lands for me, although I understand your intent.

Maybe something like "Exposing functions for common queries"?

mikepizzo (changes requested):
:clock1:

graph/patterns/query-option-aliasing.md
corranrogue9: I didn't remove *all* references to the word alias, but I made some changes, let me know if you think I need to go deeper

graph/patterns/common-query-functions.md
```diff
@@ -0,0 +1,119 @@
...
+## Problem
+
+Some APIs are very data-focused.
+These APIs are generally capable of providing robust `$fitler`ing and `$expand`ing functionality.
+However, the OData query options that need to be used by clients can become complicated and confusing.
+Because of this, API producers have a tendency to avoid supporting query options that require client developers to have OData experience; avoiding these query options results in APIs that are not fully featured and are not externally extensible, limiting what clients can actually accomplish via the APIs.
+
+## Solution
+
+The solution to this problem is to provide support for the query options and to additionally define OData functions that act as aliases for a specific set of query options.
```
mikepizzo, line 16: Although you do allude to it in the Issues and Considerations, I would mention here that the solution is applied when the request being aliased is more complex/less intuitive to construct.

```suggestion
The solution to this problem is to provide support for the query options and to additionally define OData functions that act as aliases for a specific set of query options that represent a common request that may be more complex/less intuitive to construct.
```

graph/patterns/common-query-functions.md
```diff
@@ -0,0 +1,119 @@
+# Exposing functions for common queries
+
+Microsoft Graph API Design Pattern
+
+*Exposing functions for common queries is providing functions with friendly names as an additional mechanism for clients to make certain kinds of requests that involve the use of OData query options.*
```
mikepizzo, line 5: Wording feels a bit awkward.

```suggestion
*Exposing functions for common queries provides friendly names as an additional mechanism for clients to make certain kinds of requests that involve the use of OData query options.*
```

graph/patterns/common-query-functions.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Problem
+
+Some APIs are very data-focused.
+These APIs are generally capable of providing robust `$fitler`ing and `$expand`ing functionality.
+However, the OData query options that need to be used by clients can become complicated and confusing.
```
mikepizzo, line 11:
```suggestion
However, the OData query options used by clients to express certain more complex queries can become complicated and confusing.
```

graph/patterns/common-query-functions.md
```diff
@@ -0,0 +1,119 @@
...
+Because of this, API producers have a tendency to avoid supporting query options that require client developers to have OData experience; avoiding these query options results in APIs that are not fully featured and are not externally extensible, limiting what clients can actually accomplish via the APIs.
```
mikepizzo, line 12: in this context, I would think of it not as experience with OData (most graph folks probably don't think of themselves as OData) but as experience with OData query options (subtle distinction, I know)

```suggestion
Because of this, API producers have a tendency to avoid supporting common query options to express such requests in favor of special purpose functions. However, avoiding these common query options results in APIs that are not fully featured and are not externally extensible, limiting what clients can actually accomplish via the APIs. This can also lead to a proliferation of special purpose functions to express each unique request.
```

graph/patterns/common-query-functions.md
```diff
@@ -0,0 +1,119 @@
...
+The solution to this problem is to provide support for the query options and to additionally define OData functions that act as aliases for a specific set of query options.
+Doing this gives client developers a low barrier-of-entry to using the API (familiarizing themselves with the shape of the data), and it also gives those developers extensiblity in the future to grow beyond the basic cases to write clients that are specific to their own business needs.
+The functions can further provide discoverability by using names for concepts that customers are familiar with, letting them know that those concepts are supported.
+
+## When to use this pattern
+
+This pattern can be employed in almost any circumstance.
+It is also able to be used once an API as shipped and in any order.
+It is ok to ship a data-focused API with robust `$filter`ing support, and *later* ship functions that act in the same way as certain specific filters.
+It is likewise fine to ship a handful of functions and then later ship an API that is a collection of data with a more fully-featured set of filters.
```
mikepizzo, line 25: This makes it sound like it's fine to ship a special purpose function because, if in the future they decide they want something more flexible, they could add it. I'm afraid that would lead to workloads using this guidance to say "I have a bunch of functions today, and might add more in the future, cause the guidance says that I can always add $filter later."

I would think that we would want to encourage workloads to move to the more general solution and not rely on exposing everything through custom functions.

```suggestion
While it is possible to start with one or two custom functions, you should have a plan to support a more fully-featured set of query options as the utility of your API will be significantly limited in the interim.
```

PR #519, opened 2024-01-23
Property validity using enums

corranrogue9:
Porting [this](https://dev.azure.com/msazure/One/_git/AD-AggregatorService-Wiki/pullrequest/5809345?_a=files&path=/API/Design/Modeling-patterns/Value-Based-Property-Validity.md) PR from the old repo.


graph/patterns/property-validity-using-enums.md
```diff
@@ -0,0 +1,293 @@
+# Property validity using enums
+
+Microsoft Graph API Design Pattern
+
+*There are often cases where a type has a property which denotes a kind of role that the type can fulfill. This property can evolve over time to a point where the way the type that fulfills that role needs to be configured in some way. This pattern allows configuring such properties in a forward compatible way.*
+
```
OlgaPodo, line 6: When we talk about aPI design pattern we think in terms of API resources and representation for a new customer. It is not clear what application or business problem that patter is trying to address. Can we be more specific? Good examples are  https://github.com/microsoft/api-guidelines/blob/vNext/graph/patterns/navigation-property.md, https://github.com/microsoft/api-guidelines/blob/vNext/graph/patterns/viewpoint.md

OlgaPodo: It seems that this pattern is beyond  a role use case and is about properties dependencies where one property is an enumeration.

PR #520, opened 2024-01-24
Add guidance to use composable functions instead of filter-as-segment feature

corranrogue9:
Discussed previously, this PR adds the guidance to avoid the filter-as-segment OData feature and instead use a composable function which accepts an expression parameter for those scenarios where operations on filtered collections are desirable.

### DIFF
--- a/graph/articles/filter-as-segment.md
+++ b/graph/articles/filter-as-segment.md
@@ -58,4 +58,4 @@ POST /identityProtection/riskyUsers/filter(expression='id IN (''{userId1}'',''{u
 
 NOTE: the `'` literal in the filter expression must be escaped with `''`
 
-An example implementation of a filter function can be found [here](TODO).
+An example implementation of a filter function using OData WebApi can be found [here](TODO).
--- a/graph/articles/filter-as-segment.md
+++ b/graph/articles/filter-as-segment.md
@@ -1,3 +1,51 @@
 # Filter as segment
 
-There is an OData feature which allows 
+There is an [OData feature](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_AddressingaSubsetofaCollection) which allows having a `$filter` in a URL segment. 
+This feature is useful whenever there are operations on a collection and the client wants to perform those operations on a *subset* of the collection. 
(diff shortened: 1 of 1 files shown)


graph/articles/filter-as-segment.md
```diff
@@ -0,0 +1,61 @@
...
+
+and clients could call
+
+```http
+POST /identityProtection/riskyUsers/$filter=@f/dismiss?@f=id IN ('{userId1}','{userId2}',...)
+```
+
+Doing this is beneficial due to the robust nature of OData filter expressions: clients will be able to dismiss risky users based on any supported filter without the service team needing to implement a new `dismiss` overload that filters based on the new criteria.
+However, there are some concerns about the discoverability of using the filter-as-segment feature, as well as the support of [parameter aliasing](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_ParameterAliases) that's required.
+As a result, functions should be introduced that act in the same way as the filter-as-segment:
```
OlgaPodo, line 43: I think naming a specific function as "filter" is very confusing for users who expect a standard $filter behavior. I believe it does not follow our naming guidelines without having a product-specific prefix. Unless we will treat this function similarly to delta?

graph/articles/filter-as-segment.md
```diff
@@ -0,0 +1,61 @@
+# Filter as segment
+
```
OlgaPodo, line 2: Why is this positioned as an article? i think it make sense to create a pattern and clearly specify when this solution is a good one.

graph/articles/filter-as-segment.md
mikepizzo: namespace qualification of functions is optional in msgraph.

OlgaPodo: customFilter or filterFunction

mikepizzo (approved):

graph/articles/filter-as-segment.md
dkershaw10: This should probably just be included as another scenario of #512 
We might rename 512 slightly to reflect this, but doesn't seem like this should be a separate pattern, when the pattern is "replacing a complex filter with a function".

PR #517, opened 2024-01-17
Updated LRO guidelines

JeffreyRichter:

PR #521, opened 2024-01-28
Add guidelines on returning string offsets & lengths

mikekistler:
This PR splits out the update for string offset and length from #517. I also reworked things a bit by moving the explanatory content over to ConsiderationsForServiceDesign.

It looks like my editor also trimmed some trailing whitespace from otherwise unchanged lines.

### DIFF
--- a/azure/ConsiderationsForServiceDesign.md
+++ b/azure/ConsiderationsForServiceDesign.md
@@ -6,8 +6,9 @@
 
 | Date        | Notes                                                          |
 | ----------- | -------------------------------------------------------------- |
+| 2024-Jan-17 | Added guidelines on returning string offsets & lengths         |
 | 2022-Jul-15 | Update guidance on long-running operations                     |
-| 2022-Feb-01 | Updated error guidance                                        |
+| 2022-Feb-01 | Updated error guidance                                         |
 | 2021-Sep-11 | Add long-running operations guidance                           |
 | 2021-Aug-06 | Updated Azure REST Guidelines per Azure API Stewardship Board. |
 
@@ -517,7 +518,7 @@ By computing and returning ETags for your resources, you enable clients to avoid
 
 ## Returning String Offsets & Lengths (Substrings)
 
-Some Azure services return substring offset & length values within a string. For example, the offset & length within a string to a name, email address, or phone #.
+Some Azure services return substring offset & length values within a string. For example, the offset & length within a string to a name, email address, or phone number.
 When a service response includes a string, the client's programming language deserializes that string into that language's internal string encoding. Below are the possible encodings and examples of languages that use each encoding:
 
 | Encoding    | Example languages |
@@ -526,11 +527,11 @@ When a service response includes a string, the client's programming language des
 | UTF-16 | JavaScript, Java, C# |
 | CodePoint (UTF-32) | Python |
 
-Because the service doesn't know what language a client is written in and what string encoding that language uses, the service can't return UTF-agnostic offset and length values that the client can use to index within the string. To address this, the service response must include offset & length values for all 3 possible encodings and then the client code must select the encoding it required by its language's internal string encoding.
+Because the service doesn't know in what language a client is written and what string encoding that language uses, the service can't return UTF-agnostic offset and length values that the client can use to index within the string. To address this, the service response must include offset & length values for all 3 possible encodings and then the client code must select the encoding required by its language's internal string encoding.
 
 For example, if a service response needed to identify offset & length values for "name" and "email" substrings, the JSON response would look like this:
 
-```
+```json
 {
   (... other properties not shown...)
   "fullString": "(...some string containing a name and an email address...)",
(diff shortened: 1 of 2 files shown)


JeffreyRichter (approved):

MushMal (approved):

azure/ConsiderationsForServiceDesign.md
```diff
@@ -515,6 +515,61 @@ For example, the client can specify an `If-Match` header with the last ETag valu
...
+
+```
+{
+  (... other properties not shown...)
+  "fullString": "(...some string containing a name and an email address...)",
+  "name": {
+    "offset": {
+      "utf8": 12,
+      "utf16": 10,
+      "codePoint":  4
```
weidongxu-microsoft: nit, we seems got 2 spaces here `"codePoint":  4`

azure/ConsiderationsForServiceDesign.md
```diff
@@ -515,6 +515,61 @@ For example, the client can specify an `If-Match` header with the last ETag valu
 The service processes the update only if the ETag value in the header matches the ETag of the current resource on the server.
 By computing and returning ETags for your resources, you enable clients to avoid using a strategy where the "last write always wins."
 
+## Returning String Offsets & Lengths (Substrings)
+
+Some Azure services return substring offset & length values within a string. For example, the offset & length within a string to a name, email address, or phone #.
```
weidongxu-microsoft: nit `phone #` seems too informal? Just `phone number`?

azure/ConsiderationsForServiceDesign.md
```diff
@@ -515,6 +515,61 @@ For example, the client can specify an `If-Match` header with the last ETag valu
...
+When a service response includes a string, the client's programming language deserializes that string into that language's internal string encoding. Below are the possible encodings and examples of languages that use each encoding:
+
+| Encoding    | Example languages |
+| -------- | ------- |
+| UTF-8 | Go, Rust, Ruby, PHP |
+| UTF-16 | JavaScript, Java, C# |
+| CodePoint (UTF-32) | Python |
+
+Because the service doesn't know what language a client is written in and what string encoding that language uses, the service can't return UTF-agnostic offset and length values that the client can use to index within the string. To address this, the service response must include offset & length values for all 3 possible encodings and then the client code must select the encoding it required by its language's internal string encoding.
```
heaths: Grammar nit: 
```suggestion
Because the service doesn't know in what language a client is written and what string encoding that language uses, the service can't return UTF-agnostic offset and length values that the client can use to index within the string. To address this, the service response must include offset & length values for all 3 possible encodings and then the client code must select the encoding required by its language's internal string encoding.
```

azure/ConsiderationsForServiceDesign.md
```diff
@@ -515,6 +515,61 @@ For example, the client can specify an `If-Match` header with the last ETag valu
...
+```
+
+Then, the Go developer, for example, would get the substring containing the name using code like this:
+
+```
+   var response := client.SomeMethodReturningJSONShownAbove(...)
+   name := response.fullString[ response.name.offset.utf8 : response.name.offset.utf8 + response.name.length.utf8]
+```
+
+The service must calculate the offset & length for all 3 encodings and return them because clients find it difficult working with Unicode encodings and how to convert from one encoding to another. In other words, we do this to simplify client development and ensure customer success when isolating a substring.
```
heaths, line 572: Should we also mention that it makes pass-through requests easier as well? That was the thing that really won me over. I think the same was true for @JeffreyRichter, IIRC.

azure/Guidelines.md
```diff
@@ -1098,6 +1099,14 @@ While it may be tempting to use a revision/version number for the resource as th
 
 <a href="#condreq-etag-depends-on-encoding" name="condreq-etag-depends-on-encoding">:white_check_mark:</a> **DO**, when supporting multiple representations (e.g. Content-Encodings) for the same resource, generate different ETag values for the different representations.
 
+<a href="#substrings" name="substrings"></a>
+### Returning String Offsets & Lengths (Substrings)
+
+All string values in JSON are inherently Unicode and UTF-8 encoded, but clients written in a high-level programming language must work with strings in that language's string encoding, which may be UTF-8, UTF-16, or CodePoints (UTF-32).
+When a service response includes a string offset or length value, it should specify these values in all 3 encodings to simplify client development and ensure customer success when isolating a substring.
+
+<a href="#substrings-return-value-for-each-encoding" name="substrings-return-value-for-each-encoding">:white_check_mark:</a> **DO** include all 3 encodings (UTF-8, UTF-16, and CodePoint) for every string offset or length value in a service response.
```
heaths, line 1108: I think we should document here in this doc the exact format we want e.g., `{"utf8": 2, "utf16": 1, "codePoint":1}`. We document formats for LROs, pageables, and errors. How you expanded on that in "Considerations" is perfect, but you should also link to that section e.g.,
````suggestion
<a href="#substrings-return-value-for-each-encoding" name="substrings-return-value-for-each-encoding">:white_check_mark:</a> **DO** include all 3 encodings (UTF-8, UTF-16, and CodePoint) for every string offset or length value in a service response using the schema below. See [considerations](ConsiderationsForServiceDesign.md#{actual-stub-here}) for more information.

```json
{
  "length": {
    "utf8": 2,
    "utf16": 1,
    "codePoint": 1
  }
}
```
````

heaths (approved):
A few suggestions, but otherwise LGTM.

PR #522, opened 2024-01-31
Update dictionary guidance to have the correct workload CSDL to implement a dictionary type

corranrogue9:
The current guidance does not pass the schema validation. I have updated it to reflect what actually passes schema validation (except for a single suppression that is required). You can find a repro of this guidance [here](https://msazure.visualstudio.com/One/_git/AD-AggregatorService-Workloads/pullrequest/9433485).

### DIFF
--- a/graph/patterns/dictionary.md
+++ b/graph/patterns/dictionary.md
@@ -37,7 +37,9 @@ For more information, see the [OData reference](https://github.com/oasis-tcs/oda
 
 ## Examples
 
-### Declaring a string dictionary
+### String dictionary
+
+#### CSDL declaration
 The following example demonstrates defining a dictionary that can contain string values.
 
 ```xml
@@ -60,7 +62,7 @@ The following example demonstrates defining a dictionary that can contain string
 Please note that schema validation will fail due to the casing of `Dictionary`.
 This warning should be suppressed.
 
-### Defining a dictionary property
+#### Defining a dictionary property
 The following example shows defining a dictionary property, "userTags", on the item entity type.
 
 ```xml
@@ -70,7 +72,7 @@ The following example shows defining a dictionary property, "userTags", on the i
 </EntityType>
 ```
 
-### Reading a dictionary
+#### Reading a dictionary
 Dictionaries are represented in JSON payloads as a JSON object, where the property names are comprised of the keys and their values are the corresponding key values. 
 
 The following example shows reading an item with a dictionary property named "userTags":
@@ -90,7 +92,7 @@ Response:
 }
 ```
 
-### Setting a dictionary value
+#### Setting a dictionary value
 The following example shows setting a dictionary value.  If "hairColor" already exists, it is updated, otherwise it is added.
 
 ```http
@@ -102,7 +104,7 @@ PATCH /item/userTags
 }
 ```
 
-### Deleting a dictionary value
+#### Deleting a dictionary value
 A dictionary value can be removed by setting the value to null.
 ```http
 PATCH /item/userTags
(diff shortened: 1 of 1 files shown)


graph/patterns/dictionary.md
```diff
@@ -41,15 +41,23 @@ For more information, see the [OData reference](https://github.com/oasis-tcs/oda
...
 
 ```xml
-<ComplexType Name="stringDictionary" BaseType="graph.Dictionary">
-  <Annotation Term="Org.OData.Validation.V1.OpenPropertyTypeConstraint">
-    <Collection>
-      <String>Edm.String</String>
-    </Collection>
-  </Annotation>
-</ComplexType>
+<Schema Namespace="WorkloadNamespace">
```
dkershaw10: I'm good with the change, but kind of feel like it needs a minor restructure.
Under Examples, we should have 2 sub-sections - String dictionary and Complex type dictionary and then put the current relevant sections under those two.

dkershaw10 (changes requested):
Minor change to structure.  Rest looks good to me.

graph/patterns/dictionary.md
gareth-a-jones: I think Dictionary needs to be defined in the core Microsoft.Graph namespace otherwise we will have multiple dictionary base types floating around.

gareth-a-jones (changes requested):
Note about namespaces

graph/patterns/dictionary.md
corranrogue9: That was the intent, but I think you're right this isn't explicit enough. I've updated to declare a `microsoft.graph` namespace (and also confirms that the onboarding pipeline allows doing this...)

corranrogue9: I created a subsection for string vs complex type dictionary

corranrogue9:
@garethj-msft @dkershaw10 I think I've addressed the feedback if you can take another look

graph/patterns/dictionary.md
```diff
@@ -103,34 +115,43 @@ PATCH /item/userTags
...
+    ...
+    <Property Name="assignedRoles" Type="WorkloadNamespace.assignedRoleGroupDictionary">
+  </EntityType>
+
+  <ComplexType Name="roleSettings">
+    <Property Name ="domain" Type="Edm.String" Nullable="false" />
+  </ComplexType>
+
+  <ComplexType Name="assignedRoleGroupDictionary" BaseType="microsoft.graph.Dictionary">
+    <!-- Note: Strongly-typed dictionary
```
mikepizzo: not sure why this comment is broken across three rows.
```suggestion
    <!-- Note: Strongly-typed dictionary of roleSettings keyed by name of roleGroup. -->
```

graph/patterns/dictionary.md
```diff
@@ -103,34 +115,43 @@ PATCH /item/userTags
...
-    </Collection>
-  </Annotation>
-</ComplexType>
+<Schema Namespace="microsoft.graph"> <!--NOTE: the namespace that declares the Dictionary complex type *must* be microsoft.graph-->
+  <ComplexType Name="Dictionary" OpenType="true">
+    <Annotation Term="Core.Description" String="A dictionary of name-value pairs. Names must be valid property names, values may be restricted to a list of types via an annotation with term `Validation.OpenPropertyTypeConstraint`." />
+  </ComplexType>
+</Schema>
+<Schema Namespace="WorkloadNamespace">
+  <EntityType Name="principal">
```
mikepizzo, line 132: nit: consider moving this to *after* the definition of assignedRoleGroupDictionary, so that the reader sees the progression (the complex type, the dictionary of complex type, and then the use of the dictionary)

mikepizzo (approved):
A few minor suggestions, but nothing blocking.  Thanks for taking the time to fix these examples -- it is a continual source of delay and confusion for both authors and reviewers.

OlgaPodo (approved):

PR #537, opened 2024-03-21
Add guidance for types which should not have new structural properties

tylercleveland2:
Adding guidance that there should not be new structural properties added to `user`, `group`, `device`.

### DIFF
--- a/graph/GuidelinesGraph.md
+++ b/graph/GuidelinesGraph.md
@@ -336,7 +336,7 @@ For a complete mapping of error codes to HTTP statuses, see
 ### Limitations on core types
 
 The types `user`, `group`, and `device` should not have any new structural property(s) added, without compelling justification.
-Instead, model the concept represented in those property(s) as a new entity, and do one of the following:
+Instead, model the concept represented in those property(s) as a new entity and do one of the following:
 1. Add navigation from `user`, `group`, or `device` to the new entity.
 2. Add a navigation from the new entity to `user`, `group` or `device`.
 
--- a/graph/GuidelinesGraph.md
+++ b/graph/GuidelinesGraph.md
@@ -337,8 +337,8 @@ For a complete mapping of error codes to HTTP statuses, see
 
 The types `user`, `group`, and `device` should not have any new structural property(s) added, without compelling justification.
 Instead, model the concept represented in those property(s) as a new entity, and do one of the following:
-1. Add navigation to the new entity from `user`, `group`, or `device`.
-2. Contain the new entity elsewhere, and add a navigation from the new entity to `user`, `group` or `device`.
+1. Add navigation from `user`, `group`, or `device` to the new entity.
+2. Add a navigation from the new entity to `user`, `group` or `device`.
 
 More details and examples are available in [Core types](./articles/coreTypes.md).
 
--- a/graph/GuidelinesGraph.md
+++ b/graph/GuidelinesGraph.md
@@ -14,6 +14,7 @@ Table of contents
     - [Query support](#query-support)
     - [Behavior modeling](#behavior-modeling)
     - [Error handling](#error-handling)
+    - [Limitations on core types](#limitations-on-core-types)
   - [External standards](#external-standards)
   - [API contract and nonbackward compatible changes](#api-contract-and-nonbackward-compatible-changes)
     - [Versioning and deprecation](#versioning-and-deprecation)
@@ -331,6 +332,15 @@ For a complete mapping of error codes to HTTP statuses, see
 [rfc7231 (ietf.org)](https://datatracker.ietf.org/doc/html/rfc7231#section-6).
 
 <a name="api-contract-and-non-backward-compatible-changes"></a>
+
+### Limitations on core types
+
+The types `user`, `group`, and `device` should not have any new structural property(s) added, without compelling justification.
+Instead, model the concept represented in those property(s) as a new entity, and do one of the following:
+1. Add navigation to the new entity from `user`, `group`, or `device`.
+2. Contain the new entity elsewhere, and add a navigation from the new entity to `user`, `group` or `device`.
+
+More details and examples are available in [Core types](./articles/coreTypes.md).
 
 ## External standards
 
--- a/graph/articles/coreTypes.md
+++ b/graph/articles/coreTypes.md
@@ -16,7 +16,7 @@ The following types are identified as core types, and will require strong justif
 
 ## Alternatives to Adding Structural Properties
 
-Instead of adding a structural property to the existing type (`user`, `group` or `device`), create a new type that models the information captured in the proposed structural property(s).
+Instead of adding a structural property to the existing core type (`user`, `group` or `device`), create a new type that models the information captured in the proposed structural property(s).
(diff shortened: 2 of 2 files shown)


graph/articles/coreTypes.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+Add a navigation from user to the new type:
+```xml
+<EntityType name="user">
+    <NavigationProperty Name="bankAccountInformation" Type="bankAccountInformation" />
+</EntityType>
+```
+
+#### Contain the new type in an entity set elsewhere, and add a navigation property to the existing type on the new type.
```
corranrogue9:
```suggestion
#### Option 3: Contain the new type in an entity set elsewhere, and add a navigation property to the existing type on the new type.
```

graph/articles/coreTypes.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+Add a contained navigation from user to the new entity type:
+```xml
+<EntityType name="user">
+    <NavigationProperty Name="bankAccountInformation" Type="bankAccountInformation" ContainsTarget="true"/>
+</EntityType>
+```
+
+#### Contain the new type in an entity set elsewhere, and add a navigation property to the new type on the existing type.
```
corranrogue9:
```suggestion
#### Option 2: Contain the new type in an entity set elsewhere, and add a navigation property to the new type on the existing type.
```

graph/articles/coreTypes.md
```diff
@@ -0,0 +1,97 @@
...
+    <Property Name="accountNumber" Type="Edm.string"/>
+    <Property Name="routingNumber" Type="Edm.string"/>
+</EntityType>
+```
+
+### Do:
+
+Do one of the following:
+
+#### Add a navigation property on the existing type to the new type, containing the new type.
```
corranrogue9:
```suggestion
#### Option 1: Add a navigation property on the existing type to the new type, containing the new type.
```

graph/articles/coreTypes.md
```diff
@@ -0,0 +1,97 @@
+# Core Types
+
+## Overview
+
+Types exist in graph which are highly-connected/central to the graph ecosystem. Often, these types are the position of containing many structural properties relevant to other APIs, because they are connected to many entities in the graph.
```
corranrogue9:
```suggestion
Types exist in graph which are highly-connected/central to the graph ecosystem. Often, these types are in the position of containing many structural properties relevant to other APIs, because they are connected to many entities in the graph.
```

corranrogue9 (approved):

corranrogue9:
@microsoft/graphguidelinesapprovers would someone else like to take a look at this guidance?

tylercleveland2:
Pinging @microsoft/graphguidelinesapprovers for a review

graph/GuidelinesGraph.md
```diff
@@ -332,6 +333,15 @@ For a complete mapping of error codes to HTTP statuses, see
 
 <a name="api-contract-and-non-backward-compatible-changes"></a>
 
+### Limitations on core types
```
dkershaw10, line 336: Can we add this guidance for applications and servicePrincipals too?  In my mind those are core types.

We actually now have a very real example of why this is important for downstream tooling.  Adding a structural property that is handled by a different workload (from the core workload) is really difficult to model with Bicep.  If instead it's added as a new entity + nav property, this fits cleanly with Bicep (and ARM's) nested child resources.

graph/articles/coreTypes.md
```diff
@@ -0,0 +1,97 @@
+# Core Types
+
+## Overview
+
+Types exist in graph which are highly-connected/central to the graph ecosystem. Often, these types are in the position of containing many structural properties relevant to other APIs, because they are connected to many entities in the graph.
```
dkershaw10:
```suggestion
Types exist in Microsoft Graph which are highly-connected/central to the graph ecosystem. Often, these types are in the position of containing many structural properties relevant to other APIs, because they are connected to many entities in the graph.
```
Please make this change throughout the article.

graph/articles/coreTypes.md
```diff
@@ -0,0 +1,97 @@
...
+
+Structural properties should be only added to these core types when they are properties of the entity itself and strictly not for the purpose of convenience due to the entity's position in the graph.
```
dkershaw10: Maybe another way to say this is that properties that can be added are intrinsic to the entity.

graph/articles/coreTypes.md
```diff
@@ -0,0 +1,97 @@
...
+
+The following types are identified as core types, and will require strong justification to allow new structural properties to be added in all cases.
+
+- ```user```
+- ```group```
+- ```device```
+
+## Alternatives to Adding Structural Properties
+
+Instead of adding a structural property to the existing type (`user`, `group` or `device`), create a new type that models the information captured in the proposed structural property(s).
```
dkershaw10: So we are kind of mixing core types and existing types here, which starts to get confusing.  We should stick to one.
Suggest changing existing -> core (or maybe "existing core type").

graph/articles/coreTypes.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Do:
+
+Do one of the following:
+
+#### Option 1: Add a navigation property on the existing type to the new type, containing the new type.
+
+Define the new entity type:
+```xml
+<EntityType name="bankAccountInformation">
```
dkershaw10: Suggest changing the name to `bankAccountDetail`.  It has a better plural form than "informations" which I don't think is a real word.

graph/articles/coreTypes.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+Add a contained navigation from user to the new entity type:
+```xml
+<EntityType name="user">
+    <NavigationProperty Name="bankAccountInformation" Type="bankAccountInformation" ContainsTarget="true"/>
+</EntityType>
+```
+
+#### Option 2: Contain the new type in an entity set elsewhere, and add a navigation property to the new type on the existing type.
```
dkershaw10: Are we providing any guidance on when it's appropriate to use containment vs a standard nav property?  Maybe link to an nav prop/relationships guidance to help the reader decide which is more appropriate for them.

dkershaw10 (changes requested):
Looks good, but added some suggestions.  You might want to discuss with other folks whether apps and SPs should be classed as core types (IMHO they are, but you should check).

graph/articles/coreTypes.md
tylercleveland2: Added link in options section, as well as "alternatives" description above

graph/GuidelinesGraph.md
corranrogue9: > Adding a structural property that is handled by a different workload (from the core workload)

FYI, I think we have a *different* ADR that says we should stop using composite entities.

I'm ok adding `applications` and `servicePrincipals`, but I don't think those have been brought up with the API council. Do we want to take this PR as-is and do a follow-up to add apps and sps after discussion with the council?

tylercleveland2: I am inclined to take Garrett's suggestion. If you are fine with this, I will merge as-is @dkershaw10

PR #546, opened 2024-05-27
Clarify the naming of EnumTypes and give guidance for the use of flags enums and non-flags enums simultaneously

corranrogue9:


dkershaw10 (approved):
LGTM

graph/patterns/enums.md
```diff
@@ -127,3 +129,5 @@ However, [Flagged Enums](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/
 With such enum, customers can select multiple values in a single field:
 
 `displayMethod = tip | alert`
+
+In cases where two properties want to use the same *conceptual* `EnumType`, but one property is a collection while the other is single-values, the model should have *two* duplicate `EnumType`s, marking one as a flags enum and making its name the plural form of the non-flags enum.
```
dkershaw10:
```suggestion
In cases where two properties want to use the same *conceptual* `EnumType`, but one property is a collection while the other is single-values, the model should define *two* separate `EnumType`s, one being a non-flags enum with a singular name and the other marked as a flags enum with its name being the plural form of the non-flags enum.
```

PR #547, opened 2024-06-20
fix namespace guidance

mikepizzo:
There has been confusion between onboarding and workload owners on exactly what is/is not allowed and required with respect to multiple namespaces (not surprising since we've circled several times on this).  This PR reflects the latest decisions regarding namespace use within MS Graph.  Specifically:
 - custom namespaces require aliases, as per [ADR-030](https://microsoft.sharepoint.com/teams/APICouncil/_layouts/OneNote.aspx?id=%2Fteams%2FAPICouncil%2FSiteAssets%2FAPI%20Council%20Notebook&wd=target%28ADRs.one%7CCC35B798-B670-4554-9FCF-1682729EFA23%2FADR-030%20Namespace%20Aliases%7C363534F8-EFBD-4201-9F11-DB02AD5B7FAA%2F%29) 
 - namespace can support multiple levels below microsoft.graph (as per [ADR-030])
 - meaning of a type name should be unique across namespaces; name need not be unique
 - any change to a custom namespace name is breaking

### DIFF
--- a/graph/patterns/namespace.md
+++ b/graph/patterns/namespace.md
@@ -16,31 +16,39 @@ API designers can use the namespace attribute of the CSDL schema to declare a
 namespace and logically organize related API entities in the Microsoft Graph metadata.
 
 ```XML
-<Schema Namespace="microsoft.graph.{namespace}">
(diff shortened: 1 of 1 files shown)


dkershaw10 (approved):
LGTM

PR #551, opened 2024-07-05
add guidance for transitioning from a structural collection to a navigation collection

corranrogue9:


graph/articles/collections.md
```diff
@@ -316,3 +316,148 @@ Content-Type: application/json
...
++ <Key>
++   <PropertyRef Name="prop1" />
++ </Key>
++ <Property Name="prop1" Type="Edm.String" />
++ <Property Name="prop2" Type="Edm.String" />
++</EntityType>
+```
+Clients will now be able to refer to individual `bar`s using `prop1` as a key, and they can now remove those `bar`s using `DELETE` requests:
+```http
+DELETE /foos/{fooId}/bars/{some_prop1}
```
mikepizzo: This should be `barsAsEntities`, right?
``` suggestion
DELETE /foos/{fooId}/barsAsEntities
```

graph/articles/collections.md
```diff
@@ -316,3 +316,148 @@ Content-Type: application/json
...
+
+The model can be updated to have two collections side-by-side:
+```diff
+<EntityType Name="foo">
+  <Key>
+    <PropertyRef Name="id" />
+  </Key>
+  <Property Name="id" Type="Edm.String" Nullable="false" />
+  <Property Name="bars" Type="Collection(self.bar)" />
++ <NavigationProperty Name="barsAsEntities" Type="Collection(self.barAsEntity)" ContainsTarget="true" />
```
mikepizzo: We wouldn't really want them to name `barsAsEntities`, would we?  I feel like `Entities` isn't a concept the api consumer necessarily needs to understand.

graph/articles/collections.md
```diff
@@ -316,3 +316,148 @@ Content-Type: application/json
...
+Clients will now be able to refer to individual `bar`s using `prop1` as a key, and they can now remove those `bar`s using `DELETE` requests:
+```http
+DELETE /foos/{fooId}/bars/{some_prop1}
+```
+```http
+HTTP/1.1 204 No Content
+```
+The expectation is that `bars` and `barsAsEntities` are treated as two "views" into the same data.
+To meet this expectation, workloads must:
+1. Keep the properties consistent between `bar` and `barAsEntity`.
```
mikepizzo: Do we really want to keep maintaining `bar`, or would it be better to deprecate `bar` and only add new properties to `barAsEntity`?

graph/articles/collections.md
```diff
@@ -316,3 +316,148 @@ Content-Type: application/json
...
+### 11.1 Side-by-side collection properties
+
+The model can be updated to have two collections side-by-side:
+```diff
+<EntityType Name="foo">
+  <Key>
+    <PropertyRef Name="id" />
+  </Key>
+  <Property Name="id" Type="Edm.String" Nullable="false" />
+  <Property Name="bars" Type="Collection(self.bar)" />
```
mikepizzo: In this case, I would suggest deprecating `bars` in favor of `barsAsEntities`.

graph/articles/collections.md
```diff
@@ -316,3 +316,148 @@ Content-Type: application/json
...
+      "prop2": "another value"
+    },
+    ...
+  ]
+  "bars@odata.navigationLink": "/foos('{fooId}')/bars"
+}
+```
+
+2. The default behavior for structural collections is to include them in the response payload for their containing entity. If this was the behavior of `foo` before, it must be preserved by **auto-expanding** the `bars` property now that it is a navigation property (because the default behavior for navigation properties is to **not** expand them).
+3. Structural collections are updated using `PATCH` requests to replace the entire contents of the collection. The new navigation property must preserve this behavior.
```
mikepizzo: to be clear, the patch request is to the containing entity (not the collection), and this is required only if the original structural property supported updating.
```suggestion
3. Structural collections can be updated using a `PATCH` request to the containing entity to replace the entire contents of the collection. If the service supported such updates to the structural collection, then updates to the new navigation property must preserve this behavior.
```

graph/articles/collections.md
```diff
@@ -316,3 +316,148 @@ Content-Type: application/json
    "value": []
 }
 ```
+
+## 11. Collections of structural types (complex types or primitive types)
+
+Collections of entity types are generally preferable to collections of structual types because collections of structural types must be updated as a single unit, meaning that they are overwritten entirely by new contents, rather than be updated relative to the existing contents. 
+Sometimes, structural collection properties are added to a type and then scenarios are discovered later that require a collection of entity types.
+Take the following model with an entity type `foo` that has a collection of `bar`s:
```
OlgaPodo: These updates aim to address new issues in real-life modeling scenario. Can we use the actual Graph examples to inform our guidelines?

graph/articles/collections.md
```diff
@@ -316,3 +316,148 @@ Content-Type: application/json
...
+```
+Clients will now be able to refer to individual `bar`s using `prop1` as a key, and they can now remove those `bar`s using `DELETE` requests:
+```http
+DELETE /foos/{fooId}/bars/{some_prop1}
+```
+```http
+HTTP/1.1 204 No Content
+```
+The expectation is that `bars` and `barsAsEntities` are treated as two "views" into the same data.
+To meet this expectation, workloads must:
```
OlgaPodo, line 391: I didn't understand why we would enforce ambiguity of updating two places indefinitely?  I think that we need to deprecate wrong modeling choice and have a single option to treat so called 'bars'.

graph/articles/collections.md
```diff
@@ -316,3 +316,148 @@ Content-Type: application/json
...
+200 OK
+{
+  "id": "{fooId}",
+  "bars": [
+    {
+      "prop1": "some value",
+      "prop2": "another value"
+    },
+    ...
+  ]
```
corranrogue9:
```suggestion
  ],
```

graph/articles/collections.md
```diff
@@ -316,3 +316,148 @@ Content-Type: application/json
...
+
+The model can be updated to simply switch the complex type for an entity type:
+```diff
+<EntityType Name="foo">
+  <Key>
+    <PropertyRef Name="id" />
+  </Key>
+  <Property Name="id" Type="Edm.String" Nullable="false" />
+- <Property Name="bars" Type="Collection(self.bar)" />
++ <NavigationProperty Name="bars" Type="Collection(self.bar)" ContainsTarget="true" />
```
corranrogue9: how does sdp look for workloads on this?

graph/articles/collections.md
corranrogue9: Thanks, I found a suitable example from graph

graph/articles/collections.md
corranrogue9: I'm concerned about adding too many intervening clauses and conditionals. Workload teams need to be able to understand what is expected of them in a straightforward manner. 

I don't think anything in my original statements is factually incorrect. I would love to correct anything that's wrong, but we need to do so in a way that doesn't make this sound "optional".

graph/articles/collections.md
corranrogue9: I'm open to other naming conventions, but I do think a naming convention needs to be established for this. If we take your other suggestions and drive workloads to deprecate the old property, then this new property is temporary and can be moved to use the old name if desired. I don't personally have an issue with this name because a client that is wanting to update this collection at least needs to know enough about OData to understand that this is the property that they need to use (instead of the old property). I'm normally opposed to hungarian notation, but in this instance, the difference between the two properties is *literally* the type, and so the naming needs to reflect that.

graph/articles/collections.md
corranrogue9: At the very least, this section needs to be honored by workloads for the period that both properties exist. I think we are agreed on that?

I personally don't feel strongly about driving workloads to deprecate the old property. I think that their own supportability overhead will be a forcing function for them, and if they don't have significant overhead, I don't see a problem with having both properties. 

Maybe I should just add a sentence that says that the old property should be marked deprecated (and leave out whether it ever gets removed)?

graph/articles/collections.md
corranrogue9: I think the [other comment](https://github.com/microsoft/api-guidelines/pull/551#discussion_r1669269517) has more context, but I'll leave this open for now pending that outcome.

graph/articles/collections.md
corranrogue9: I'm happy to add a sentence that the old property should be marked deprecated. I do think these expectations need to be honored at least while both properties exist side-by-side. I personally don't think we need to drive workloads to actually remove the property, their own supportability overhead should handle that if necessary.

Do you agree that these expectations need to be met while we have both properties? If so, does it work for you if I just add a sentence to mark the old property as deprecated?

graph/articles/collections.md
```diff
@@ -48,7 +48,7 @@ GET https://graph.microsoft.com/beta/teamwork/devices/0f3ce432-e432-0f3c-32e4-3c
   "value": {
     "@odata.type": "#microsoft.graph.teamworkDevice",
     "id": "0f3ce432-e432-0f3c-32e4-3c0f32e43c0f",
-    "deviceType": "CollaborationBar",
+    "deviceType": "CollaborationBar",371
```
mikepizzo: >371 [](http://example.com/codeflow?start=37&length=3)

Is this change intentional?

graph/articles/collections.md
```diff
@@ -316,3 +316,151 @@ Content-Type: application/json
    "value": []
 }
 ```
+
+## 11. Collections of structural types (complex types or primitive types)
+
+Collections of entity types are generally preferable to collections of structual types because collections of structural types must be updated as a single unit, meaning that they are overwritten entirely by new contents, rather than be updated relative to the existing contents.
```
mikepizzo, line 322: this is one of the reasons entities are favorable over complex types within a collection, but limiting to just this case often gets us in trouble.  i.e., a workload may say "this is read-only", or "I'm okay with them replacing the whole collection", and then get into trouble later, either because they do need to update an individual member (or something nested within an individual member), or because they need to refer to an individual member for some other reason (i.e., an auditLog or recommendation might want to reference an individual key credential). Also, it's harder to page collections of complex types.

So, for a variety of reasons, the strong recommendation is to use entity types in collections.

mikepizzo (changes requested):

graph/articles/collections.md
```diff
@@ -316,3 +316,151 @@ Content-Type: application/json
    "value": []
 }
 ```
+
+## 11. Collections of structural types (complex types or primitive types)
```
mikepizzo, line 320: the second option (changing the property to a nav prop) doesn't really work for collections of primitive types, right?

graph/articles/collections.md
mikepizzo: This is one of the reasons entities are favorable over complex types within a collection, but limiting to just this case often gets us in trouble. i.e., a workload may say "this is read-only", or "I'm okay with them replacing the whole collection", and then get into trouble later, either because they do need to update an individual member (or something nested within an individual member), or because they need to refer to an individual member for some other reason (i.e., an auditLog or recommendation might want to reference an individual key credential). Also, it's harder to page collections of complex types.

So, for a variety of reasons, the strong recommendation is to use entity types in collections.

```suggestion
Entity types are generally preferred for collections since complex types within a collection cannot be individually referenced. Collections of complex types, including any nested properties, must be updated as a single unit, entirely replacing the existing contents. Even if your API is read-only today, modeling it as a collection of entities gives you more flexibility in referencing individual members now and in the future. 
```

graph/articles/collections.md
mikepizzo: I agree these semantics need to hold while both properties are supported, and that we should add a sentence that the old property should be deprecated.

graph/articles/collections.md
```diff
@@ -316,3 +316,151 @@ Content-Type: application/json
...
+  "error": {
+    "code": "badRequest",
+    "message": "'keyCredentials' and 'keyCredentialsAsEntities' cannot be updated in the same request.",
+}
+```
+
+TODO should this be a 409 conflict instead?
+TODO implement this in WebApi
+
+### 11.2 `$select` overloading
```
mikepizzo: I'm not sure that calling it "$select overloading" conveys the intent of this option -- i.e., if I was thinking about different ways to evolve the collection, I wouldn't look for an option called "Select overloading". That is more of a by-product than an option. The option is to replace the complex type with an entity type -- in order for this to work, however, existing select semantics need to be maintained.

```suggestion
### 11.2 Redefine as Entity Type
```

graph/articles/collections.md
```diff
@@ -316,3 +316,151 @@ Content-Type: application/json
...
+</EntityType>
+
+<ComplexType Name="keyCredential">
+  <Property Name="keyId" Type="Edm.Guid" />
+  <Property Name="endDateTime" Type="Edm.DateTimeOffset" />
+  ...
+</ComplexType>
+```
+and a scenario arises that requires, for example, to remove individual `keyCredential`s from the collection. 
+There are two options forward: //// TODO do we want to offer both options, or just one?
```
mikepizzo: I think it's okay to offer both options, although I think the second option has strong advantages over the first.

graph/articles/collections.md
mikepizzo: I think the versioning and deprecation doc says, if you can't find a better name, use _v2.  Should we use that for this example?

graph/articles/collections.md
mikepizzo: Agree that these semantics need to hold for the duration during which both versions of the property are supported. I think this is called out in the deprecation spec as well.  Happy to leave this here, and add that the old property should be marked as deprecated.

corranrogue9: Sounds good, the old property got marked deprecated in a previous iteration, so I'm resolving this comment

PR #563, opened 2024-11-07
Include nextLink in header

heaths:
Allows client libraries that would otherwise deserialize lazily to return an abstraction that returns the page of results without having to deserialize the body.

Azure SDKs that will (or may) benefit from this include:

* Rust
* Java
* (Possibly) .NET
* (Possibly) C++

For Rust, we provide a `Stream` implementation where the customer deserializing our models or theirs is actually polling the future (backing Rust's async). We are not in the stack so we can't get the `nextLink` unless we first deserialize the response body, which also means that we are having to poll the future to completion instead of the customer during deserialization. This is a waste of time and resources: if we were to reuse the same models we might otherwise give them in the golden path (simple method call to deserialize into our "suggested" model), much time is spent allocating strings on the heap, for example. Even if we create private models just to deserialize the `nextLink`, we're still polling the future to completion and parsing through all the bytes just to get a `String` for the `nextLink` out of it. We then do a shallow copy of that buffer and pass it back to the client so they can deserialize our models or theirs. Java, I'm told, works similarly and .NET is considering similar designs.

### DIFF
--- a/azure/Guidelines.md
+++ b/azure/Guidelines.md
@@ -566,6 +566,14 @@ Note: The service is responsible for performing any URL-encoding required on the
 
 <a href="#collections-nextlink-includes-all-query-params" name="collections-nextlink-includes-all-query-params">:white_check_mark:</a> **DO** include any query parameters required by the service in `nextLink`, including `api-version`.
 
+<a href="#collections-include-nextlink-header-for-more-results" name="collections-include-nextlink-header-for-more-results">:white_check_mark:</a> **DO** also include the same `nextLink` URL and query parameters in the [`Link` header](https://httpwg.org/specs/rfc8288.html#link-target) with `rel="next"`:
+
+```
+Link: <{opaqueUrl}>; rel="next"
+```
+
+This allows client libraries to provide abstractions to loop through results without having to spend computational resources to deserialize the response body just to get the `nextLink`.
+
 <a href="#collections-response-array-name" name="collections-response-array-name">:ballot_box_with_check:</a> **YOU SHOULD** use `value` as the name of the top-level array field unless a more appropriate name is available.
 
 <a href="#collections-no-nextlink-on-last-page" name="collections-no-nextlink-on-last-page">:no_entry:</a> **DO NOT** return the `nextLink` field at all when returning the last page of the collection.


heaths:
To note, I also consider just a simple purpose-built header name like `azure-next-link` but did find precedent for a `link` header using `rel="next"`, so I went with that here initially.

johanste:
I am not in favor of adding more ways of doing the same thing (i.e. duplicating information in responses). And while it is unfortunate to have to deserialize twice, I could make similar "make it more performant" arguments against using `json` as the serialization format - there are certainly more efficient formats that could be argued for in the name of performance.

Please note that there are dragons lurking here - e.g. what rewrites have to be performed to reliably put the API behind an API gateway etc. 

From a philosophical perspective; I  very much believe we need to be conservative in adding expected behaviors for services (as opposed to clarifying guidelines) or else we will be causing more disparity among various azure services. Consistency includes being consistent with how existing services are implemented.

But if we are going to do this, we should use the [`Link` header](https://httpwg.org/specs/rfc8288.html#header) with the appropriate `rel`/[link relations](https://www.iana.org/assignments/link-relations/link-relations.xhtml).

heaths:
Maybe we don't treat it as another way to do the same thing, then, but a better way of accomplishing something. We've learned from experience there can be a better way. Isn't that why we have 4+ different ways of doing LROs that our SDKs all must support but have settled on a [recommendation to use the `operation-location` response header](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#lro-returns-operation-location)?

Initially this was to be in addition to the body parameter, but we could make whether to find it in the body or the header just like we make how exactly the LRO is implemented an authorable behavior.

Yes, there's still two ways of doing things: the old, less-ideal way; and the new, ideal way. Same as LROs. But at least it's not duplicative if it's authorable in the spec. And, if we do so, I recommend that newer Azure Core TypeSpec models default to the header.

johanste:
We have certainly made mistakes with LROs. And I would not be in favor of making a small incremental improvement at the expense of cross-service consistency there either. 

There were many discussions on this when the link-in-body pattern was decided on (although many of which predates my active participation on the review board, so I have to do some serious archeology to see if I can even find any of the old discussions). Having the response body be self describing and not needing additional metadata (e.g. from headers) has a lot of value. As has the ability to paginate sub-collections. In other words, using a header was considered and ultimately decided not to be the best solution. 

We will figure out how to do this in TypeSpec separately, but in practical terms, new guidelines pretty much have to be opt-in since existing services are unlikely to adhere to them.

heaths:
> As has the ability to paginate sub-collections.

Which service(s) have this? I've not the experience you have, but I haven't seen one. Still, if there are any or would be any, they could opt for the body parameter.

> There were many discussions on this when the link-in-body pattern was decided on

I also brought up perf when I joined the Azure SDK team and it wasn't a priority them either. Not until about 2 years later, and by then a lot of APIs had GA'd for older SDKs. But newer SDKs adopted more performant patterns. And those are for GC'd languages, of which Rust is not, which is also why performance is so much better and more important.

We're funded to work on Azure SDKs for Rust because of major performance improvements. Just today I was reading about an Indonesian bank that saw something like a 460% improvement in memory consumption. For larger pages, this isn't small gains.

If the Board holds firm to your stance, we'll have to consider "trimmed" models that have only the `nextLink` parameter (field) but that still means buffering the entire response and reading through it in its entirety. At least we can avoid allocating a bunch of data on heap we don't need. /cc @rickwinter @jhendrixMSFT

johanste:
I see no way to *not* consider to handle services that use what is the current standard pagination pattern for Azure services since adopting a new guideline will not automatically mean that existing services implement it.

johanste:
At least one of the IoT family of services had nested pageable sub-collections. We try to avoid them. But when they are needed, they are needed. I know of a couple of other APIs that really should have (or at least would have been better off having) used them because they ended up with API patterns that required *a lot* of small calls to gather the data needed.

heaths:
> I see no way to _not_ consider to handle services that use what is the current standard pagination pattern for Azure services since adopting a new guideline will not automatically mean that existing services implement it.

I didn't say we wouldn't support `nextLink` in the body. Quite the opposite. I said we would have to support it because of all the services out there. That's not to say we can't improve going forward.

mikekistler:
I think I agree with @johanste that making this a "DO" guideline is too strong.

But I have no concerns with making this a "MAY" and possibly a "SHOULD".

As an argument for "MAY", I can imagine a scenario where a service team comes to the API Stewardship board asking for a recommendation about how best to provide better performance for paginated responses -- essentially the scenario Heath is trying to target. If the solution Heath has described is indeed the best solution (and we should only have one because otherwise SDK support will be either challenging or non-existent), we should document that _somehow_. And the API Guidelines seem like the right place. Making it a "MAY" guideline seems very low risk, but still will potential benefit.

Can we agree on including this in the guidelines but as "MAY" rather than "DO"?

If someone wants to argue strongly for "SHOULD" we could debate that. But if we can get quick agreement on "MAY" I propose we take that path and see how well this is received in service teams before making any stronger recommendation.

heaths:
Per discussion in our LT meeting, we'll not proceed with this for now but could revisit it as a "MAY" later.

PR #569, opened 2025-04-30
Enhanced the guidance for HTTP response code

apiarya:
This pull request updates the Microsoft Graph API guidelines to provide clearer guidance on HTTP response codes and error handling. The changes include a detailed table of recommended HTTP response codes, a new section on choosing between 400 and 422 status codes, and the removal of outdated error rules.

### Enhancements to HTTP response code guidelines:

* Added a comprehensive table of HTTP response codes with descriptions, handling locations, and links to further documentation. This includes codes like `200 OK`, `201 Created`, `400 Bad Request`, `422 Unprocessable Content`, and more.
* Introduced a new section to clarify the distinction between `400 Bad Request` and `422 Unprocessable Content`, providing examples and validation scenarios for each.

Ref: https://learn.microsoft.com/en-us/graph/errors

### DIFF
--- a/graph/GuidelinesGraph.md
+++ b/graph/GuidelinesGraph.md
@@ -277,6 +277,37 @@ Operation resources MUST have a binding parameter that matches the type of the b
 
 For an additional list of standard HTTP methods, see the [RFC7230](https://www.rfc-editor.org/rfc/rfc9112)).
 
+### HTTP Response Codes
+
+When designing Graph APIs, it is crucial to use appropriate HTTP response codes to ensure consistency and clarity for API consumers. The following table outlines the recommended HTTP response codes for various scenarios. All 4XX and 5XX error responses **MUST** include a JSON body as specified in the [Error Response Modeling](#error-handling) section. For further details, refer to [HTTP response status codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status).
+
+| HTTP Code                 |  Description | Handled At         |
+|---------------------------|--------------|--------------------|
+| 200 OK                    | Indicates that the request was successful, and the response contains the requested data or confirmation.                                 | Service            |
+| 201 Created               | Indicates that a new resource has been successfully created. The `Location` header SHOULD contain the URI of the newly created resource. | Service            |
+| 202 Accepted              | Indicates that the request has been accepted for processing, but the processing is not yet complete. See [Long-Running operations](/graph/patterns/long-running-operations.md). | Service            |
+| 204 No Content            | Indicates that the request was successful, but there is no content to return in the response body.                                       | Service            |
+| 400 Bad Request           | Indicates that the request is malformed or contains invalid parameters. The response SHOULD include details about the error.             | Service            |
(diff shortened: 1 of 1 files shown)


graph/GuidelinesGraph.md
```diff
@@ -277,6 +278,37 @@ Operation resources MUST have a binding parameter that matches the type of the b
...
+| 202 Accepted              | Indicates that the request has been accepted for processing, but the processing is not yet complete. See [Long-Running operations](/graph/patterns/long-running-operations.md). | Service            |
+| 204 No Content            | Indicates that the request was successful, but there is no content to return in the response body.                                       | Service            |
+| 400 Bad Request           | Indicates that the request is malformed or contains invalid parameters. The response SHOULD include details about the error.             | Service            |
+| 401 Unauthorized          | Indicates that the request is unauthenticated. The client MUST authenticate before making the request.                                      | Gateway            |
+| 403 Forbidden             | Indicates that the client is authenticated but does not have sufficient permissions to perform the requested operation.    | Gateway or Service |
+| 404 Not Found             | Indicates that the requested resource could not be found.                                                                                | Gateway or Service |
```
TarkanSevilmis, line 294: Missing 409.

TarkanSevilmis, line 294: Need clarification on what does requesting a resource means specifically. Examples when foo with id 123 doesn't exist, which ones return 404, they are all "requesting" a resource in the general sense?
GET /foo/123

GET /foo/?$filter=id eq 123

POST /createBarForFoo
{'fooId' = 123}

GET /bar/?$filter=fooId eq 123

graph/GuidelinesGraph.md
```diff
@@ -277,6 +278,37 @@ Operation resources MUST have a binding parameter that matches the type of the b
...
+| 422 Unprocessable Content | Indicates that a request is syntactically correct but semantically wrong per the business policy, server state or domain rules.          | Service            |
+| 429 Too Many Requests     | Indicates that the client has exceeded the rate limits or throttling thresholds.                                                         | Gateway or Service     |
+| 500 Internal Server Error | Indicates an internal server error. This is a generic error message for unexpected failures.                                             | Service            |
```
TarkanSevilmis, line 297: Can we add this should not be intentionally returned for an understood error case? How do people feel about that?

graph/GuidelinesGraph.md
```diff
@@ -277,6 +278,37 @@ Operation resources MUST have a binding parameter that matches the type of the b
...
+| 503 Service Unavailable   | Indicates that the service is temporarily unavailable, often due to maintenance or overload.                                             | Gateway or Service |
+
+#### Choosing Between 400 and 422 Response Codes
+
+When designing APIs, it's important to differentiate between client errors that are syntactically incorrect (400 Bad Request) and those that are semantically incorrect (422 Unprocessable Content). The following table provides guidance on when to use each response code, along with examples.
+
+|   | 400 Bad Request      | 422 Unprocessable Content     |
```
TarkanSevilmis, line 304: What about:
/foo/$filter= 'name' eq 'x'
when the service doesn't support name filtering (or insert any other non-preference based request that is valid in the protocol)
Is that also 422, or something else?

PR #586, opened 2026-05-12
Add nullable enum guidance: null vs none sentinel values

OlgaPodo:
## Summary

Add a new 'Nullable enums' section to `graph/patterns/enums.md` with guidance on when to use `null` vs a `none` sentinel member for enum properties.

The existing page mentions that 'a nullable enum can very likely be avoided by adding a none member' but does not explain:
- When `null` is the right choice vs `none`
- The anti-pattern of combining both
- Decision criteria for API designers

## Changes

New `#### Nullable enums` section added after 'Enum or Booleans', containing:
- **null vs none comparison table** — clarifies the semantic difference
- **Note on unknownFutureValue** — reinforces that it is always required and unrelated to nullability
- **Prefer none over nullable** — recommended pattern with CSDL example
- **When nullable is appropriate** — three criteria that must all apply
- **Anti-pattern: nullable + none** — warns against combining both with example

## Context

Identified as a guideline gap during Graph API review discussions.

### DIFF
--- a/graph/patterns/enums.md
+++ b/graph/patterns/enums.md
@@ -102,6 +102,72 @@ Additionally speaking, depending on the situation, a nullable enum can very like
 
 If used, `EnumType` names should be singular if the are non-flags enums, and the names should be plural if they are flags enums.
 
+
+#### Nullable enums
+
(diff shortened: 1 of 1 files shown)


chrisspre (commented):
looks good to me. much clearer and explains the context

